PR #7, opened 2021-12-04
added automatic pull for missing images

j4ckr3d:
Check it out.
If it's fine for you, i'll add immediate action prompts for all those error messages/command requests, so the user could use it right away without worrying about the rest

### DIFF
--- a/distrobox-create
+++ b/distrobox-create
@@ -1,20 +1,4 @@
 #!/bin/sh
-# Copyright (C) 2021 Luca Di Maio
-# This file is part of distrobox <https://github.com/89luca89/distrobox>.
-#
-# distrobox is free software: you can redistribute it and/or modify
-# it under the terms of the GNU General Public License as published by
-# the Free Software Foundation, either version 3 of the License, or
-# (at your option) any later version.
-#
-# distrobox is distributed in the hope that it will be useful,
-# but WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-# GNU General Public License for more details.
-#
-# You should have received a copy of the GNU General Public License
-# along with distrobox. If not, see <http://www.gnu.org/licenses/>.
-
 # POSIX
 # Expected env variables:
 #	HOME
--- a/distrobox-create
+++ b/distrobox-create
@@ -162,7 +162,7 @@ generate_command() {
 
 # check that we have a complete distrobox installation or
 # entrypoint and export will not work.
-[ -z "${distrobox_entrypoint_path}" ] && echo "Error: no distrobox-init found in $PATH" && exit 1
+[ -z "${distrobox_entrypoint_path}" ] && echo "Error: no distrobox-init found in ${PATH}" && exit 1
 
 # First, check if the image exists in the host
 if ! podman image exists "${container_image}"; then
--- a/distrobox-create
+++ b/distrobox-create
@@ -192,7 +192,7 @@ if ! podman image exists "${container_image}"; then
     read -r yn
     case $yn in
         y|Y|yes )
-            podman pull ${container_image}
+            podman pull "${container_image}"
             ;;
         n|N|no )
             echo >&2 "next time, run this command first:"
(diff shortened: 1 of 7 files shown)


distrobox-create
```diff
@@ -166,8 +166,19 @@ generate_command() {
 
 # First, check if the image exists in the host
 if ! podman image exists "${container_image}"; then
-	echo "Image not found, run this command first:"
-	echo "	podman pull ${container_image}"
+        echo "Image not found."
+    read -n1 -p "Do you want to pull the image now?[y/n] " yn
```
89luca89, line 170: As above, check if you can use stderr for this (maybe splitting between an echo and a read) 

If the user says yes and the pull is successful, we should continue with the creation and we need a valid command for that

89luca89:
Thanks a lot for the contribution, I like this idea!

Just a couple of comments on the use of stderr
In the current code stdout is used because a missing image will halt the distrobox creation, in this case we need to pay attention to this, as a successful "yes" and pull should continue normally with the distrobox creation

j4ckr3d:
Fixed errors and also the redirects in distrobox-enter

distrobox-create
```diff
@@ -166,9 +166,20 @@ generate_command() {
 
 # First, check if the image exists in the host
 if ! podman image exists "${container_image}"; then
-	echo "Image not found, run this command first:"
-	echo "	podman pull ${container_image}"
-	exit 1
+    echo >&2 "Image not found."
+    echo >&2 "Do you want to pull the image now?[y/n]"
```
89luca89, line 170: By default if I do not enter anything and just go with `enter` it defaults to Yes, so I'd put the y capital in [Y/n]

distrobox-create
```diff
@@ -166,9 +166,20 @@ generate_command() {
...
+    read -r yn
+    case $yn in
+        [Yy]* )
+            podman pull ${container_image}
```
89luca89, line 174: I'd quote the variable to  prevent globbing and word splitting

distrobox-create
```diff
@@ -166,9 +166,20 @@ generate_command() {
 
 # First, check if the image exists in the host
 if ! podman image exists "${container_image}"; then
-	echo "Image not found, run this command first:"
```
89luca89, line 169: can you run

`shfmt -s <file-name>`

on the files? It unifies the indentation with the rest of the code this way

to install it `go install mvdan.cc/sh/v3/cmd/shfmt@latest`

89luca89:
> Fixed errors and also the redirects in distrobox-enter

thanks a lot! I've added a couple of comments about quoting and indentation

distrobox-create
```diff
@@ -166,9 +166,20 @@ generate_command() {
...
-	echo "	podman pull ${container_image}"
-	exit 1
+    echo >&2 "Image not found."
+    echo >&2 "Do you want to pull the image now?[y/n]"
+    read -r yn
+    case $yn in
+        [Yy]* )
```
89luca89, line 173: please use just

y | Y | yes ) 

to be more precise with the input (like replying yNO will count as yes)

distrobox-create
```diff
@@ -166,9 +166,20 @@ generate_command() {
...
+            podman pull ${container_image}
+            break
+            ;;
+        [Nn]* )
```
89luca89, line 177: please use just

n | N  | no ) 

to be more precise with the input (like replying nYes will count as no)

distrobox-create
```diff
@@ -166,9 +166,20 @@ generate_command() {
...
+            echo >&2 "next time, run this command first:"
+            echo >&2 "  podman pull ${container_image}"
+            exit 1
+            ;;
```
89luca89, line 181: to remove the shellcheck warning about the brake at line 175, you have to remove it from there and add a default case for it:

```
	*) # Default case: If no more options then break out of the loop.
		break ;;
```

this way we also solve the default to yes

distrobox-create
89luca89: More than that, I'd not default to Yes but to No, just because this potentially could cause a big download.

distrobox-create
j4ckr3d: The warning still remains, i'll just add a warning message and stop it there.

Issue #8, opened 2021-12-04
[Feature] Add support for docker


89luca89:
We should work with both podman and docker
if both are available, better use podman

89luca89:
Right now there are a couple of issues to sort out compatibility:

```
--ulimit host
--userns keep-id
```

won't work on `docker`, we need to find the equivalent 

---

Right now we check if we're inside a container by checking the existence of the file

`/run/.containerenv`

We should go from

```
if [ ! -f /run/.containerenv ]; then ...
```

to

```
if [ ! -f /run/.containerenv ] && [ ! -f /.dockerenv ]; then  ...
```

Or we can set a `container_env_file` variable when we discover which container manager we're using.

---

89luca89:
To discover which container manager we will use, I'd just add this to the top

```
container_manager="podman"
if ! command -v podman >/dev/null; then
        container_manager="docker"
fi
if ! command -v docker >/dev/null; then
	printf >&2 "Missing dependency: cannot find a container manager. Install either podman or docker.\n"
	exit 127
fi
```

then just replace all the instances of `podman` with `${container_manager}`

PR #31, opened 2021-12-15
Feature/docker support

89luca89:
Work in progress to add `docker` support to distrobox.
This work is part of the issue #8 

Right now the container manager is automatically detected, with a preference of podman over docker.

Not all flags are supported by docker so these flags will be behind a container manager check:

```sh
if [ "${container_manager}" = "podman" ]; then
--userns keep-id
--ulimit host
--mount type=devpts,destination=/dev/pts"
fi
```

---

The add of a `--headless` flag to `distrobox-enter` was needed because `docker` has problems instantiating a tty if not ran from a terminal itself, this is a non-problem on `podman`, so to work around this, I've added this flag to optionally disable `--interactive --tty` flags.

This is applied to services and apps exported.

---

This should be ok for the non-rootless use-case of docker (adding yourself to docker group).

Need to investigate rootless docker setups.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -30,7 +30,7 @@ jobs:
         # uses: luizm/action-sh-checker@7f44869033b40ee4ffe7dc76c87a1bc66e3d025a
         uses: luizm/action-sh-checker@v0.3.0
         env:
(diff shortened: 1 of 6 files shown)


alegrey91 (commented):
LGTM

j4ckr3d (approved):
Noice.

Issue #11, opened 2021-12-04
[ Feature ] Export arbitrary binary from container & direnv


FrauH0lle:
Hello and thank you very much for creating this project!

As far as I understand, `distrobox-export --app` can only export applications which have a `.desktop` file. Would it be possible to extend this to programs in `/bin` or `usr/bin`? Or have an option like `distrobox-export --bin` which does that?

Background:

I discovered your project and [dahbox](https://github.com/jecicorp/dahbox) around the same time during the last days when I was searching for a possibility to have project specific environments (and not using the nix package manager or nixos) with [direnv](https://github.com/direnv/direnv).
Both projects have a similar scope but yours seems to be more generic in terms of available container images. That's why I am opening this request.

The workflow I mean is described [here](https://github.com/jecicorp/dahbox#boxes-per-projet), and I would imagine it could look similar to this:

```bash
# Direnv setup
echo "PATH_add $PWD/.distrobox-bin" > .envrc
direnv allow

# Export from distrobox
# I currently work a lot in R, so I will use it as an example
distrobox-export --bin R --export-to "$PWD/.distrobox-bin"

# This would create a bash script $PWD/.distrobox-bin/R which would point to R in the container which in turn would be added to $PATH via direnv when the user works in this project folder.
```

Would this be feasible for this project or is it out of scope?

89luca89:
Hello, and thank you for taking some time to open this issue!

I see no problems with including a `--bin/-b` option to `distrobox-export` if done correctly we could use it as a base for also `--app/-a` and `--service/-s` 

What I would like to find, is a way to make it a bit more seamless with the host
I mean without having to fiddle with `direnv` or rc files in general

On one hand `direnv` easily solves the problem of having multiple versions of the same `bin` exported 
On the other hand having to manage direnvs and rc files is not an easy fit to do correctly and safely

Having the `--export-to` flag, would allow you to have this direnv workflow (which I like)

But on the other hand, I would like to have a way (or a sane default) so that it works without that flag, and exports the bin in the user home's $PATH

The little problem is that $PATH is shared between host and container, so it would create a loop if you launch the exported bin inside the container :smile: 

A quick way would be to add this snippet to a `~/.profile` 

```
	if [ ! -f /run/.containerenv ]; then
            # we are on the host
            PATH=$PATH:$HOME/.distrobox-bin/
         fi
```

[...]

PR #34, opened 2021-12-16
Add Gentoo to supported host distros

FrauH0lle:
As I tested distrobox on Gentoo during #11 and still use it, I think we can add Gentoo to the list.


89luca89:
Cool glad it's all working well :)

Merging right now!

PR #57, opened 2021-12-27
Update README.md

alcir:
CentOS Stream 8 and 9 as Containers Distros


89luca89:
Cool stuff thanks a lot 😄 
Any particularity in the process or it worked straight away?

alcir:
> Cool stuff thanks a lot smile Any particularity in the process or it worked straight away?

As far as I can say, they worked straight away.

89luca89:
Cool!

alcir:
For the records, fr what it worth, also Red Had UBI images seem to work.
`distrobox-create -i registry.access.redhat.com/ubi8/ubi:latest --name ubi8`

You have to install `vte-profile` to avoid `bash: __vte_prompt_command: command not found` messages
`sudo dnf install vte-profile`
and `glibc-langpack-en` to avoid `Failed to set locale, defaulting to C.UTF-8`
`sudo dnf install glibc-langpack-en`

89luca89:
If you want you can open a new issue about this so we can detect the various packages needed (and in case add it to other distros) so we can ensure the support

PR #58, opened 2021-12-27
Add license header and SPDX-License-Identifier

alcir:
I'm not a lawyer, but IMHO a license header in each script looks more professional :sweat_smile: and maybe in line with some policy.

### DIFF
--- a/distrobox-create
+++ b/distrobox-create
@@ -3,6 +3,7 @@
 #
 # This file is part of the distrobox project: https://github.com/89luca89/distrobox
 #
+# Copyright (C) 2021 distrobox contributors
 # distrobox is free software; you can redistribute it and/or modify it
 # under the terms of the GNU General Public License version 3
 # as published by the Free Software Foundation.
--- a/distrobox-create
+++ b/distrobox-create
@@ -1,4 +1,20 @@
 #!/bin/sh
+# SPDX-License-Identifier: GPL-3.0-only
+#
+# This file is part of the distrobox project: https://github.com/89luca89/distrobox
+#
+# distrobox is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License version 3
+# as published by the Free Software Foundation.
+#
+# distrobox is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+# General Public License for more details.
+#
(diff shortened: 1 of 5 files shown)


TheEvilSkeleton:
The FSF and I agree with you. https://github.com/89luca89/distrobox/blob/7a51438b6fa6e6a24d5ce86fa8734dd9de4d4343/LICENSE#L629-L648

TheEvilSkeleton (approved):
The FSF and I agree with you. https://github.com/89luca89/distrobox/blob/7a51438b6fa6e6a24d5ce86fa8734dd9de4d4343/LICENSE#L629-L648

TheEvilSkeleton (commented):
I suggest moving the LICENSE file to COPYING because the FSF recommends so: https://www.gnu.org/licenses/gpl-howto.en.html. Perhaps even use the [Markdown](https://www.gnu.org/licenses/gpl-3.0.md) format to make it far more readable than RAW.

89luca89:
Oh, nice I like the markdown format of the license 😄 

I was also wondering about how to maintain the copyright present but general enough to cover the contributors, I think this solution works well I like it 👍

TheEvilSkeleton:
> I was also wondering about how to maintain the copyright present but general enough to cover the contributors, I think this solution works well I like it +1

Yup. Either way, "2021" implies "or later", so it doesn't have to be constantly updated.

89luca89:
> > I was also wondering about how to maintain the copyright present but general enough to cover the contributors, I think this solution works well I like it +1
> 
> Yup. Either way, "2021" implies "or later", so it doesn't have to be constantly updated.

I was referring to the copyright owners, having it as " distrobox contributors" should be general enough to not have problems with contributions copyright

89luca89:
@alcir should be done by last commit right?

TheEvilSkeleton (changes requested):
I suggest moving LICENSE to COPYING (or even Markdown version).

89luca89:
> I suggest moving LICENSE to COPYING (or even Markdown version).

So we remove the LICENSE file, and put [THIS](https://www.gnu.org/licenses/gpl-3.0.md) in it's place?

TheEvilSkeleton:
Yup, and call it `COPYING.md`. So it's `LICENSE` -> `COPYING.md`.

89luca89:
Cool, let's keep this PR for the license headers, I'll do that change in the `main` directly

89luca89:
Implemented :+1:

TheEvilSkeleton (approved):

PR #63, opened 2021-12-29
Update README.md

alcir:
Add silent option to curl, since it could "hide" the sudo password request


89luca89:
Seems the right thing to do 😄

PR #64, opened 2021-12-29
Proposals

digiserg:


distrobox-create
```diff
@@ -118,9 +124,13 @@ if [ -z "${container_clone}" ] && [ -z "${container_image}" ]; then
 	container_image="${container_image_default}"
 fi
 
+if [ "${container_name}" == "" ]; then
```
89luca89, line 127: This behavior is not defined in POSIX sh, probably it's better to check if it's empty using

`if [ -z  "${container_name}" ]; then`

distrobox-create
```diff
@@ -118,9 +124,13 @@ if [ -z "${container_clone}" ] && [ -z "${container_image}" ]; then
...
+	container_name=$(basename $container_image | sed -E 's/:/_/g')
```
89luca89, line 128: I like this idea :smile: what about using an hyphen instead of underscore? so we can build the default name we had before: `fedora-toolbox-35` from `fedora-toolbox:35`

distrobox-create
```diff
@@ -308,10 +318,14 @@ fi
 # First, check if the image exists in the host.
 # If not prompt to download it.
 if ! ${container_manager} inspect --type image "${container_image}" >/dev/null 2>&1; then
-	# Prompt to download it.
-	printf >&2 "Image not found.\n"
-	printf >&2 "Do you want to pull the image now? [y/n] "
-	read -r response
+	if [ "${non_interactive}" == "0" ]; then
```
89luca89: We should use a single `=`, as `==` is not defined in standard POSIX sh

` 	if [ "${non_interactive}" = "0" ]; then`

distrobox-create
```diff
@@ -90,6 +92,10 @@ while :; do
 			shift
 		fi
 		;;
+	-n | --non-interactive)
```
89luca89, line 95: `-n` is already taken from `--name`, we could use something like `-N` or `-ni`?

distrobox-create
89luca89: Also there are a couple of fixes about quoting stuff, to put it in line with the rest of the POSIX stile:

`container_name="$(basename "${container_image}" | sed -E 's/:/-/g')"`

distrobox-create
```diff
@@ -118,9 +124,13 @@ if [ -z "${container_clone}" ] && [ -z "${container_image}" ]; then
...
+if [ "${container_name}" == "" ]; then
+	container_name=$(basename $container_image | sed -E 's/:/_/g')
+fi
+
 # We depend on a container manager let's be sure we have it
 # First we use podman, else docker
-container_manager="podman"
+container_manager=${DB_CONTAINER_MANAGER:-"podman"}
 # Be sure we have a container manager to work with.
 if ! command -v podman >/dev/null; then
```
89luca89, line 135: This way if we set DB_CONTAINER_MANAGER=docker (for example) and we do not have podman installed, it will fail
we should at this point change it to 

` if ! command -v  "${container_manager}" >/dev/null; then`

So that the first check is always the one we declared

89luca89 (changes requested):

89luca89:
Hey @digiserg thanks a lot for the contribution :smile: 
I want to tackle a couple of issues before merging, more about some bashisms than anything else :+1:

distrobox-create
```diff
@@ -308,10 +318,14 @@ fi
 # First, check if the image exists in the host.
 # If not prompt to download it.
 if ! ${container_manager} inspect --type image "${container_image}" >/dev/null 2>&1; then
-	# Prompt to download it.
-	printf >&2 "Image not found.\n"
-	printf >&2 "Do you want to pull the image now? [y/n] "
-	read -r response
+	if [ ${non_interactive} -eq 0 ]; then
```
89luca89, line 321: We should quote this, linting will fail with:

`Prefer double quoting even when variables don't contain special characters. [SC2248]`

89luca89 (changes requested):

89luca89 (approved):

PR #68, opened 2021-12-29
Create uninstall script file

misobarisic:
Since installing distrobox was simple, why shouldn't uninstallation have the same trait? I have created a simple uninstall script and updated the README.

One thing that has gotten my attention was the mysterious `voice` located next to [download zip](https://github.com/89luca89/distrobox#installation). Can't figure out what it's supposed to mean.


alcir:
> One thing that has gotten my attention was the mysterious `voice` located next to [download zip](https://github.com/89luca89/distrobox#installation). Can't figure out what it's supposed to mean.

I think that it is referred to: here on github, you can clone the repository with the git command using the provided url, or download the repository as a zip file when you click on the green "clone" button.

89luca89:
I don't have problems with this idea yes, I'd like to test it a bit before merging 
What about instead of using the `--path` we just find the various distrobox commands in PATH?

misobarisic:
Finding distrobox commands in PATH seems reasonable until we come to the problem of there being multiple identically named files in PATH matching distrobox's __footprint__. What if the user wanted to uninstall one but keep the other. I don't know... Maybe the user wants to keep multiple versions installed?

89luca89:
> Finding distrobox commands in PATH seems reasonable until we come to the problem of there being multiple identically named files in PATH matching distrobox's **footprint**. What if the user wanted to uninstall one but keep the other. I don't know... Maybe the user wants to keep multiple versions installed?

mmmh thinking about it you're right yea I use it this way myself :sweat_smile:

misobarisic:
> > One thing that has gotten my attention was the mysterious `voice` located next to [download zip](https://github.com/89luca89/distrobox#installation). Can't figure out what it's supposed to mean.
> 
> I think that it is referred to: here on github, you can clone the repository with the git command using the provided url, or download the repository as a zip file when you click on the green "Code" button located top right.

Now that I think of it. If `--path` is not specified the script could check for matches in PATH and take additional input for each file or directory (dir makes more sense than just file by file), whereas were it specified, it would only delete there.

89luca89:
> > > One thing that has gotten my attention was the mysterious `voice` located next to [download zip](https://github.com/89luca89/distrobox#installation). Can't figure out what it's supposed to mean.
> > 
> > 
> > I think that it is referred to: here on github, you can clone the repository with the git command using the provided url, or download the repository as a zip file when you click on the green "Code" button located top right.
> 
> Now that I think of it. If `--path` is not specified the script could check for matches in PATH and take additional input for each file or directory (dir makes more sense than just file by file), whereas were it specified, it would only delete there.

In the install, it defaults to /usr/local/bin if not specified
in the uninstall we could do with PATH, and with rm -i so it's interactive, just to be super-sure we do not delete something important for error

misobarisic:
Being super-sure sounds good but have a look at this:

There are 2 installations of distrobox. One in /usr/bin and one in /usr/local/bin

Without specifying `--path` the script would ask for 8 file deletion confirmations max.
Instead of asking:
- do you want to delete `/usr/bin/distrobox-enter`
- do you want to delete `/usr/bin/distrobox-create`
- ...

we could just ask for confirmations regarding the 2 paths since for each path you'd have duplicate and unnecessary user inputs.

89luca89:
> Being super-sure sounds good but have a look at this:
> 
> There are 2 installations of distrobox. One in /usr/bin and one in /usr/local/bin
> 
> Without specifying `--path` the script would ask for 8 file deletion confirmations max. Instead of asking:
> 
>     * do you want to delete `/usr/bin/distrobox-enter`
> 
>     * do you want to delete `/usr/bin/distrobox-create`
> 
>     * ...
> 
> 
> we could just ask for confirmations regarding the 2 paths since for each path you'd have duplicate and unnecessary user inputs.

We should just do 1 delete for each file, respecting the PATH priority
If we do something in the lines of

`rm -i $(command -v distrobox-enter)`

it will delete only the *first* appearance in order of priority

So if for example the PATH is

.local/bin:/usr/bin

if a .local/bin/distrobox-enter exists, it will be prompted for deletion, leaving /usr/bin/distrobox-enter intact

misobarisic:
> We should just do 1 delete for each file, respecting the PATH priority

There could also be a scenario where not all distrobox files are present in the same directory.

That's why it should delete all distrobox files it is able to find in the first path containing **a** distrobox file...? 

My previous idea of per path answers could be implemented as an interactive mode. `--interactive`.

89luca89:
> > We should just do 1 delete for each file, respecting the PATH priority
> 
> There could also be a scenario where not all distrobox files are present in the same directory.
> 
> That's why it should delete all distrobox files it is able to find in the first path containing **a** distrobox file...?
> 
> My previous idea of per path answers could be implemented as an interactive mode. `--interactive`.

Using `command -v` you should be able to find all the (first in path) instances of the various files
And if you have some strange setup with only 2 distrobox commands in a custom path, well you should use the `--path` option then :smile: 
I'd say if the `--path` options is not set, we're dealing with a quite plain use-case of defaults, so either we use `command -v` or we can even default to `/usr/local/bin/distrobox-*` much like the installer itself

what do you think?

misobarisic:
I have no affinity to either `command -v` or defaulting to `/usr/local/bin/distrobox-*`. It is your choice.

89luca89:
I'd say let's follow the install and just default to /usr/local/bin if no --path specified

misobarisic:
> I'd say let's follow the install and just default to /usr/local/bin if no --path specified 

That is already present in my commits.

89luca89:
OK perfect I just have a couple of comments to resolve and we should be good to go 👍

uninstall
```diff
@@ -0,0 +1,67 @@
...
+
+# Ensure the foundamental variables are set and not empty, we will not proceed if
+# they are not all set.
+[ ! -d "${dest_path}" ] && printf >&2 "Path %s does not exist.\n" "${dest_path} " && exit 1
+[ ! -w "${dest_path}" ] && printf >&2 "Cannot write into %s, permission denied.\n" "${dest_path}" && exit 1
+
+# uninstall
+if [ -f "${dest_path}/distrobox-enter" ]; then
+	for file in distrobox-*; do
+		rm "${dest_path}${file}"
```
89luca89, line 62: Let's use `rm -i` just to be safe 😄

uninstall
```diff
@@ -0,0 +1,67 @@
...
+	set -o xtrace
+fi
+
+# Ensure the foundamental variables are set and not empty, we will not proceed if
+# they are not all set.
+[ ! -d "${dest_path}" ] && printf >&2 "Path %s does not exist.\n" "${dest_path} " && exit 1
+[ ! -w "${dest_path}" ] && printf >&2 "Cannot write into %s, permission denied.\n" "${dest_path}" && exit 1
+
+# uninstall
+if [ -f "${dest_path}/distrobox-enter" ]; then
```
89luca89, line 60: Instead of this, we can simply have the list of files declared explicitly (enter, create etc etc)
And loop over that, so that if (for example) an user removes create file, but leaves the others, the uninstaller will work anyway

89luca89 (changes requested):

uninstall
misobarisic: I quite like this approach to be honest

uninstall
misobarisic: Better safe than sorry

misobarisic (commented):
If this script helps at least one person, it will be deemed as success!

89luca89:
@misobarisic I've added some new files to the repository:

- distrobox
- distrobox-rm
- distrobox-list

let's include them in the uninstall also :smile:

misobarisic:
> @misobarisic I've added some new files to the repository:
> 
>     * distrobox
> 
>     * distrobox-rm
> 
>     * distrobox-list
> 
> 
> let's include them in the uninstall also smile

Changed `distrobox-*` to `distrobox*` to also catch `distrobox*`.

89luca89:
@misobarisic I think there is a bit of a conflict with the new README file configuration to be solved

misobarisic:
@89luca89 Will do later today.

Thank you for reminding me! 😅

misobarisic:
I believe everything's been resolved with my questionable commits

89luca89:
> I believe everything's been resolved with my questionable commits

😆 all OK just missing the license headers and were done 👍

misobarisic:
> > I believe everything's been resolved with my questionable commits
> 
> laughing all OK just missing the license headers and were done +1

What license headers? What are you talking about?

89luca89:
> > > I believe everything's been resolved with my questionable commits
> > 
> > 
> > laughing all OK just missing the license headers and were done +1
> 
> What license headers? What are you talking about?

If you used also the other scripts there are the headers for the gplv3 license put just under the 1st line, in the comment on the code review I've added the snippet to add 👍

misobarisic:
> If you ~used~ see also the other scripts there are the headers for the gplv3 license put just under the 1st line, in the comment on the code review I've added the snippet to add +1

As far as I am concerned, I have done it.

89luca89 (approved):

Issue #60, opened 2021-12-28
[Feature] add support for RedHat UBI


89luca89:
As pointed out in #57 by @alcir 

> For the records, for what it worth, also Red Had UBI images seem to work.
distrobox-create -i registry.access.redhat.com/ubi8/ubi:latest --name ubi8
You have to install vte-profile to avoid bash: __vte_prompt_command: command not found messages
sudo dnf install vte-profile
and glibc-langpack-en to avoid Failed to set locale, defaulting to C.UTF-8
sudo dnf install glibc-langpack-en

Let's investigate on this and see if we can integrate the missing dependencies also in other distribution families

alcir:
Well also CentOS Stream 9 requires to install `vte-profile` in order to avoid the message 
`bash: __vte_prompt_command: command not found`

89luca89:
Fixed with merge of #69

PR #69, opened 2021-12-29
Fix for #60

alcir:
I don't know if this introduces some other issue, but now CentOS Stream 9 and RH UBI8 don't show the reported messages anymore

### DIFF
--- a/distrobox-init
+++ b/distrobox-init
@@ -201,7 +201,9 @@ if ! command -v mount || ! command -v passwd ||
 			procps-ng \
 			shadow-utils \
 			sudo \
-			util-linux
+			util-linux \
+			vte-profile \
+			glibc-langpack-en
 
 	elif command -v pacman; then
 		pacman -Sy --noconfirm \


89luca89:
Cool thanks, I'd like to test this for all the redhat/fedora family containers we support, but I see no issue

I would like to find also the equivalent files for the other families so that we can maintain a coherent list between all the distributions :+1:

89luca89:
So, I've tested with all the distributions families and the vte integration is not a problem, we can add to all of them the seguent package list:

```shell
	if command -v apk; then
		apk add \
			"${shell_pkg}" \
			findutils \
			ncurses \
			procps \
			shadow \
			sudo \
			util-linux \
			vte3

	elif command -v apt-get; then
		export DEBIAN_FRONTEND=noninteractive
		apt-get update
		apt-get install -y \
			"${shell_pkg}" \
			findutils \
			libvte-common \
			ncurses-base \
			passwd \
			procps \
			sudo \
			util-linux

	elif command -v dnf; then
		dnf install -y \
			"${shell_pkg}" \
			findutils \
			vte-profile \
			ncurses-base \
			passwd \
			procps-ng \
			shadow-utils \
			sudo \
			util-linux

	elif command -v pacman; then
		pacman -Sy --noconfirm \
			"${shell_pkg}" \
			findutils \
			ncurses \
			procps-ng \
			shadow \
			sudo \
			util-linux \
			vte-common

	elif command -v slackpkg; then
		slackpkg update
		yes | slackpkg install -default_answer=yes -batch=yes \
			"${shell_pkg}" \
			libvte-2 \
			findutils \
			ncurses \
			procps \
			shadow \
			sudo \
			util-linux

	elif command -v xbps-install; then
		xbps-install -Sy \
			"${shell_pkg}" \
			findutils \
			ncurses-base \
			procps-ng \
			shadow \
			sudo \
			util-linux

	elif command -v yum; then
		yum install -y \
			"${shell_pkg}" \
			findutils \
			ncurses-base \
			passwd \
			procps \
			shadow-utils \
			sudo \
			util-linux \
			vte-profile

	elif command -v zypper; then
		zypper install -y \
			"${shell_pkg}" \
			findutils \
			libvte-2 \
			ncurses \
			procps \
			shadow \
			sudo \
			util-linux
```

---

Regarding the glibc-en it's a bit of a problem as in the same families is either present or not (see C7 vs C9 etc etc) So I'd leave it out as it does not create problems to initialize and use the container, and can be part of a pet container configuration that later the user can do

89luca89:
Hey @alcir any news on this?

89luca89:
Merged this as it effectively solves 60, I've took care of the other distros :+1: 
thanks a lot for the contribution :tada:

PR #88, opened 2022-01-01
Improve support for NixOS, and add more info about NixOS support to README.

FedX-sudo:
There are still a lot of bugs with NixOS which I am working to iron out, such as the following:

* Distrobox cannot be installed in /nix/store, not a huge deal, but very nice to have.
* Container distros don't support sudo, I don't no why, I just get an error whenever I try to become a super user.
* Kali does not come with DPKG or apt preinstalled, again I don't know why, but it is a pretty important feature of Kali

However, I was able to imporve NixOS support with this PR. I changed distrobox-init so that it recognizes symbolicly linked dotfiles, which many NixOS users (including myself) have. ﻿

### DIFF
--- a/distrobox-init
+++ b/distrobox-init
@@ -385,7 +385,7 @@ passwd --delete root
 if [ -d "/etc/skel" ]; then
 	skel_files="$(find /etc/skel/ -type f || :)"
 	for skel_file in ${skel_files}; do
-        if [ ! -f "${container_user_home}/$(basename "${skel_file}")" -a !-L "${container_user_home}/$(basename "${skel_file}")" ]; then
+		if [ ! -f "${container_user_home}/$(basename "${skel_file}")" -a !-L "${container_user_home}/$(basename "${skel_file}")" ]; then
 			cp "${skel_file}" "${container_user_home}"
 			chown "${container_user_uid}":"${container_user_gid}" "${container_user_home}/$(basename "${skel_file}")"
 		fi
--- a/distrobox-init
+++ b/distrobox-init
@@ -385,7 +385,7 @@ passwd --delete root
 if [ -d "/etc/skel" ]; then
 	skel_files="$(find /etc/skel/ -type f || :)"
 	for skel_file in ${skel_files}; do
-		if [ ! -f "${container_user_home}/$(basename "${skel_file}")" || ! -L "${container_user_home}/$(basename "${skel_file}")" ]; then
+		if [ ! -f "${container_user_home}/$(basename "${skel_file}")" ] && [ !-L "${container_user_home}/$(basename "${skel_file}")" ]; then
 			cp "${skel_file}" "${container_user_home}"
 			chown "${container_user_uid}":"${container_user_gid}" "${container_user_home}/$(basename "${skel_file}")"
 		fi
--- a/distrobox-init
+++ b/distrobox-init
@@ -385,7 +385,7 @@ passwd --delete root
 if [ -d "/etc/skel" ]; then
 	skel_files="$(find /etc/skel/ -type f || :)"
 	for skel_file in ${skel_files}; do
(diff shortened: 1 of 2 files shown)


89luca89:
Hi @FedX-sudo thanks for the contribution!

There are a couple of problems on code level, just check the run it will give you info on what to change (formatting and replace `-a` with && for example)

You can use shellcheck locally if you want to validate it before committing.

About support, at least using docker on my NixOS VM I have no problems with sudo nor with apt/dpkg on kali linux, seems like it's probably some problem with rootless podman setup on NixOS itself? I had much better luck with docker on that regard (but I'm a NixOS super-noob so it was expected :smile: )

FedX-sudo:
Ok, that sounds right for the NixOS Podman issue. :D I will see if I can figure out how to switch to Docker, and as far as spell check goes, I still have not gotten around to installing a spell checker (which I very much should) and I am the crazy person who develops everything in NeoVim.

89luca89:
> Ok, that sounds right for the NixOS Podman issue. :D I will see if I can figure out how to switch to Docker, and as far as spell check goes, I still have not gotten around to installing a spell checker (which I very much should) and I am the crazy person who develops everything in NeoVim.

I do everything in vim myself :smile: I just use ALE and installed shellcheck from the repositories

Anyway there is a complete report in the action log below

distrobox-init
```diff
@@ -385,7 +385,7 @@ passwd --delete root
 if [ -d "/etc/skel" ]; then
 	skel_files="$(find /etc/skel/ -type f || :)"
 	for skel_file in ${skel_files}; do
-		if [ ! -f "${container_user_home}/$(basename "${skel_file}")" ]; then
+		if [ ! -f "${container_user_home}/$(basename "${skel_file}")" -a !-L "${container_user_home}/$(basename "${skel_file}")" ]; then
```
89luca89, line 388: instead of `-a` let's use `&&` to be like the rest of the code

`		if [ ! -f "${container_user_home}/$(basename "${skel_file}")"  ] && [ !-L "${container_user_home}/$(basename "${skel_file}")" ]; then`

89luca89 (changes requested):

distrobox-init
```diff
@@ -385,7 +385,7 @@ passwd --delete root
 if [ -d "/etc/skel" ]; then
 	skel_files="$(find /etc/skel/ -type f || :)"
 	for skel_file in ${skel_files}; do
-		if [ ! -f "${container_user_home}/$(basename "${skel_file}")" ]; then
+		if [ ! -f "${container_user_home}/$(basename "${skel_file}")" || ! -L "${container_user_home}/$(basename "${skel_file}")" ]; then
```
89luca89, line 388: this won't work on posix sh
See the last comment on how to implement it :+1:

89luca89 (changes requested):

distrobox-init
89luca89: Also I think we want an and not an or right?

89luca89 (approved):

distrobox-init
```diff
@@ -385,7 +385,7 @@ passwd --delete root
 if [ -d "/etc/skel" ]; then
 	skel_files="$(find /etc/skel/ -type f || :)"
 	for skel_file in ${skel_files}; do
-		if [ ! -f "${container_user_home}/$(basename "${skel_file}")" ]; then
+		if [ ! -f "${container_user_home}/$(basename "${skel_file}")" ] && [ !-L "${container_user_home}/$(basename "${skel_file}")" ]; then
```
89luca89, line 388: `distrobox-init:388:73: error: You are missing a required space here. [SC1035]`

Need a space between the `!` and `-L`

89luca89 (changes requested):

89luca89 (approved):

89luca89:
Cool stuff, merging :+1: 
Thanks a lot!

FedX-sudo:
Of course! Thanks for such a cool script!

PR #102, opened 2022-01-06
Update compatibility.md

misobarisic:
Since `xhost` specific errors (not being able to connect to the xserver, the app not exporting) were not only bound to Void Linux, I decided to add a more inclusive message to compatibility.md.
I encountered the issue of not being able to connect to the host xserver from the container on a minimal Arch installation which previously didn't have the need for configuring `xhost`.

### DIFF
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -49,9 +49,11 @@ Distrobox has been successfully tested on:
 | EndlessOS | 4.0.0 | |
 | OpenSUSE | Leap 15<br>Tumbleweed | |
 | OpenSUSE MicroOS | 20211209 | |
-| Void Linux | glibc | Systemd service export will not work.<br>To setup graphical app exporting, you need to install `xhost` and add this to your `~/.xinitrc`: <br> `xhost +si:localuser:$USER`.  |
+| Void Linux | glibc | Systemd service export will not work. |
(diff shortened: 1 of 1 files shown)


docs/compatibility.md
```diff
@@ -49,10 +49,12 @@ Distrobox has been successfully tested on:
...
 | Windows WSL2 | | __NOTE WSL2 support is preliminary, and there are many bugs present, any help in improving support is appreciated__ <br> Currently you must work around some incompatibility between WSL2 and Podman, namely [THIS](https://github.com/containers/podman/issues/12236). <br>Install into WSL2 any of the supported distributions in this list. <br> Ensure you have an entry in the `fstab` for the `/tmp` folder:<br> `echo 'tmpfs /tmp tmps defaults 0 0' >> /etc/fstab`.<br>Then reboot the WSL machine `wsl --shutdown` <br>Note that `distrobox export` is not supported on WSL2 and will not work. |
 
+If your container is not able to connect to your host xserver, make sure to install `xhost` on the host machine and run `xhost +si:localuser:$USER` or add it to your `~/.xinitrc`.
```
89luca89, line 56: Should Be **and** add it to xinirc to make it permanent

89luca89 (changes requested):

89luca89:
Cool thanks a lot greatly appreciated 😄

docs/compatibility.md
misobarisic: The goal is to make it run on every x session startup. Editing `.xinitrc` may not affect it if you're, let's say, using KDE Plasma but your `.xinitrc` starts Gnome.

misobarisic:
I've updated it with a better and less ambiguous message.

89luca89 (approved):

89luca89:
Cool stuff thanks!

PR #111, opened 2022-01-10
init: add support for gentoo

nullpointerarray:
This adds support for Gentoo.

### DIFF
--- a/distrobox-init
+++ b/distrobox-init
@@ -400,8 +400,9 @@ if ! grep -q "${container_user_name}" /etc/passwd; then
 	fi
 fi
 # Delete password for root and user
-printf "password1234ADFG_+\npassword1234ADFG_+\n" | passwd "${container_user_name}"
-printf "password1234ADFG_+\npassword1234ADFG_+\n" | passwd root
+TEMP_PASSWORD=$(cat /proc/sys/kernel/random/uuid)
+printf "%s\n%s\n" "${TEMP_PASSWORD}" "${TEMP_PASSWORD}" | passwd "${container_user_name}"
+printf "%s\n%s\n" "${TEMP_PASSWORD}" "${TEMP_PASSWORD}" | passwd root
 passwd --delete "${container_user_name}"
 passwd --delete root
 
--- a/distrobox-init
+++ b/distrobox-init
@@ -275,6 +275,21 @@ if ! command -v mount || ! command -v mount || ! command -v passwd ||
 			shadow \
 			sudo \
 			util-linux
+	elif command -v emerge; then
+		# set the password requirements to the minimum. gentoos passwd policy is stricter than other distros
+		# otherwise setting up the container can fail.
+		# this is needed on ubuntu 21.10
+		# works by setting /etc/security/passwdqc.conf to min=1,1,1,1,1 + everything except the first line from
+		# the original
+		printf "min=1,1,1,1,1\n%s" "$(tail -n +2 /etc/security/passwdqc.conf)" > /etc/security/passwdqc.conf
+		# update repos
+		emerge --sync &&
+			emerge --ask=n --autounmask-continue "${shell_pkg}" \
+				findutils \
+				procps \
+				shadow \
+				sudo
+
 	else
 		printf "Error: could not find a supported package manager.\n"
 		printf "Error: could not set up base dependencies.\n"
--- a/distrobox-init
+++ b/distrobox-init
@@ -281,7 +281,7 @@ if ! command -v mount || ! command -v mount || ! command -v passwd ||
 		# this is needed on ubuntu 21.10
 		# works by setting /etc/security/passwdqc.conf to min=1,1,1,1,1 + everything except the first line from
 		# the original
-		printf "min=1,1,1,1,1\n%s" "$(tail -n +2 /etc/security/passwdqc.conf)" > /etc/security/passwdqc.conf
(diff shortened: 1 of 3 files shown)


89luca89:
Hi @nullpointerarray thanks for the contribution!

I'll just put a couple of comments and we should be good to go :+1:

distrobox-init
```diff
@@ -275,6 +275,21 @@ if ! command -v mount || ! command -v mount || ! command -v passwd ||
 			shadow \
 			sudo \
 			util-linux
+	elif command -v emerge; then
+		# set the password requirements to the minimum. gentoos passwd policy is stricter than other distros
+		# otherwise setting up the container can fail.
+		# this is needed on ubuntu 21.10
+		# works by setting /etc/security/passwdqc.conf to min=1,1,1,1,1 + everything except the first line from
+		# the original
+		printf "min=1,1,1,1,1\n%s" "$(tail -n +2 /etc/security/passwdqc.conf)" > /etc/security/passwdqc.conf
```
89luca89, line 284: There are a couple of whitespaces problems, run a `shfmt -s` on the file and it should be fixed
currently the linting github actions fails for this

distrobox-init
```diff
@@ -275,6 +275,21 @@ if ! command -v mount || ! command -v mount || ! command -v passwd ||
 			shadow \
 			sudo \
 			util-linux
+	elif command -v emerge; then
+		# set the password requirements to the minimum. gentoos passwd policy is stricter than other distros
+		# otherwise setting up the container can fail.
+		# this is needed on ubuntu 21.10
+		# works by setting /etc/security/passwdqc.conf to min=1,1,1,1,1 + everything except the first line from
```
89luca89, line 282: is this mandatory on gentoo?
A new approach now is to initialize user's password to a value and then delete the password, gentoo should allow that without problems right?

From your comments seems like it's needed also on Ubuntu 21.10+ but I've been using them daily no problems

89luca89 (changes requested):

distrobox-init
nullpointerarray: The new approach works for me. I misunderstand the code and thought distrobox uses the host password, please ignore the comment. It was needed when initializing the Gentoo container, not when using Ubuntu 21.10.

distrobox-init
```diff
@@ -391,8 +400,8 @@ if ! grep -q "${container_user_name}" /etc/passwd; then
 	fi
 fi
 # Delete password for root and user
-printf "password\npassword\n" | passwd "${container_user_name}"
-printf "password\npassword\n" | passwd root
+printf "password1234ADFG_+\npassword1234ADFG_+\n" | passwd "${container_user_name}"
```
89luca89, line 403: I was thinking it would be nice at this point if we use a random string for the password
We set it in a variable (so that it's possible to use it again for confirmation)

Maybe something like `cat /proc/sys/kernel/random/uuid`?

89luca89 (changes requested):

docs/compatibility.md
```diff
@@ -95,7 +95,7 @@ Distrobox guests tested successfully with the following container images:
 | Ubuntu | 14.04<br>16.04<br>18.04<br>20.04<br>21.10<br>22.04 | docker.io/library/ubuntu:14.04<br>docker.io/library/ubuntu:16.04<br>docker.io/library/ubuntu:18.04<br>docker.io/library/ubuntu:20.04<br>docker.io/library/ubuntu:21.10<br>docker.io/library/ubuntu:22.04    |
 | Kali Linux | rolling | docker.io/kalilinux/kali-rolling:latest |
 | Void Linux | | ghcr.io/void-linux/void-linux:latest-thin-bb-x86_64 <br> ghcr.io/void-linux/void-linux:latest-thin-bb-x86_64-musl <br> ghcr.io/void-linux/void-linux:latest-full-x86_64 <br> ghcr.io/void-linux/void-linux:latest-full-x86_64-musl |
-
+| Gentoo Linux | rolling | [DOCKERFILE HERE](https://github.com/89luca89/distrobox/issues/112#issue-1098097949)
```
89luca89, line 98: Would be better to really explain here, let's point to the official resource here: https://github.com/gentoo/gentoo-docker-images#using-the-portage-container-in-a-multi-stage-build

and maybe provide the example and specify to the user that they need to create the image themselves and it's not ready to use on a registry

89luca89 (changes requested):

distrobox-init
89luca89: Yep

distrobox-init
89luca89: Alternatively

`< /dev/urandom tr -dc '[:alpha:]' | head -c20`

nullpointerarray: Added that. Whats the advantage of that?

docs/compatibility.md
nullpointerarray: I'll create a file and link to it.

distrobox-init
89luca89: > Added that. Whats the advantage of that?

Just that we do not have the password in the repo this way and being just a temp one, we can (and should) use a random one at this point

89luca89 (approved):

89luca89:
Looks good to me now, I'll just probably create a new section in the docs for custom images but that's out of the scope now :smile: 

Thanks a lot for the contribution! Merging :+1:

PR #116, opened 2022-01-13
gen-man: exit if dependency is not met

misobarisic:
Self explanatory title. I have updated the `gen-man` script to exit if `ronn` is not installed.


89luca89:
Ooops 😄 thanks!

PR #133, opened 2022-01-19
create/export: Prefer 'uname -m' to /etc/hostname

nathanchance:
In a Fedora VM that I created on my MacBook Air, `/etc/hostname` is empty,
which means the container's hostname ends with just a period, which
looks odd:

```
$ cat /etc/hostname

$ ./distrobox enter -n fedora -- cat /etc/hostname
fedora.
```

Prefer `uname -n`, which shows the host machine's hostname properly.

```
$ uname -n
fedora

$ ./distrobox enter -n fedora -- uname -n
fedora.fedora

$ ./distrobox enter -n fedora -- cat /etc/hostname
fedora.fedora
```

`uname` is described by POSIX and should be available by default on every
modern Linux system, as it is typically a part of the coreutils package.

Link: https://pubs.opengroup.org/onlinepubs/9699919799/utilities/uname.html

### DIFF
--- a/distrobox-create
+++ b/distrobox-create
@@ -244,7 +244,7 @@ generate_command() {
 	result_command="${container_manager} create"
 	# use the host's namespace for ipc, network, pid, ulimit
 	result_command="${result_command}
-		--hostname ${container_name}.$(cut -d'.' -f1 /etc/hostname)
+		--hostname ${container_name}.$(uname -n)
 		--ipc host
 		--name ${container_name}
 		--network host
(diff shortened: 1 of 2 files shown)


nathanchance:
Sigh, just realized that the commit has a typo in the title (`uname -m` vs. `uname -n`) but at least the actual change is right... Thanks a lot for the quick responses!

89luca89:
Ahah no problem at all, thanks to you for all the contributions you're making :smile:

PR #137, opened 2022-01-22
init: Fix check for find

nathanchance:
Commit 0550f86 ("init: check if find is installed") aimed to check that
find was installed but it duplicated the check for mount instead.

Fix the check to look for find and mention that find is required in the
comment below.


89luca89:
Aha! Nice `find` 😜 
Thanks a lot for the correction

Merging

PR #138, opened 2022-01-22
create, enter: more concise messages

misobarisic:
I removed the `--name` flag from the output message in `distrobox-create` and `distrobox-enter` making it straight to the point.


89luca89:
Cool thanks!

Merging.

PR #141, opened 2022-01-22
create: default to Y

### LINKED ISSUE #139: [Suggestion] create: default to "Y" instead of "N"
Imagine this scenario. You type this command in:

```bash
distrobox create fedora-rawhide --image registry.fedoraproject.org/fedora:rawhide
```

You explicitly provided the image name yet you don't have it installed. In this scenario, `distrobox-create` defaults to `NO` if you just press enter. My suggestion is to default to `YES` since the user's intent would be to use that image even in the case it might have slipped their mind when it comes to previously downloading the image.

misobarisic:
Resolves #139


89luca89:
Cool, thanks for the contribution!

PR #143, opened 2022-01-23
Specify image name flaglessly

misobarisic:
Modified `distrobox-create` to also accept flaglessly specifying image.

Now these would equal the same thing:
- `distrobox-create alpine alpine:latest`
- `distrobox-create -n alpine alpine:latest`
- `distrobox-create alpine -i alpine:latest`

Hopefully somebody finds this useful.

### DIFF
--- a/distrobox-create
+++ b/distrobox-create
@@ -155,6 +155,9 @@ while :; do
 		if [ -z "${container_name}" ] && [ -n "$1" ]; then
 			container_name="$1"
 			shift
+		elif [ -z "${container_image}" ] && [ -n "$1" ]; then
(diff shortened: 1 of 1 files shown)


89luca89:
Thanks @misobarisic 

I'm a bit worried about having multiple flagless inputs, because it then input becomes positional, which is not a solid choice if you ask me
Podman/Docker do this with the image, while enforcing name with `--name`, Toolbx does this with the name while enforcing `-i` for the image, so still just 1 flagless input.

misobarisic:
Alright, yeah. Sounds reasonable.

PR #145, opened 2022-01-27
create: follow /dev/shm symlinks if present

### LINKED ISSUE #144: [Error] `distrobox-enter` does not work when `/dev/shm` points to `/run/shm`
**Describe the bug**

On WSL2, `/dev/shm` is a symlink to `/run/shm` (for whatever reason, seems like Ubuntu used to do that too?). When `/dev` gets bind mounted into the container without `/run/shm`, the symlink is broken and the container runtime rightly complains:

```
$ realpath /dev/shm
/run/shm

$ distrobox enter dev-arch
Error: unable to start container "cbf9efdae48aae4730f9810e73e5322e686bcee1596091c27a257d7a1d9d8b0d": openat2 `dev/shm`: No such file or directory: OCI runtime attempted to invoke a command that was not found
```

**To Reproduce**

In WSL2 or another OS that has this same symlink behavior, a simple `distrobox-create` + `distrobox-enter` should reproduce it.

**Expected behavior**

There should be no error :)

**Additional context**

The issue is fixed if I add `--volume /run/shm:/run/shm` to my `distrobox-create` command but it seems like this is something that should be handled within `distrobox-create`, perhaps with something like:

```diff
diff --git a/distrobox-create b/distrobox-create
index 336fd72..ae355f6 100755
--- a/distrobox-create
+++ b/distrobox-create
@@ -292,6 +292,12 @@ generate_command() {
                --volume /sys:/sys:rslave
                --volume /tmp:/tmp:rslave"

+       dev_shm_path="$(realpath /dev/shm)"
+       if [ "${dev_shm_path}" != "/dev/shm" ]; then
+               result_command="${result_command}
+                       --volume ${dev_shm_path}:${dev_shm_path}:rslave"
+       fi
+
        # If we have a custom home to use,
        #       1- override the HOME env variable
        #       2- expor the DISTROBOX_HOST_HOME env variable pointing to original HOME
```

89luca89:
Resolves #144 

@nathanchance can you test if this resolves? Thanks

### DIFF
--- a/distrobox-create
+++ b/distrobox-create
@@ -298,7 +298,7 @@ generate_command() {
 	# source also in the container.
 	if [ -L "/dev/shm" ]; then
 		result_command="${result_command}
-			--volume $(realpath /dev/shm)=$(realpath /dev/shm)"
+			--volume $(realpath /dev/shm):$(realpath /dev/shm)"
 	fi
 
 	# If we have a custom home to use,
--- a/distrobox-create
+++ b/distrobox-create
@@ -292,6 +292,15 @@ generate_command() {
 		--volume /sys:/sys:rslave
 		--volume /tmp:/tmp:rslave"
 
+	# In some systems, for example using sysvinit, /dev/shm is a symlink
+	# to /run/shm, instead of the other way around.
+	# Resolve this detecting if /dev/shm is a symlink and mount original
+	# source also in the container.
+	if [ -L "/dev/shm" ]; then
+		result_command="${result_command}
+			--volume $(realpath /dev/shm)=$(realpath /dev/shm)"
+	fi
+
 	# If we have a custom home to use,
 	#	1- override the HOME env variable
 	#	2- expor the DISTROBOX_HOST_HOME env variable pointing to original HOME


distrobox-create
```diff
@@ -292,6 +292,15 @@ generate_command() {
 		--volume /sys:/sys:rslave
 		--volume /tmp:/tmp:rslave"
 
+	# In some systems, for example using sysvinit, /dev/shm is a symlink
+	# to /run/shm, instead of the other way around.
+	# Resolve this detecting if /dev/shm is a symlink and mount original
+	# source also in the container.
+	if [ -L "/dev/shm" ]; then
+		result_command="${result_command}
+			--volume $(realpath /dev/shm)=$(realpath /dev/shm)"
```
nathanchance, line 301: This needs to be `$(realpath /dev/shm):$(realpath /dev/shm)` I think; once I make that change, everything works (as long as I give it a few seconds between `distrobox-create` and `distrobox-enter`, otherwise I get the same error but I assume that is something weird with WSL2, not `distrobox`).

89luca89: oh yea my bad on this

nathanchance (approved):
Thanks a lot for the quick fix!

89luca89:
> Thanks a lot for the quick fix!

Is it working as intended? If yes I'll merge :+1:

nathanchance:
Yes, sorry, that is what I intended with the approval :)

89luca89:
Cool stuff, merging. :smile:

PR #147, opened 2022-01-29
[Improvement] XDG_.*_DIRS environment fixes

deinferno:
Change adds helper function `append_env` and appends default FHS values to `XDG_DATA_DIRS XDG_CONFIG_DIRS` envs, this fixes VSCode's file picker and Nautilus when running Archlinux container on NixOS as host.

### DIFF
--- a/distrobox-enter
+++ b/distrobox-enter
@@ -155,6 +155,8 @@ fi
 #
 # This is not necessary on docker as it is already handled
 # in this way.
+
+# shellcheck disable=SC2312
 if [ -z "${container_manager#*podman*}" ] &&
 	[ -S "/run/user/$(id -ru)/podman/podman.sock" ] &&
 	systemctl --user status podman.socket >/dev/null; then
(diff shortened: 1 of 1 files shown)


89luca89:
Thanks a lot @deinferno 

It's not clear why do you remove the XDG_*_DIRS from the `printenv` and then define them anyway?
I'd advise to just add them (so maintain just the part in line 231 and 232) in case they're not defined. If they're defined they are already passed to the container anyway

Also if possible I'd prefer to use less functions so just use the cycle inside the `generate_command`, as it maintains an easier readability of what's happening in my opinion

deinferno:
`PATH` is excluded from printenv too, i done same with `XDG_DATA_DIRS` `XDG_CONFIG_DIRS` to avoid useless --env pass that would be overwritten by later --env anyway. 

I will replace `append_env` with pre PR `PATH` cycle but will copy paste it two times for `XDG_DATA_DIRS` `XDG_CONFIG_DIRS`, is that gonna be ok?

89luca89:
I think it is OK for now to leave the cycles, we can review it later if needed :+1:

89luca89:
Hey @deinferno 
I've just merged another fix for Nix, can you check if this all works well with that fix included? Thanks!

deinferno:
Yes, /nix is getting mounted and i can execute binaries from /nix/store

deinferno:
It's unrelated to this PR , but It's gonna be a good idea to replace shebangs in every shell script with #!/usr/bin/env bash to make it more compatible with any non-fhs distro in general. I will try to do that tomorrow.

deinferno:
It's unrelated to this PR , but It's gonna be a good idea to replace shebangs in every shell script with #!/usr/bin/env bash to make it more compatible with any non-fhs distro in general. I will maybe try to do that tomorrow.

89luca89:
> It's unrelated to this PR , but It's gonna be a good idea to replace shebangs in every shell script with #!/usr/bin/env bash to make it more compatible with any non-fhs distro in general. I will maybe try to do that tomorrow.

Not really
This project aims to be posix compliant and not every distribution has bash or /usr/bin/env, POSIX standard is `#!/bin/sh`

distrobox-enter
```diff
@@ -156,7 +156,7 @@ fi
 # This is not necessary on docker as it is already handled
 # in this way.
 if [ -z "${container_manager#*podman*}" ] &&
-	[ -S "/run/user/$(id -ru)/podman/podman.sock" ] &&
+	[ -S "/run/user/$(id -ru || :)/podman/podman.sock" ] &&
```
89luca89, line 159: We can remove the `|| :` if this fails for any reason, it's better to fail generally (if `id` fails there must be some problems anyway :smile:

distrobox-enter
```diff
@@ -202,23 +202,43 @@ generate_command() {
...
 	# add to the PATH only after the host's paths, and only if not already present.
 	for standard_path in ${standard_paths}; do
-		if [ -n "${container_paths##*:${standard_path}*}" ]; then
+		if [ -n "${container_paths##*:"${standard_path}"*}" ]; then
 			container_paths="${container_paths}:${standard_path}"
 		fi
 	done
 	result_command="${result_command} --env=\"PATH=${container_paths}\""
+
+	standard_paths="/usr/local/share /usr/share"
```
89luca89, line 224: let's add some more comments explaining why we're doing this?

distrobox-enter
```diff
@@ -202,23 +202,43 @@ generate_command() {
...
+	container_paths="${XDG_DATA_DIRS:=}"
+	for standard_path in ${standard_paths}; do
+		if [ -n "${container_paths##*:"${standard_path}"*}" ]; then
+			container_paths="${container_paths}:${standard_path}"
+		fi
+	done
+	result_command="${result_command} --env=\"XDG_DATA_DIRS=${container_paths}\""
+
+	standard_paths="/etc/xdg"
```
89luca89, line 233: let's add some more comments explaining why we're doing this?

89luca89 (changes requested):

89luca89:
@deinferno added a couple of comments but other than those, I'm ok with this PR :smile:

distrobox-enter
deinferno: Sadly removing it it causes shellcheck to complain `distrobox-enter:159:20: note: Consider invoking this command separately to avoid masking its return value (or use '|| true' to ignore). [SC2312]` any tip on how to avoid it?

89luca89: Aha! That's new :smile:  it was introduces in 0.8.0 :smile: 

For now we can suppress it, with a simple comment before the if
`# shellcheck disable=SC2312`

I'll decide later if this check will be disabled or not in the future :+1:

89luca89 (approved):

89luca89:
Thanks also for this contribution! 😄

Merging.

PR #148, opened 2022-01-29
Add support for /nix/ volume within the container

FedX-sudo:
I have been messing around with Distrobox, and one issue I discovered was it did not mount /nix/ in the container which is needed for NixOS and systems with Nix. I therefore quickly added that feature, and support for NixOS is a lot better. Thank you again for all the work that has gone into this!


89luca89:
Thanks @FedX-sudo 
I was wondering tho what will happen with this Mount if we create a nix distrobox on a nix host?
Will this create problems?

FedX-sudo:
Now that I think about it, yeah, that probably would, in which case, it would also create problems if Nix was installed on the system in general.

FedX-sudo:
So I should probably have it set up more like: 
```
if [ -d /nix/ ] && ![ echo $container_image | grep "nix" ] && ![ echo $container_name | grep "nix"]; then 
```

89luca89:
We could move the thing in the init
And check if /nix does not exist inside the container, you could link it to /run/host/nix?

FedX-sudo:
It would have to be on /nix because stuff is symbolically linked to /nix, and without that directory in that path it kind of just plain breaks on NixOS.

89luca89:
> It would have to be on /nix because stuff is symbolically linked to /nix, and without that directory in that path it kind of just plain breaks on NixOS.

Yes you could check if /nix already exist in the container (in which case let's just skip to not break stuff)
If not exists, we can bind Mount /run/host/nix to /nix in the container, that should work right?

FedX-sudo:
I don't see why not. That is going to be one crazy chain of mounts and links...

89luca89:
> I don't see why not. That is going to be one crazy chain of mounts and links...

Absolutely :joy: 
But what things are not working if /nix is not present in the container?
Trying to understand what  does this mount fix

FedX-sudo:
On NixOS that is where all the binaries are stored. Everything is symlinked from /nix/store, and some config stuff from /nix/var, so if /nix does not exist, none of the NixOS binaries exist...

FedX-sudo:
Actually, this may be a non-issue. I was testing NixOS to see how it would break, and it looks like it's breaking for completely unrelated reasons.

89luca89:
> On NixOS that is where all the binaries are stored. Everything is symlinked from /nix/store, and some config stuff from /nix/var, so if /nix does not exist, none of the NixOS binaries exist...

Makes sense
Then the support in init approach should be doable 
In the end bindmounting in the init is not that different from doing that in the --volume phase

89luca89:
> Actually, this may be a non-issue. I was testing NixOS to see how it would break, and it looks like it's breaking for completely unrelated reasons.

Probably the mount would fix the issue of usinf stuff in tour ~/.nix directory nothing more right?

FedX-sudo:
> > Actually, this may be a non-issue. I was testing NixOS to see how it would break, and it looks like it's breaking for completely unrelated reasons.
> 
> Probably the mount would fix the issue of usinf stuff in tour ~/.nix directory nothing more right?


No, I reverted that change to test and ran it verbose and got this error: 
```
++ podman --log-level debug logs -t --since 2022-01-29T09:19:07.891457283-07:00 nixos
+ container_manager_log='2022-01-29T09:19:08.207913000-07:00 /root/.nix-profile/bin/find
2022-01-29T09:19:08.210838000-07:00 Error: could not find a supported package manager.
2022-01-29T09:19:08.211919000-07:00 Error: could not set up base dependencies.
2022-01-29T09:19:08.212317000-07:00 Error: An error occurred'
+ case "${container_manager_log}" in
+ printf '%s\n' '2022-01-29T09:19:08.207913000-07:00 /root/.nix-profile/bin/find
2022-01-29T09:19:08.210838000-07:00 Error: could not find a supported package manager.
2022-01-29T09:19:08.211919000-07:00 Error: could not set up base dependencies.
2022-01-29T09:19:08.212317000-07:00 Error: An error occurred'
2022-01-29T09:19:08.207913000-07:00 /root/.nix-profile/bin/find
2022-01-29T09:19:08.210838000-07:00 Error: could not find a supported package manager.
2022-01-29T09:19:08.211919000-07:00 Error: could not set up base dependencies.
2022-01-29T09:19:08.212317000-07:00 Error: An error occurred
```

89luca89:
> > > Actually, this may be a non-issue. I was testing NixOS to see how it would break, and it looks like it's breaking for completely unrelated reasons.
> > 
> > 
> > Probably the mount would fix the issue of usinf stuff in tour ~/.nix directory nothing more right?
> 
> No, I reverted that change to test and ran it verbose and got this error:
> 
> ```
> ++ podman --log-level debug logs -t --since 2022-01-29T09:19:07.891457283-07:00 nixos
> + container_manager_log='2022-01-29T09:19:08.207913000-07:00 /root/.nix-profile/bin/find
> 2022-01-29T09:19:08.210838000-07:00 Error: could not find a supported package manager.
> 2022-01-29T09:19:08.211919000-07:00 Error: could not set up base dependencies.
> 2022-01-29T09:19:08.212317000-07:00 Error: An error occurred'
> + case "${container_manager_log}" in
> + printf '%s\n' '2022-01-29T09:19:08.207913000-07:00 /root/.nix-profile/bin/find
> 2022-01-29T09:19:08.210838000-07:00 Error: could not find a supported package manager.
> 2022-01-29T09:19:08.211919000-07:00 Error: could not set up base dependencies.
> 2022-01-29T09:19:08.212317000-07:00 Error: An error occurred'
> 2022-01-29T09:19:08.207913000-07:00 /root/.nix-profile/bin/find
> 2022-01-29T09:19:08.210838000-07:00 Error: could not find a supported package manager.
> 2022-01-29T09:19:08.211919000-07:00 Error: could not set up base dependencies.
> 2022-01-29T09:19:08.212317000-07:00 Error: An error occurred
> ```

This is with normal distrobox, nix host and what type of container?

FedX-sudo:
Nixos from docker.io/nixos/nix:latest.

FedX-sudo:
Also not a NixOS host specific issue, I have the exact same error on Fedora35 host.

89luca89:
Ahaha completely forgot that nix container are not supported in the first place 😅

FedX-sudo:
So I assume this branch is ready to merge then?

89luca89:
Yep sorry was AFK :smile: 

I don't see in the foreseeable future support for nix docker images as there is not way to just support them in the init.
If in the future the situation changes, we can move it to the init aswell

Merging.

FedX-sudo:
Awesome! Thank you! 

Also, Nix has it's own unprivileged container system thing. Maybe point users looking for NixOS over there?

89luca89:
Can you inform me better on this? I'm completely new to the nix world :sweat_smile: 
You could open a PR to point it in the readme if you want

FedX-sudo:
Yeah totally! Nix is... unique to say the least.

89luca89:
It's something I'll need to try seriously in my life :sweat_smile: But for now immutable/ostree distros are just a lot more appealing to me

FedX-sudo:
Totally, coming from MacOS I was looking for the freedom to mess around on root however I wanted, so I can't say ostree stuff is super appealing to me, but I do like the idea, and /nix/store is immutible so it strikes a rather nice balence in my opinion.

89luca89:
At which point I just think... why not...

image

FedX-sudo:
Absolutely. I probably should try Kinoite, just add it to the list with LFS...

PR #151, opened 2022-01-29
update compatibility.md

FedX-sudo:
I updated support notes for NixOS, and added a note about using Nix shells instead of a NixOS container.

### DIFF
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -108,7 +108,7 @@ can take a while as it will download and install the missing dependencies.
 A small time tax to pay for the ability to use any type of image.
 This will **not** occur after the first time, **subsequent enters will be much faster.**
 
-NixOS is not a supported container distro, and there are currently no plans to bing support to it. If you are looking for unprivlaged NixOS environments, we suggest you look into [nix-shell](https://nixos.org/manual/nix/unstable/command-ref/nix-shell.html). 
+NixOS is not a supported container distro, and there are currently no plans to bring support to it. If you are looking for unprivlaged NixOS environments, we suggest you look into [nix-shell](https://nixos.org/manual/nix/unstable/command-ref/nix-shell.html). 
 #### New Distro support
 
 If your distro of choice is not on the list, open an issue requesting support for it,
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -51,7 +51,7 @@ Distrobox has been successfully tested on:
 | OpenSUSE | Leap 15<br>Tumbleweed | |
 | OpenSUSE MicroOS | 20211209 | |
 | Void Linux | glibc | Systemd service export will not work. |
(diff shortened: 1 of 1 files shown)


docs/compatibility.md
```diff
@@ -108,6 +108,7 @@ can take a while as it will download and install the missing dependencies.
 A small time tax to pay for the ability to use any type of image.
 This will **not** occur after the first time, **subsequent enters will be much faster.**
 
+NixOS is not a supported container distro, and there are currently no plans to bing support to it. If you are looking for unprivlaged NixOS environments, we suggest you look into [nix-shell](https://nixos.org/manual/nix/unstable/command-ref/nix-shell.html).
```
89luca89, line 111: Little typo *to bring

docs/compatibility.md
```diff
@@ -51,7 +51,7 @@ Distrobox has been successfully tested on:
...
+| NixOS | 21.11 | Currently you must have your default shell set to Bash, if it is not, make sure you edit your configuration.nix so that it is. <br>Also make sure to mind your executable paths. Sometimes a container will not have nix paths, and sometimes it will not have its own paths. <br> To install distrobox:<br>`mkdir -p ~/.local/bin`<br>Add `PATH=$PATH:$HOME/.local/bin` to your bashrc<br>Execute [THIS](#installation) command without sudo.<br>To setup Docker, look [HERE](https://nixos.wiki/wiki/Docker) <br>To setup Podman, look [HERE](https://nixos.wiki/wiki/Podman) and [HERE](https://gist.github.com/adisbladis/187204cb772800489ee3dac4acdd9947) |
```
89luca89, line 54: I think we take care of the standard FSH PATH inside the container anyway [HERE](https://github.com/89luca89/distrobox/blob/e1b8eb73554164bcb5020987ff1598f50c25854c/distrobox-enter#L213)

89luca89 (changes requested):

docs/compatibility.md
FedX-sudo: I am horrible at typos. I seriously need a new keyboard.

89luca89: Don't worry, half of the whole project is a typo itself.

89luca89 (approved):

FedX-sudo:
Thank you! Also, found a bug, I will be opening a bug report, which is why I added that note to the document. On some Debian based systems like Debian and Kali (but not Ubunut?), it's not importing non-standard paths like .local/bin, and I am not sure why...

89luca89:
> Thank you! Also, found a bug, I will be opening a bug report, which is why I added that note to the document. On some Debian based systems like Debian and Kali (but not Ubuntu?), it's not importing non-standard paths like .local/bin, and I am not sure why...

Let's open a bug and investigate :+1:

PR #157, opened 2022-02-01
Packages links and install instructions for openSUSE and SLE as host distros

dfaggioli:


89luca89:
Thanks a lot @dfaggioli for the contribution!

PR #160, opened 2022-02-05
create/enter: path regression fix

deinferno:
My merged commit [5ecfe6d](https://github.com/89luca89/distrobox/pull/146/commits/5ecfe6d333d714f948cb1a82afa9a1427c1d06b5) caused regression for users that created container via `./distrobox-create` instead of adding it to `$PATH` causing volumes to be relative to current folder instead of absolute path. 

This commit fixes this by returning `realpath` but with `--strip` argument that will allow absolute path but won't it to follow symlinks with is ideal case for both FHS and NixOS users. Also adds `realpath --strip` to `DISTROBOX_ENTER_PATH` env to prevent breakage of exported programs between updates of NixOS.


89luca89:
Hi @deinferno 
Thanks for the contribution!

Sadly it seems that `realpath` flags are a bashism and not POSIX standard, in fact this won't work if `/bin/sh` is busybox (eg on Alpine)

We could try with `readlink` at this point?

deinferno:
I replaced realpath -s with extra check that only uses realpath if path is not absolute. I tried readlink it still follows first link and changes path to unstable real /nix/store path.

89luca89:
> I replaced realpath -s with extra check that only uses realpath if path is not absolute. I tried readlink it still follows first link and changes path to unstable real /nix/store path.

Cool I'll test with my nixos VM when back at PC
Are you using another mean of installation of distrobox? As this seems like you're installing distrobox itself using a nix package?

deinferno:
Yes, i use derivation that i adopted from other user. https://github.com/NixOS/nixpkgs/issues/156514
I don't push it into nixpkgs cause it got hacky stuff like `xhost` into `shellInit`

89luca89:
> Yes, i use derivation that i adopted from other user. [NixOS/nixpkgs#156514](https://github.com/NixOS/nixpkgs/issues/156514) I don't push it into nixpkgs cause it got hacky stuff like `xhost` into `shellInit`

Ah cool didn't know they were creating a package for it 😄 
Males sense this problem never arose for me as I always tested the `curl` way

Are you using that specific package?

deinferno:
Yes, you can force it to use local folder by replacing `src = pkgs.fetchgit` with `src = /absolute/path/to/distrobox` and rebuild with `nixos-rebuild switch --impure`

89luca89:
@deinferno 

What about instead of doing the various `realpath` or `readlinks` or checks for what matters, we use something like:

```sh
distrobox_entrypoint_path="$(cd $(dirname "$0"); pwd)/distrobox-init
distrobox_export_path="$(cd $(dirname "$0"); pwd)/distrobox-export
```

This resolves the issue in #163 and it does not resolve the symlink (so it's nix friendly)

What do you think?

deinferno:
Yes, i think this will work fine. I change commit to that soon.

89luca89:
Looks good to me, merging!

Thanks for your contribution!

PR #167, opened 2022-02-11
install: conditional default install path

### LINKED ISSUE #166: [Suggestion] Change install/uninstall script based on whether running with sudo privileges or not
The install script defaults to `/usr/local/bin`. What if it defaulted to `~/.local/bin` if ran with `./install` and `/usr/local/bin` with `sudo ./install`. 

The install script should also give some output when successfully installed instead of the current nothing.

misobarisic:
Resolves #166.

I've also added a `successfully installed` message.

### DIFF
--- a/install
+++ b/install
@@ -20,6 +20,10 @@
 # POSIX
 
 dest_path="/usr/local/bin"
+if [ "$EUID" -ne 0 ]; then
+	dest_path="$HOME/.local/bin"
+fi
+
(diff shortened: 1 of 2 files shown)


misobarisic:
Shellcheck errors on main are causing this to fail as well.

89luca89:
@misobarisic sorry my bad, fixed on main

PR #174, opened 2022-02-19
FIX #173

### LINKED ISSUE #173: [Error] Quotes are not preserved in `distrobox-enter`
**Describe the bug**
If you need to run a command with full shell env (Eg. referencing the `$HOME` var in a distrobox created with custom home folder)
You must run the command with `distrobox-enter --name sandbox -- bash -l -c "command arg1 arg2"`.

The issue here is that quotes are not preserved and the the command executed is `podman exec [...] sandbox bash -l -c command arg1 arg2`. For that reason `command` never receives the arguments `arg1 arg2`.

**To Reproduce**

1) create a box with custom home:
```
$ mkdir /tmp/home && distrobox-create -i registry.fedoraproject.org/fedora-toolbox:35 --name tmp --home /tmp/home
```
2) echo the `$HOME` variable from the box
```
$ distrobox-enter tmp -- echo '$HOME'
/var/home/pietro
``` 
This is expected because the box environment is not loaded, nor the default entrypoint script is executed

3) echo the `$HOME` variable from the box after loading the environment
```
$ distrobox-enter tmp -- bash -l -c 'echo $HOME'
   <-- an empty line here
```

**Expected behavior**

At point 3) should print `/tmp/home`

**Additional context**
The issue is created by the following assignment in `distrobox-enter` (`$*` do not preserve quotes):
```
        -e | --exec | --)
                shift
                container_command=$*
                break
```

**Possible solutions**
1) use `container_command=${*@Q}` instead of `container_command=$*`, this preserves the quotes but requires Bash 4 or later
2) do some magics parsing the arguments and escaping the quotes

pbertera:
This small change fixes #173 but introduces a requirement for bash > 4.
I'm not sure if is acceptable

### DIFF
--- a/distrobox-enter
+++ b/distrobox-enter
@@ -117,7 +117,7 @@ while :; do
 		;;
 	-e | --exec | --)
 		shift
-		container_command=$*
+		container_command=${*@Q}
 		break
 		;;
 	*) # Default case: If no more options then break out of the loop.


89luca89:
Hey @pbertera 
Thanks for the contribution, but I'd like to leave it as POSIX compliant and avoid bashisms

But completely agree on the issue to be fixed, let's find another way around this :+1:

89luca89:
I'm closing this as on the issue we're discussing a way to work without code modifications

PR #176, opened 2022-02-19
init: If user already exists in /etc/passwd set props again, do not skip

### LINKED ISSUE #175: [Error] wrong home path ends up in /etc/passwd when requesting custom --home during create
**Describe the bug**  
Incorrect user home path ends up in guest's /etc/passwd when using --home option during distrobox-create  
Somehow user entry in /etc/passwd inside guest is using values mentioned in host's /etc/passwd user entry

**To Reproduce**  
```
    distrobox-create --home /tmp/pathq/ --image ubuntu:latest ubuntu-studio.android.q
    distrobox-enter ubuntu-studio.android.q
```
```
    echo ${HOME}
    # this returns correct home (the one assigned with --home)

    cat /etc/passwd | grep $(whoami)
    # this shows something not right (shows host's user home path)
```

**Expected behavior**  
The path passed using --home should be used to populate (and maintain) user's entry in /etc/passwd

**Desktop**  
Host OS: Fedora 34 (x86_64)
Guest OS: Ubuntu 20.04 (docker ubuntu:latest)

**Additional context**  
The discrepancy causes visible issues. 
For example Android Studio uses user home path set in /etc/passwd to create many directories 
(.gradle, .android etc). This ends up polluting the Host's home because of the bug mentioned above.
It is expected that inside the container the home directory provided in create command will be used.

ennec-e:
This fixes #175 where home field in /etc/passwd diverges from value of $HOME
which then causes unexpected behaviour.
When user is pre-prepopulated due to some reason the old behaviour was to skip useradd.
So path in /etc/passwd could be different than one manually set (ex. when user asks for custom home).

Fix bug by force-setting properties with _usermod_ if user already exists.

### DIFF
--- a/distrobox-init
+++ b/distrobox-init
@@ -461,6 +461,8 @@ if ! grep -q "${container_user_name}" /etc/passwd; then
 		printf "Warning: there was a problem setting up the user\n"
 	fi
 else
+	# This situation is presented when podman or docker already creates the user for us inside container.
+	# We should modify the user's prepopulated shadowfile entry though as per user's active preferences. 
 	if ! usermod \
 		--home "${container_user_home}" \
 		--shell "${SHELL:-"/bin/bash"}" \
--- a/distrobox-init
+++ b/distrobox-init
@@ -448,11 +448,21 @@ fi
 if ! grep -q "${container_user_name}" /etc/group; then
 	groupadd --force --gid "${container_user_gid}" "${container_user_name}"
 fi
-# Let's add our user to the container if the user already exists, just go ahead.
+# Let's add our user to the container. if the user already exists, enforce properties.
 if ! grep -q "${container_user_name}" /etc/passwd; then
 	if ! useradd \
 		--home-dir "${container_user_home}" \
 		--no-create-home \
+		--shell "${SHELL:-"/bin/bash"}" \
+		--uid "${container_user_uid}" \
+		--gid "${container_user_gid}" \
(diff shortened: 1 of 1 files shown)


89luca89:
Hey @ennec-e thanks for the contribution!

Just a couple of comments on this before merging, also does this fix #175 for you?

distrobox-init
```diff
@@ -458,6 +458,16 @@ if ! grep -q "${container_user_name}" /etc/passwd; then
 		--gid "${container_user_gid}" \
 		"${container_user_name}"; then
 
+		printf "Warning: there was a problem setting up the user\n"
+	fi
+else
```
89luca89, line 463: Let's just add a couple of comments here
this situation is presented when podman or docker already creates the user for us, so we find it already present

distrobox-init
```diff
@@ -458,6 +458,16 @@ if ! grep -q "${container_user_name}" /etc/passwd; then
...
+	if ! usermod \
```
89luca89, line 464: Given that we're adding usermod command, it would be nice if you add it to the list of checks in lines 194-197

89luca89 (changes requested):

ennec-e:
> also does this fix #175 for you?

Yes. I've been using distrobox locally with this patch. It does resolve #175 for me.
I'll get on to making the changes you requested now.

PR #178, opened 2022-02-21
install: remove redundant dot

misobarisic:
Previous install output would be like this `Be sure that /usr/local/bin. is in your $PATH environment variable for it to work.`. I have removed the trailing dot right after the path.


89luca89:
Thanks! :+1:

PR #180, opened 2022-02-21
install: Remove permissions check and dest_path creation

nathanchance:
1caa478c52786 tried to improve the writeability check to avoid false
positives, such as the one reported in a DistroWatch.com article.
However, it is still possible to break this check by having the
destination and its parent directory not exist:

```
$ ./install -p $HOME/tmp/directory-does-not-exist/bin
Cannot write into /home/nathan/tmp/directory-does-not-exist/bin, permission denied.
```

The script already uses `install -D`, which will create the target
directory if it does not exist, so trying to create `dest_path`
preemptively only increases the complexity of install and risks
introducing false positive errors like above.

Instead, just check the calls to install and print a message about
potential issues with permissions if they fail.

### DIFF
--- a/install
+++ b/install
@@ -74,34 +74,6 @@ if [ "${verbose}" -ne 0 ]; then
 	set -o xtrace
 fi
 
-# Ensure we can write the dest_path specified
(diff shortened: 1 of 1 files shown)


89luca89:
Thanks @nathanchance 
tested and it works as expected :+1: 

merging, thanks for the contribution!

Issue #181, opened 2022-02-22
[Error] Cannot enter directory having special characters.


nm004:
**Describe the bug**
`distrobox-enter --name sandbox` issues an error when inside directory that has special characters, like white space, parentheses, quotes, etc.

**To Reproduce**

White space:
```
mkdir 'white space'
cd 'white space'
distrobox-enter --name sandbox
```

and you get,
```
Error: no container with name or ID "space" found: no such container

An error occurred
```

Parentheses:
```
mkdir '(parentheses)'
cd '(parentheses)'
distrobox-enter --name sandbox
```
and you get,

```
/var/home/n/.local/bin/distrobox-enter: eval: line 351: syntax error near unexpected token `('

An error occurred
```

**Expected behavior**
Entering container with no errors, like `toolbox enter` does.

**Additional context**
This bug caused by the last `eval ${cmd}` line. Escaping `--workdir` argument may solve this issue.

89luca89:
Fixed by #182

PR #182, opened 2022-02-22
enter: escape --workdir argument #181

nm004:


nm004:
There could be a better approach than escapaing all chars, I think.

distrobox-enter
```diff
@@ -217,8 +217,9 @@ generate_command() {
 	# Entering container using our user and workdir.
 	# Start container from working directory. Else default to home. Else do /.
 	# pass distrobox-enter path, it will be used in the distrobox-export tool.
+	workdir=$(printf "%s" "${PWD:-${HOME:-/}}" | sed 's/./\\&/g')
 	result_command="${result_command}
-		--workdir=${PWD:-${HOME:-"/"}}
+		--workdir=${workdir}
```
89luca89, line 222: Seems to be enough to just quote (escaping) the workdir and it works with spaces and special chars:

`--workdir=\"${PWD:-${HOME:-"/"}}\"`

89luca89 (changes requested):

89luca89:
> There could be a better approach than escapaing all chars, I think.

Hi @nm004 thanks a lot for the contribution
I've commented another way to make the workdir play nicely with special chars, from a quick test it works with the examples you reported on #182 
can you check also? thanks! :+1:

distrobox-enter
89luca89: I'll correct myself, this won't cover 1 use-case, the presence of quotes in the path name. using this should solve our prolems:

```sh
	workdir="$(echo "${PWD:-${HOME:-"/"}}" | sed -e 's/"/\\\"/g' )"
		--workdir=\"${workdir}\"
```

89luca89 (changes requested):

distrobox-enter
nm004: @89luca89 Thank you for your reply.

It works perfectly.

89luca89:
@nm004 thanks, you just need to run a `shfmt -s` on the script before pushing again so that it solves the formatting issue that are blocking the shellcheck action :+1:

nm004:
@89luca89

> @nm004 thanks, you just need to run a `shfmt -s` on the script before pushing again so that it solves the formatting issue that are blocking the shellcheck action +1

oh, ok thx. Is it ok now?

89luca89 (approved):

89luca89:
Yep thanks for the contribution! Merging!

nm004:
Thanks :)

Issue #177, opened 2022-02-21
[Usability] Improve usability for users


89luca89:
As stated out in this review: https://distrowatch.com/weekly.php?issue=20220221#distrobox

There are areas in which Distrobox can improve a lot in documentation, usability and discoverablity.

- [x] Add a way to consult the compatibility table from distrobox itself
- [x] Detect if we're running as sudo and point the user to the documentation if that's the case
- [x] Improve the installer to create the `dest_path` if not exists 
- [x] Improve the installer informing the user to add `dest_path` to their $PATH
- [x] Improve the error messages for missing `container_manager` and point out to the docs in case of missing one
- [x] Improve the error messages in `distrobox rm` when container is running, suggesting the podman/docker command to run to stop it
- [x] Add confirmation prompt to `distrobox-rm`, together with a `--yes` option
- [x] Add `distrobox stop` to the list of commands
- [ ] Integrate `host-exec` in the container?

misobarisic:
>     * Detect if we're running as sudo and point the user to the documentation if that's the case
> 
>     * Improve the installer to create the `dest_path` if not exists
> 
>     * Improve the installer informing the user to add `dest_path` to their $PATH

None of these points are `distrobox` related.

The problems they had with docker are not our fault. Familiarize yourself with docker. `dest_path` is not in $PATH. If you choose one that isn't in $PATH then, yeah, deal with it.

I am not really user friendly, am I?

89luca89:
Ahahah I think it's just good to point them to the correct documentation
I see a lot of issues are opened for running distrobox as sudo, or using docker incorrectly so it doesn't hurt to add a couple of links in the docs (and man pages automatically)
For the installer just adding a `it doesn't exist, should I create it? Y/n` also doesn't hurt and helps a lot the user 
For the $PATH we cannot do much, as adding a dir to a PATH is quite invasive, but printing a message to remind the user about it also it's just a quality of usage improvement that can also lower the number of bugs/issues opened for this stuff

89luca89:
If there are other usability improvement proposals let's just comment here :+1:

nathanchance:
One suggestion I have for `install` would be requiring the installation prefix for `-p`, rather than the `bin` folder. This is more traditional (e.g. `configure --prefix=...` + `make install`). I know that I would personally be a little confused as a new user if I supplied `-p $HOME/distrobox` and ended up with `$HOME/share` because of the `../share` for `man_dest_path`.

[...]

Issue #28, opened 2021-12-11
[Feature] Sandboxed mode


89luca89:
Right now the distrobox's containers are created in privileged mode and share a lot of sensitive host's folder.

This is done because the aim is tight integration with the host, not sandboxing.

---

It would be nice to have an **optional** (see: disabled if not specified)  `--unprivileged` or a `--sandbox` flag in `distrobox-create` to have a more isolated container to work with.

89luca89:
Sandboxed distrobox should **not** integrate with the host as tightly as the normal one

`$HOME` should not be bind mounted
Host's rootfs should be mounted in readonly (or ideally not mounted at all)
Sockets should not be passed to the container

Sandboxed distrobox should **not** use `--privileged` and should **not** use `--ipc host, --pid host`

Thinking of functionality:

- `--sandbox` seems an understandable flag to use
- `--allow-path` could be used to allow some path to be mounted as read-write

89luca89:
Depends on #70

BlauerHunger:
I think only implementing "fully sandboxed+unprivileged" and "fully integrated+privileged" modes are not the way to go, but rather adding options to configure it finer-grained like making only a specified set of devices available, only select sockets, mapping a custom selection of directories or other things possible for Flatpaks using Flatseal. 

Based on that you could then add presets for integrated vs. sandboxed.

89luca89:
Well there is a big difference between distrobox and Flatpak, distrobox is not an application distribution method.
On one hand I agree having some more fine grained support is good
On the other, it's and extreme work to add support to filter specific devices (like, usb pen1 passes, usb pen2 does not pass etc etc)

I was thinking more on the lines of

Sanbox mode:

1 - custom home
2 - you specify which path can the container access
3 - you decide if devices in /dev/ are accessible or not
4 - you decide if you share:
     - network namespace
     - process namespace
     - ipc namespace
     - mount namespace
     - (others)

And on that one can do a couple of profiles (low to high sandboxing)

89luca89:
Adding to before
we should give options to:

- mount in overlay, or not mount at all
    - /home
    - /
    - /dev
- disable additional group mappings: `run.oci.keep_original_groups=0`
- disable stuff like
    - network namespace
    - process namespace
    - ipc namespace
    - mount namespace
    - (others)

ennec-e:
I had some ideas on sandboxing modes. 
This gives overview of the modes I thought of, describing general use cases and overall access to Host. 
[...]

PR #186, opened 2022-02-23
enter: Enter at workdir through child path of '/run/host'

ennec-e:
Despite tight integration, some host workdir paths may not exist in container.
So always drop at workdir through '/run/host' path which is bound to exist.
User can then cd as desired.

This also avoids some confusion. For example:
```
user@host:/etc$
user@host:/etc$ distrobox-enter guest
user@guest:/etc$ 
```
`/etc` on line-1 and `/etc` on line-3 are two different directories. 
It's not very intuitive that the working directory has changed while name remains identical.

Especially when for a different starting path we see different behaviour :
```
user@host:/home/user$
user@host:/home/user$ distrobox-enter guest
user@guest:/home/user$
```
Here the `/home/user` on line-1 and `/home/user` on line-3 _are the same_ directory. 

With this pull request, on enter, user will always drop to the same directory they were in on host
using a path through /run/host subtree.

```
user@host:/etc$
user@host:/etc$ distrobox-enter guest
user@guest:/run/host/etc$
```

### DIFF
--- a/distrobox-enter
+++ b/distrobox-enter
@@ -218,7 +218,7 @@ generate_command() {
 	# Start container from working directory. Else default to home. Else do /.
 	# pass distrobox-enter path, it will be used in the distrobox-export tool.
 	workdir="$(echo "${PWD:-${HOME:-"/"}}" | sed -e 's/"/\\\"/g')"
-	container_available_workdir="$(echo "/run/host/${workdir}" | sed -e 's/"/\\\"/g')"
+	container_available_workdir="/run/host/${workdir}"
 	result_command="${result_command}
 		--workdir=\"${container_available_workdir}\"
 		--env=\"DISTROBOX_ENTER_PATH=${distrobox_enter_path}\""
--- a/distrobox-enter
+++ b/distrobox-enter
@@ -219,9 +219,13 @@ generate_command() {
 	# Since we are entering from host, drop at workdir through '/run/host'
 	# which represents host's root inside container. Any directory on host
 	# even if not explicitly mounted is bound to exist under /run/host.
+	# Since user $HOME is very likely present in container, enter there directly
+	# to avoid confusing the user about shifted paths.
 	# pass distrobox-enter path, it will be used in the distrobox-export tool.
 	workdir="$(echo "${PWD:-${HOME:-"/"}}" | sed -e 's/"/\\\"/g')"
-	workdir="/run/host/${workdir}"
+	if [ -n "${workdir##*"${HOME}"*}" ]; then
+		workdir="/run/host/${workdir}"
+	fi
 	result_command="${result_command}
 		--workdir=\"${workdir}\"
 		--env=\"DISTROBOX_ENTER_PATH=${distrobox_enter_path}\""
--- a/distrobox-enter
+++ b/distrobox-enter
@@ -218,9 +218,10 @@ generate_command() {
 	# Start container from working directory. Else default to home. Else do /.
 	# pass distrobox-enter path, it will be used in the distrobox-export tool.
 	workdir="$(echo "${PWD:-${HOME:-"/"}}" | sed -e 's/"/\\\"/g')"
+	container_available_workdir="$(echo "/run/host/${workdir}" | sed -e 's/"/\\\"/g')"
 	result_command="${result_command}
-		--workdir=\"${workdir}\"
-		--env=\"DISTROBOX_ENTER_PATH=${distrobox_enter_path}"\"
+		--workdir=\"${container_available_workdir}\"
+		--env=\"DISTROBOX_ENTER_PATH=${distrobox_enter_path}\""
 
 	# Loop through all the environment vars
 	# and export them to the container.
--- a/distrobox-enter
+++ b/distrobox-enter
@@ -216,11 +216,14 @@ generate_command() {
 
 	# Entering container using our user and workdir.
 	# Start container from working directory. Else default to home. Else do /.
+	# Since we are entering from host, drop at workdir through '/run/host'
+	# which represents host's root inside container. Any directory on host
+	# even if not explicitly mounted is bound to exist under /run/host.
 	# pass distrobox-enter path, it will be used in the distrobox-export tool.
 	workdir="$(echo "${PWD:-${HOME:-"/"}}" | sed -e 's/"/\\\"/g')"
-	container_available_workdir="/run/host/${workdir}"
+	workdir="/run/host/${workdir}"
 	result_command="${result_command}
-		--workdir=\"${container_available_workdir}\"
+		--workdir=\"${workdir}\"
 		--env=\"DISTROBOX_ENTER_PATH=${distrobox_enter_path}\""
 
 	# Loop through all the environment vars


ennec-e:
relate #177

89luca89:
Mmh it seems that current behavior is consistent with toolbx 
I don't dislike your behavior but I'd like for it to be 'run/host' only if it is not present in the container
Otherwise this will apply even on home which could be quite confusing 

Let's discuss this in #177 ?

ennec-e:
> Otherwise this will apply even on home which could be quite confusing

Quicknote: I don't exactly know how (magic), But on my computer this does seem to let any subdirectory of user home to automagically resolve to their natural path. 

example:

```
user@host:/home/user/subdir$
user@host:/home/user/subdir$ distrobox-enter guest
user@guest:/home/user/subdir$
```
Seems like podman or something along the way understands that */run/host/home/user* == */home/user* inside container
and so on enter I just see /home/user/subdir as the current working dir.

Similarly for /dev and /tmp
Directries like /var or /etc on the other hand do show up as /run/host/var and /run/host/etc respectively.

I suppose this is ideal?

---
Please do see if you can replicate these results. I have only tested this on one config (Fedora Host + podman + Ubuntu Guest)

89luca89:
I tested this and verified and it works as expected :smile: 
Just a couple of comments and it's ok to merge for me

distrobox-enter
```diff
@@ -218,9 +218,10 @@ generate_command() {
 	# Start container from working directory. Else default to home. Else do /.
 	# pass distrobox-enter path, it will be used in the distrobox-export tool.
 	workdir="$(echo "${PWD:-${HOME:-"/"}}" | sed -e 's/"/\\\"/g')"
+	container_available_workdir="/run/host/${workdir}"
```
89luca89, line 221: let's reuse `workdir` variable, no need for a new one

89luca89, line 221: also let's explain the behavior of this briefly in a comment so that in future we still have idea on how this works :+1:

89luca89 (changes requested):

89luca89:
> I tested this and verified and it works as expected smile Just a couple of comments and it's ok to merge for me

mmmh after some more testing I saw that using it from an existing shell, works as expected
but it does not when using a direct shell (for example using a gnome-terminal profile)

What about simply checking if `home` is in the workdir path?

```sh
workdir="$(echo "${PWD:-${HOME:-"/"}}" | sed -e 's/"/\\\"/g')"
if [ -n "${workdir##*${HOME}*}" ]; then
                 workdir="/run/host${workdir}"
fi
```

This is  a quite naive check but should ensure that HOME (of the host) is always respected 
what do you think?

ennec-e:
> mmmh after some more testing I saw that using it from an existing shell, works as expected
> but it does not when using a direct shell (for example using a gnome-terminal profile)

Thank you for trying it out. What is the unexpected behaviour that you are seeing above? Can you please share an example output or screenshot; I am unable to grok the distinctions between _direct_ shells vs _existing_ shells 😅
Also can you test if this is also true of /tmp and /dev ?

89luca89:
> > mmmh after some more testing I saw that using it from an existing shell, works as expected
> > but it does not when using a direct shell (for example using a gnome-terminal profile)
> 
> Thank you for trying it out. What is the unexpected behaviour that you are seeing above? Can you please share an example output or screenshot; I am unable to grok the distinctions between _direct_ shells vs _existing_ shells sweat_smile Also can you test if this is also true of /tmp and /dev ?

Simply put if I'm on the host and do `distrobox enter` it works as you expect:

```sh
luca-linux@thinkpad-l14:~$ distrobox enter 
luca-linux@fedora-toolbox-35:~$ 
```

But if I use for example a gnome-terminal profile like this:
https://user-images.githubusercontent.com/598882/146691460-b8a5bb0a-a83d-4e32-abd0-4a0ff9f50eb7.png

And then launch the terminal with that profile, I find myself in:

```sh
luca-linux@fedora-rawhide:/run/host/home/luca-linux$ 
```
---

You can check in my previous comment a solution that should be ok and works correctly also with custom homes

ennec-e:
> What about simply checking if `home` is in the workdir path?

This might become inconsistent when user has chosen a custom home. How does toolbx deal with custom homes?
Also, one of the future considerations I had in mind with this change was about light sandboxing #28. When using a custom home, the host home does not even _need_ to be bind mounted by default. If it isn't mounted (like [here](https://github.com/ennec-e/distrobox/commit/437ed48c09491968df6ad26f638008474b770c62) then it is just another un-special host exclusive directory that can still be accessed through /run/host. And so shouldn't warrant special treatment.

Btw, since a home reaching path going from /run/host is exactly the _same directory_ as a normal user home subpath, just bind mounted at different location it shouldn't really matter right? Any file-system (create, edit etc) changes will have the same effect. The only difference is the initial drop-in **path** (the directory remains consistent). A simple `cd` would move user to regular path. 

Btw, a simple trick would be for the check to happen _inside_ the container. Could this be done as part of init?
A simple init-hook to move user to regular home subdir path if they are at the subdir through /run/host.

ennec-e:
> You can check in my previous comment a solution that should be ok and works correctly also with custom homes

Cool!,  Am trying this now.

ennec-e:
> But if I use for example a gnome-terminal profile like this: https://user-images.githubusercontent.com/598882/146691460-b8a5bb0a-a83d-4e32-abd0-4a0ff9f50eb7.png
> 
> And then launch the terminal with that profile, I find myself in:
> 
> ```shell
> luca-linux@fedora-rawhide:/run/host/home/luca-linux$ 
> ```
> 
> You can check in my previous comment a solution that should be ok and works correctly also with custom homes

Am able to replicate your finding. Unfortunately I see this with the solution you mentioned too (for custom home). 
I think it's something to do with the way `tmux` initialises vs `bash -l` (The login shell might be doing some magic to translate the paths if it detects directory is are under user home)

ennec-e:
Aaah! Of course. Bash I believe uses $PWD to check if it can truncate $HOME part in the current directory to `~`
Since we're passing the environment variable $PWD unaltered the magic works.
 
Another program like tmux may not be doing that directly.

distrobox-enter
ennec-e: Done!

ennec-e: Done.

89luca89:
> This might become inconsistent when user has chosen a custom home.

Yep works fine :+1:

>  How does toolbx deal with custom homes?

It does not support them

> Also, one of the future considerations I had in mind with this change was about light sandboxing https://github.com/89luca89/distrobox/issues/28. When using a custom home, the host home does not even need to be bind mounted by default. If it isn't mounted (like [here](https://github.com/ennec-e/distrobox/commit/437ed48c09491968df6ad26f638008474b770c62)) then it is just another un-special host exclusive directory that can still be accessed through /run/host. And so shouldn't warrant special treatment.

Yea about #28 I think we will not mount /run/host at all, so we'll just skip workdir at that time? 

> Btw, since a home reaching path going from /run/host is exactly the same directory as a normal user home subpath, just bind mounted at different location it shouldn't really matter right? Any file-system (create, edit etc) changes will have the same effect. The only difference is the initial drop-in path (the directory remains consistent). A simple cd would move user to regular path.

It's not what the user expects, and it will break current behavior (and branch away from how farious `toolbox` tools work)

distrobox-enter
```diff
@@ -216,11 +216,15 @@ generate_command() {
 
 	# Entering container using our user and workdir.
 	# Start container from working directory. Else default to home. Else do /.
+	# Since we are entering from host, drop at workdir through '/run/host'
+	# which represents host's root inside container. Any directory on host
+	# even if not explicitly mounted is bound to exist under /run/host.
 	# pass distrobox-enter path, it will be used in the distrobox-export tool.
 	workdir="$(echo "${PWD:-${HOME:-"/"}}" | sed -e 's/"/\\\"/g')"
+	workdir="/run/host/${workdir}"
```
89luca89, line 224: Let's continue discussing this solution:

```sh
workdir="$(echo "${PWD:-${HOME:-"/"}}" | sed -e 's/"/\\\"/g')"
if [ -n "${workdir##*${HOME}*}" ]; then
                 workdir="/run/host${workdir}"
fi
```

89luca89 (changes requested):

distrobox-enter
ennec-e: I've been testing this soln. And it does make the exclusion explicit for Host's $HOME subdirectories. 👍
But I must note it does not have any effect on custom homes' subdirectories (that are not under Host $HOME).

See following commit for tests output.

ennec-e: ## Test Setup

```
[ennec-e@fedora distrobox]$ git diff
diff --git a/distrobox-enter b/distrobox-enter
index b08451f..e5628c7 100755
--- a/distrobox-enter
+++ b/distrobox-enter
@@ -218,6 +218,9 @@ generate_command() {
        # Start container from working directory. Else default to home. Else do /.
        workdir="$(echo "${PWD:-${HOME:-"/"}}" | sed -e 's/"/\\\"/g')"
+       if [ -n "${workdir##*${HOME}*}" ]; then
+               workdir="/run/host${workdir}"
+       fi
                --workdir=\"${workdir}\"
                --env=\"DISTROBOX_ENTER_PATH=${distrobox_enter_path}"\"
[ennec-e@fedora distrobox]$ distrobox-create -i ubuntu:latest -H /tmp/test-exception-custom-home
```
Also install tmux in guest.

## Test Scenarios

### Simple enter from host $HOME subdir

```
[ennec-e@fedora distrobox]$ distrobox-enter ubuntu-latest -- bash -l
ennec-e@ubuntu-latest:/home/ennec-e/Code/distrobox$ 
```
👍

### Tmux enter from host $HOME subdir
```
[ennec-e@fedora distrobox]$ distrobox-enter ubuntu-latest -- tmux
---
ennec-e@ubuntu-latest:/home/ennec-e/Code/distrobox$ 
---
```
👍

### Tmux enter from custom home subdir
```
[ennec-e@fedora test-exception-custom-home]$ distrobox-enter ubuntu-latest -- tmux
---
ennec-e@ubuntu-latest:/run/host/tmp/test-exception-custom-home$ 
---
```
❗😵‍💫

### Simple enter from custom home subdir
 
```
[ennec-e@fedora test-exception-custom-home]$ distrobox-enter ubuntu-latest -- bash -l
ennec-e@ubuntu-latest:~$ 
```
👍

ennec-e: The fourth case works as usual because of bash magic as described in a [previous comment]. (Same applies to case 1)
The solution presented does explicitly exclude, _host's_ $HOME so it does work for case 1 and 2. 

At enter time distrobox-enter does not know about _guest's_ custom $HOME dir, so it may not be excludeable from there. 
The solution for 3rd case I think would have to run somewhere inside the guest or from a script that is aware of custom home value. 

I personally don't really mind at all. case 3 might be some sort of an edge case anyway (custom home _outside_ Host home + custom shell). And perhaps can be left documented.

Otherwise this looks good enough : ) It makes this patch consistent with toolbox behaviour. So great!

[previous comment]:  https://github.com/89luca89/distrobox/pull/186#issuecomment-1050421175

89luca89: Technically case 3 (custom home) is not thought to enter in the custom home by default, but still preserve the workdir you are.  It is there just to have a custom home, so I think it's an edge case that we can work out later :+1:

ennec-e: Have added a commit with the requested change made. Please review.

89luca89:
Perfect! Merging! Thanks for the contribution! :tada:

PR #191, opened 2022-02-27
FIx fail on installation dirs not existing

pepa65:


nathanchance:
`install -D` should take care of creating `dest_path` and `man_dest_path` if they do not exist. What is the failure you are seeing?

misobarisic:
Please elaborate on the failing condition. As previously said, `install -D` should take care of creating the directory. Are you, for some reason, missing the `install` coreutil?

pepa65:
> As previously said, `install -D` should take care of creating the directory.
Sorry, I did not use -D. On the other hand, why not just always create missing directories? I was going off this: https://distrowatch.com/weekly.php?issue=current#distrobox
But if the code already is creating the directories, I am sorry for the noise, and please disregard this pull.

misobarisic:
> > As previously said, `install -D` should take care of creating the directory.
> > Sorry, I did not use -D. On the other hand, why not just always create missing directories? I was going off this: https://distrowatch.com/weekly.php?issue=current#distrobox
> > But if the code already is creating the directories, I am sorry for the noise, and please disregard this pull.

The install script has been changed since the publishing of distrowatch's article. One notable change is #180.

PR #197, opened 2022-03-09
Fix grep match of very short user or group names.

lynxnot:
my host linux user is a single character (aka "d"), `distrobox-init` was failing to create user and group because the character `d` matches many users and groups. This fix ensure very short usernames are handled correctly. 

I've tested it only in a Centos 7 container, but it should work on other distros.


89luca89:
Talking about corner cases :joy: this makes a lot sense and looks good to me!
Merging and thanks for the contribution!

PR #199, opened 2022-03-11
Feature/compatibility test

89luca89:
This is a pretty simplistic test, for now it implements only a `whoami` 
This will ensure that at least the `distrobox-init` process does not error out on all the distributions supported.
This will also ensure that `distrobox-rm`, `distrobox-stop` and `distrobox-list` work as expected.

The source of truth is the compatibility table in the documentation, so we ensure it is always up to date and tested fully.

For now the test is performed manually as it is quite intensive (more than 50 images to test), we will also need to find a service that can host such a large and time consuming test

Todo list:

- [ ] find other commands to test in the exec
- [ ] find a service where we can automate this

### DIFF
--- a/tests/compatibility-test
+++ b/tests/compatibility-test
@@ -66,7 +66,9 @@ if [ "${verbose}" -ne 0 ]; then
 fi
 
 image_list=$(grep -E "docker.io|quay.io|ghcr|registry.|ecr." "$(dirname "${0}")"/../docs/compatibility.md | cut -d'|' -f 4 | sed 's/<br>/\n/g' | tr -d ' ' | tail -n +2)
-image_total=$(grep -E "docker.io|quay.io|ghcr|registry.|ecr." "$(dirname "${0}")"/../docs/compatibility.md | cut -d'|' -f 4 | sed 's/<br>/\n/g' | tr -d ' ' | tail -n +2 | wc -l)
+image_total=$(echo "${image_list}" | wc -l)
+
+exit 1
 
 index=0
 for image in ${image_list}; do
--- a/tests/compatibility-test
+++ b/tests/compatibility-test
@@ -1,0 +1,84 @@
+#!/bin/sh
+
+trap '[ "$?" -ne 0 ] && printf "\nAn error occurred\n"' EXIT
+
+verbose=0
+image_index=0
+# Dont' run this command as sudo.
+if [ "$(id -u)" -eq 0 ]; then
+	printf >&2 "Running %s as sudo is not supported.\n" "${0}"
+	printf >&2 "Please check the documentation on:\n"
+	printf >&2 "\tman distrobox-compatibility\t"
+	printf >&2 "or consult the documentation page on:\n"
+	printf >&2 "\thttps://github.com/89luca89/distrobox/blob/main/docs/compatibility.md\n"
+	exit 1
+fi
+
+# Print usage to stdout.
+# Arguments:
+#   None
+# Outputs:
+#   print usage with examples.
+show_help() {
+	cat <<EOF
+Usage:
+
+	compatibility-test [--index X]
+
+Options:
+
+	--index/i:		from which index image to start
+	--help/-h:		show this message
+	--verbose/-v:		show more verbosity
+
+EOF
+}
+
+# Parse arguments
+while :; do
+	case $1 in
+	-h | --help)
+		# Call a "show_help" function to display a synopsis, then exit.
+		show_help
+		exit 0
+		;;
+	-v | --verbose)
+		shift
+		verbose=1
+		;;
+	-i | --index)
+		if [ -n "$2" ]; then
+			image_index="$2"
+			shift
+			shift
+		fi
+		;;
+	*) # Default case: If no more options then break out of the loop.
+		break ;;
+	esac
+done
+
+set -o errexit
+set -o nounset
+# set verbosity
+if [ "${verbose}" -ne 0 ]; then
+	set -o xtrace
+fi
+
+image_list=$(grep -E "docker.io|quay.io|ghcr|registry.|ecr." "$(dirname "${0}")"/../docs/compatibility.md | cut -d'|' -f 4 | sed 's/<br>/\n/g' | tr -d ' ' | tail -n +2)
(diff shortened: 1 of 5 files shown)


tests/compatibility-test
```diff
@@ -0,0 +1,84 @@
...
+
+set -o errexit
+set -o nounset
+# set verbosity
+if [ "${verbose}" -ne 0 ]; then
+	set -o xtrace
+fi
+
+image_list=$(grep -E "docker.io|quay.io|ghcr|registry.|ecr." "$(dirname "${0}")"/../docs/compatibility.md | cut -d'|' -f 4 | sed 's/<br>/\n/g' | tr -d ' ' | tail -n +2)
+image_total=$(grep -E "docker.io|quay.io|ghcr|registry.|ecr." "$(dirname "${0}")"/../docs/compatibility.md | cut -d'|' -f 4 | sed 's/<br>/\n/g' | tr -d ' ' | tail -n +2 | wc -l)
```
nathanchance, line 69: `image_total=$(echo "$image_list" | wc -l)`

would likely save you some cycles :)

89luca89: he he you're right :smile:

misobarisic:
Making a custom github action is always an option. Not an ideal solution but still worthwhile.

89luca89:
> Making a custom github action is always an option. Not an ideal solution but still worthwhile.

I don't think they allow a workload this heavy sadly, I was using travisCI for another project here, but it is not free anymore

nathanchance:
> > Making a custom github action is always an option. Not an ideal solution but still worthwhile.
> 
> I don't think they allow a workload this heavy sadly, I was using travisCI for another project here, but it is not free anymore

GitHub Actions allows a single job in a workflow to run for six hours and a whole workflow to run for 72 hours:

https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration

I would not expect this whole set of tests to take longer than six hours (unless I am vastly underestimating how long each run would take), so you could just run this script as a single job.

However, it would probably be quicker to have each distro be a matrix item so that each distro test could be run in parallel:

https://docs.github.com/en/actions/using-jobs/using-a-build-matrix-for-your-jobs

I don’t think that would be too hard to script. I am fairly well versed in GitHub Actions, so I’d be happy to help answer questions or hack on this at some point over the next week if I have some downtime.

89luca89:
> > > Making a custom github action is always an option. Not an ideal solution but still worthwhile.
> > 
> > 
> > I don't think they allow a workload this heavy sadly, I was using travisCI for another project here, but it is not free anymore
> 
> GitHub Actions allows a single job in a workflow to run for six hours and a whole workflow to run for 72 hours:
> 
> https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration
> 
> I would not expect this whole set of tests to take longer than six hours (unless I am vastly underestimating how long each run would take), so you could just run this script as a single job.
> 
> However, it would probably be quicker to have each distro be a matrix item so that each distro test could be run in parallel:
> 
> https://docs.github.com/en/actions/using-jobs/using-a-build-matrix-for-your-jobs
> 
> I don’t think that would be too hard to script. I am fairly well versed in GitHub Actions, so I’d be happy to help answer questions or hack on this at some point over the next week if I have some downtime.

Yep i'm going to make them, I'm quite practical with them too 😄 
Probably will make them only on tag releases or something like this, I'm not too worried about the execution time, and we can just parallelize by index with this script (with 8 jobs it is about 7 distros per job)

I'm more worried about docker hub's quota limits 😅  probably will need in future to use some sort of mirrors 
If we're going to test both podman and docker it is going to be more than 100 pulls per-test

nathanchance:
> Yep i'm going to make them, I'm quite practical with them too 😄 Probably will make them only on tag releases or something like this, I'm not too worried about the execution time, and we can just parallelize by index with this script (with 8 jobs it is about 7 distros per job)

Given how few limits there are on GitHub Actions usage for open source project, you might just consider running it on every commit and pull until there is a problem :)

> I'm more worried about docker hub's quota limits 😅 probably will need in future to use some sort of mirrors If we're going to test both podman and docker it is going to be more than 100 pulls per-test

You could potentially look into mirroring images to `ghcr.io`, as I don't think there is any limit for pulling from there.

You could even consider taking it a step further and look into using `ghcr.io` to provide `distrobox` tailored images through GitHub Actions. I know half of the appeal of `distrobox` is that it should work with any container image but it might be nice to provide images that already have all the dependencies installed.

misobarisic:
> > Yep i'm going to make them, I'm quite practical with them too 😄 Probably will make them only on tag releases or something like this, I'm not too worried about the execution time, and we can just parallelize by index with this script (with 8 jobs it is about 7 distros per job)
> 
> Given how few limits there are on GitHub Actions usage for open source project, you might just consider running it on every commit and pull until there is a problem :)

What about running it on every commit with changes to distrobox* files?

89luca89:
> What about running it on every commit with changes to distrobox* files?

Let's see how much it takes but I suspect it would be a bit too much, maybe on PR && distrobox-init file (which is in the end the focus of the compatibility test)

 > Given how few limits there are on GitHub Actions usage for open source project, you might just consider running it on every commit and pull until there is a problem :)
 
 Yea in the next days I'll modify the script to accept not only a start-index, but also an end-index so we can parallelize easier

> You could potentially look into mirroring images to ghcr.io, as I don't think there is any limit for pulling from there.

Cool I'll check if every one is available there we may just pull from that :+1:

> You could even consider taking it a step further and look into using ghcr.io to provide distrobox tailored images through GitHub Actions. I know half of the appeal of distrobox is that it should work with any container image but it might be nice to provide images that already have all the dependencies installed.

Well yes there are already OpenSUSE and Fedora toolbox images, but the point is that if I'll provide them, I'll also need to maintain them which is (possibly) quite the effort in time (being over 50 of them)

89luca89:
Cool @nathanchance @misobarisic added the github action instead of the script, it goes in like 5 minutes so it's pretty doable to be done on push on master and on PRs
Leaving the script in case anyone wants to test manually or one day we run away from github :smile:

89luca89:
Also now we test some enter/exec combinations and exporting on all distros, as a conseguence the compatibility list is updated :smile:

PR #205, opened 2022-03-16
Support LDAP/AD mail usernames

89luca89:
sometimes usernames can be in mail form or containing special chars,
this commit will force the creation of the user and group even if it has
invalid names.

In this process we also remove a dependency (passwd) and we use chpasswd
instead.


89luca89:
Seems like the `--badname` flag for `useradd` is not supported on clearlinux
Will need to find a solutions for this.

89luca89:
To support systems where usermod flags are not supported, I've added a manual setup of the user.
On some systems the root user is not initialized so `chpasswd` will fail, i've added back `passwd` for the initial root initialization.

PR #209, opened 2022-03-18
fix: use container's $PATH when entering

gbritton:
The PATH fixes in 6b60c73999c5147f732a1f9933d0b52d7d090161 use the
host's $PATH instead of the $PATH defined in the container.  This
change fixes that by inspecting the container and parsing $PATH
out of its environment.  The inspect was already being done so
this shouldn't be a signficiant performance impact.  Future
features may desire further information out of the inspect
invocation and the parsing method can be extended.

The container's $PATH is merged with the host $PATH and the
standard set of binary paths as the existing logic uses.

### DIFF
--- a/distrobox-enter
+++ b/distrobox-enter
@@ -243,11 +243,21 @@ generate_command() {
 		result_command="${result_command} --env=\"${i}\""
 	done
 
+	# Start with the $PATH set in the container's config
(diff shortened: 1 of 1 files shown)


89luca89:
Great stuff! Thanks for this contribution!

PR #224, opened 2022-04-01
init: Allow all package manager checks to run

nathanchance:
The problem is that a user might have the first two directories on Arch
Linux if they have installed the `rpm` and `dpkg` commands respectively.
While that seems odd on the surface, these tools are useful for building
packages for `rpm` or `dpkg`-based distributions on Arch Linux, such as the
Linux kernel, which has `bindeb-pkg` and `binrpm-pkg` targets to build a
`.deb` and `.rpm` package respectively, which does not require a Debian or
Fedora host.

As a result of the `rpm` and `dpkg` directories being checked first, the
third else if clause never gets executed in the scenario described
above. Move the libalpm check to the beginning so that `pacman`-based
images benefit from the new change regardless of `rpm` and `dpkg` being
installed. While here, add a comment similar to the other checks.

`distrobox-init` checks for three directories to make decisions based on
what package manager is being used in the following order:

```
/usr/lib/rpm/macros.d
/etc/dpkg/dpkg.cfg.d
/usr/share/libalpm/scripts
```

The problem is that a user might have the first two directories on Arch
Linux if they have installed the `rpm-tools` and `dpkg` packages
respectively.  While that seems odd on the surface, these tools are
useful for building packages for `rpm` or `dpkg`-based distributions on
Arch Linux, such as the Linux kernel, which has `bindeb-pkg` and
`binrpm-pkg` targets to build a `.deb` and `.rpm` package respectively,
which does not require a Debian or Fedora host. As a result of the `rpm`
and `dpkg` directories being checked first, the third else if clause
never gets executed in the scenario described above.

Instead of making all these if statements mutually exclusive, allow them
all to be run, which should not impact anything during runtime, as only
one package manager will actually be used within the container. Arch
Linux explicitly notes this when installing the `dpkg` and `rpm-tools`
packages:

https://github.com/archlinux/svntogit-community/blob/55c1554978e24cdc5c53d1265a211e5c9dc72a0b/trunk/dpkg.install
https://github.com/archlinux/svntogit-community/blob/13188cb24c4c2ee318f1efad3281e44c6516fde1/trunk/rpm-tools.install

While here, add a comment similar to the other checks.

### DIFF
--- a/distrobox-init
+++ b/distrobox-init
@@ -464,9 +464,11 @@ if [ -d "/usr/lib/rpm/macros.d/" ]; then
 %_netsharedpath ${net_mounts}
 EOF
 
+fi
+
 # In case of an DEB distro, we can specify that our bind_mount directories
 # have to be ignored. This prevents conflicts during package installations.
-elif [ -d "/etc/dpkg/dpkg.cfg.d/" ]; then
+if [ -d "/etc/dpkg/dpkg.cfg.d/" ]; then
 	# Loop through all the environment vars
 	# and export them to the container.
 	set +o xtrace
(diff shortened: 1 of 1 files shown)


nathanchance:
I do not really see what went wrong with the Alma Linux container, I can see that it still hits the RPM check, so this change should be a no-op?

89luca89:
> I do not really see what went wrong with the Alma Linux container, I can see that it still hits the RPM check, so this change should be a no-op?

No worries sometimes the github action goes buts probably 100 jobs is not healthy for then :joy:

Anyway on the PR I never thought of that use case, maybe we can replace if/elif with a series of separated ifs?
This way we also support eventual debs on rpms and so on what do you think?

nathanchance:
Sure, I can make that change and we will see what breaks :)

89luca89:
Perfect nothing broke :joy: 
At least this way we don't have to worry about orders 🎉

Issue #70, opened 2021-12-29
[Feature] Create custom HOME for containers


89luca89:
This is needed for #28 

We should be able to use something like `--home-dir /path/i/like` so that a container can use a specific directory as HOME, instead of the $HOME of the host.

This can be needed for various reasons, from isolation (as per #28) or to keep clean the host's HOME when we try software or build stuff.

Original HOME will anyway be reachable on /run/host/ so we will be still able to reach for files and such

89luca89:
Update: we need to keep an eye also on `distrobox-enter` as it will set the HOME environment value and will thwart the value set in `distrobox-create`

A simple workaround could be that we enforce in `distrobox-init` the value of HOME passed to the init via the `--home` flag we already have and use

yilkalargaw:
Yes you might also need to keep eyes files like .bashrc, .bash_profile and .profile. The need to be autogenerated by the guest distros default mechanism and they need to persistent between sessions. For some reason podman might reset them if you use the approach used by [tlbx](https://gitlab.com/uppercat/tlbx).

89luca89:
> Yes you might also need to keep eyes files like .bashrc, .bash_profile and .profile. The need to be autogenerated by the guest distros default mechanism and they need to persistent between sessions. For some reason podman might reset them if you use the approach used by [tlbx](https://gitlab.com/uppercat/tlbx).

Yep I was working on this lightly now, and the only thing remaining is really this, having those default files ready.

For now the adjustments needed are:

- **distrobox-create:** add the flag that sets the already present `$container_user_home`, be sure to set the HOME env variable contextually
- **distrobox-enter:** ignore the HOME env variable, as it is set at `distrobox-create` time now

And it works as it is, we need to sort out the shell default files

89luca89:
@yilkalargaw I've opened a pull request if you want to test it

`distrobox-create --image alpine --name alpine-test --home /tmp/`

and

`distrobox-enter --name alpine-test`

This is still not having the default profile/bashrc file done

89luca89:
> For some reason podman might reset them if you use the approach used by [tlbx](https://gitlab.com/uppercat/tlbx).

I didn't know this project, neat :smile: 

Anyway it's been resetting because each start it does:

`	su "$user" sh -c "cp -p /etc/skel/.* /home/$user"`

Which overwrites the files with the default ones in skel.

The skel directory is not a ubiquitous default, so i'd say i would set like this in the `distrobox-init`

```shell
[...]

PR #230, opened 2022-04-08
[DRAFT] Don't mount host $HOME if custom home specified

gunnarx:
This is more of a discussion item still but I felt some code might be easier to discuss around.  Let me know if you want a discussion Issue instead.

----

I prefer that $HOME isn't mounted if a custom home was specified.

[Choosing the --workdir seems not perfect](https://github.com/89luca89/distrobox/blob/bf7c0ba556f244c33d8d81156c54e1e0d8f79b39/distrobox-enter#L249) (in both **main**, and in this PR), but in my view it should at least not be host $HOME by default, if another HOME has been specified...

Ref: 
- https://github.com/89luca89/distrobox/issues/70

...noted one of the problems that can still appear, that --[workdir set to a non-existing location will cause an error](https://github.com/89luca89/distrobox/issues/70#issuecomment-1002889436).

Sandboxing apps and isolating them from personal files in $HOME is actually my main reason for trying out distrobox, but it's [not the primary direction of the project](https://github.com/89luca89/distrobox#aims). 

I feel would be good to protect the host's `$HOME` dir from applications in the container **_if_** a custom home was selected.  Not mounting it at `/home/<user>` is [a good start](https://github.com/gunnarx/distrobox/blob/b532d4f4f1ae569cf2138d2fbbbe617d14d59a08/distrobox-create#L423), but of course not ultimately secure since it is still at `/run/host/home/<user>` as things stand now.

Ultimately, this is only a step towards Sandbox implementation which is an open issue, but I wonder if you also think changing to this behavior now is logical.

This behavior is at least what I expected, and the modification is what I use now.   There's more tweaking possible probably so consider this a draft PR for discussion?

Ref:  
- https://github.com/89luca89/distrobox/issues/28

### DIFF
--- a/distrobox-create
+++ b/distrobox-create
@@ -379,8 +379,6 @@ generate_command() {
 	# Mount also the distrobox-init utility as the container entrypoint.
 	result_command="${result_command}
 		--env \"SHELL=${SHELL:-"/bin/bash"}\"
-		--env \"HOME=${container_user_home}\"
-		--volume \"${container_user_home}\":\"${container_user_home}\":rslave
 		--volume \"${distrobox_entrypoint_path}\":/usr/bin/entrypoint:ro
 		--volume \"${distrobox_export_path}\":/usr/bin/distrobox-export:ro
 		--volume /:/run/host:rslave
@@ -421,13 +419,17 @@ generate_command() {
 
 	# If we have a custom home to use,
 	#	1- override the HOME env variable
-	#	2- expor the DISTROBOX_HOST_HOME env variable pointing to original HOME
-	# 	3- mount the custom home inside the container.
+	# 	2- mount the custom home inside the container.
 	if [ -n "${container_user_custom_home}" ]; then
 		result_command="${result_command}
 		--env \"HOME=${container_user_custom_home}\"
 		--env \"DISTROBOX_HOST_HOME=${container_user_home}\"
 		--volume ${container_user_custom_home}:${container_user_custom_home}:rslave"
+	else
+		# No custom home -> mount original home dir from host
+		result_command="${result_command}
+		--env \"HOME=${container_user_home}\"
+		--volume \"${container_user_home}\":\"${container_user_home}\":rslave"
 	fi
 
 	# Mount also the /var/home dir on ostree based systems
--- a/distrobox-enter
+++ b/distrobox-enter
@@ -251,11 +251,23 @@ generate_command() {
 	# Since we are entering from host, drop at workdir through '/run/host'
 	# which represents host's root inside container. Any directory on host
 	# even if not explicitly mounted is bound to exist under /run/host.
-	# Since user $HOME is very likely present in container, enter there directly
-	# to avoid confusing the user about shifted paths.
+	# $HOME variable is very likely present in container.  It is either equal
+	# to the user's normal $HOME on host, or a custom dir specified when the
+	# container was created.
 	# pass distrobox-enter path, it will be used in the distrobox-export tool.
-	workdir="$(echo "${PWD:-${HOME:-"/"}}" | sed -e 's/"/\\\"/g')"
-	if [ -n "${workdir##*"${HOME}"*}" ]; then
+
+	# Find $HOME as it is defined in the container (could be a custom home)
+	local home=$(${container_manager} inspect --type container --format='{{.Config.Env}}' "${container_name}" \
+		| sed 's/ /\n/g' | awk '/^HOME=/' | cut -d= -f 2)
+
+	# Custom home specified?  Let's then start there.
+	if [ "$home" != "$HOME" ] ; then
+		workdir="$(echo "$home" | sed -e 's/"/\\\"/g')"
+	else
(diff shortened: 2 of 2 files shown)


distrobox-create
```diff
@@ -379,8 +379,6 @@ generate_command() {
 	# Mount also the distrobox-init utility as the container entrypoint.
 	result_command="${result_command}
 		--env \"SHELL=${SHELL:-"/bin/bash"}\"
-		--env \"HOME=${container_user_home}\"
-		--volume \"${container_user_home}\":\"${container_user_home}\":rslave
```
gunnarx, line 383: Both of these can be done later, since there is a new conditional following.  HOME will not be set and then set again (overridden) later.

distrobox-enter
```diff
@@ -251,11 +251,23 @@ generate_command() {
...
-	# to avoid confusing the user about shifted paths.
+	# $HOME variable is very likely present in container.  It is either equal
+	# to the user's normal $HOME on host, or a custom dir specified when the
+	# container was created.
 	# pass distrobox-enter path, it will be used in the distrobox-export tool.
-	workdir="$(echo "${PWD:-${HOME:-"/"}}" | sed -e 's/"/\\\"/g')"
-	if [ -n "${workdir##*"${HOME}"*}" ]; then
+
+	# Find $HOME as it is defined in the container (could be a custom home)
+	local home=$(${container_manager} inspect --type container --format='{{.Config.Env}}' "${container_name}" \
```
gunnarx, line 260: I believe it's necessary to look into the container to know the custom --home directory?

distrobox-enter
```diff
@@ -251,11 +251,23 @@ generate_command() {
...
+		| sed 's/ /\n/g' | awk '/^HOME=/' | cut -d= -f 2)
+
+	# Custom home specified?  Let's then start there.
+	if [ "$home" != "$HOME" ] ; then
+		workdir="$(echo "$home" | sed -e 's/"/\\\"/g')"
+	else
+		workdir="$(echo "${PWD:-${home:-"/"}}" | sed -e 's/"/\\\"/g')"
```
gunnarx, line 267: The determination of workdir still feels a bit suspect (both here and in the original, IMHO).  Let's discuss it?

NOTE in these two lines I have used the previous expressions, but using the lower-case $home that represents either $HOME or the custom home inside the container.

89luca89:
Hey @gunnarx thanks for the discussion!

So thinking of it, in #28 we were discussing the introduction of some pre-made profiles for sandboxing (with manual flags if you want to do things other ways)

What about instead of this, simply implement a `--no-host-home` flag?
This could be useful both for custom homes AND on without them (eg you just want a container without the home in common)

gunnarx:
> What about instead of this, simply implement a --no-host-home flag?

Yeah, although I feel it would be logical to not map host-home when a custom home is provided.  
But **--no-host-home** is more flexible. And to be backward compatible I guess it needs to be this instead of the opposite positive logic (**--map-host-home**, otherwise not).

No matter what the solution, it seems tricky to decide on an appropriate --workdir that never fails.

gunnarx:
Assuming this gets reworked, how do you feel about the [lookup of the custom home in the container](https://github.com/gunnarx/distrobox/blob/dont_mount_host_home/distrobox-enter#L259) to use that as the workdir?

89luca89:
> Yeah, although I feel it would be logical to not map host-home when a custom home is provided.

Well it depends on how you view the use of pet containers, coming from toolbox (which is the pioneer really of this type of approach) you have some seamless transition between container and host

Custom homes were born because a lot of tools use $HOME to litter with their dots (eg, npm, pip, various app etc) so to keep a clean home on the host, custom $HOME is a solution
Or also to have different dotfiles between host and guest

This new flag (`--no-host-home`, because as you rightly said a positive one would break compatibility) would have the pros of not being forcefully tied to the custom home (eg, I want a guest with no host home and no custom home), and comes handy for the sandboxing feature :+1:

> Assuming this gets reworked, how do you feel about the [lookup of the custom home in the container](https://github.com/gunnarx/distrobox/blob/dont_mount_host_home/distrobox-enter#L259) to use that as the workdir?

I was thinking the same as you, this would be a much cleaner and sure way to extrapolate the home 

One thing that worries me are performances, doing an inspect can take up to 200ms on a slower machine, we already to an inspect in line 345-350:

```sh
container_status=unknown
container_PATH="${PATH}"
eval "$(${container_manager} inspect --type container "${container_name}" --format \
	'container_status={{.State.Status}};
	{{range .Config.Env}}{{if slice . 0 5 | eq "PATH="}}container_PATH={{slice . 5 | printf "%q"}}{{end}}{{end}}')"
container_exists="$?"
```

At this point I'd add here in the eval of the inspect another variable to extrapolate the HOME, so that we can do all with only one inspect command, maintaining down the distrobox enter time

gunnarx:
> At this point I'd add here in the eval of the inspect another variable to extrapolate the HOME, so that we can do all with only one inspect command, maintaining down the distrobox enter time

That makes sense.  Well I think we've exhausted the discussion around the code I provided, so I would close and perhaps return to it with another code approach later.  Thanks.

89luca89:
Cool will wait for it :wink:

PR #236, opened 2022-04-15
mount-rshared

SoftwareRat:
Fixing "Error response from daemon: path /dev is mounted on /dev but it is not a shared or slave mount" with `mount --make-rshared /` command

### DIFF
--- a/install
+++ b/install
@@ -97,6 +97,9 @@ if [ "${verbose}" -ne 0 ]; then
 	set -o xtrace
 fi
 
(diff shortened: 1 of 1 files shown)


89luca89:
Hi @SoftwareRat 
Sadly I cannot accept this Pull Request

this issue is with how a distribution is set up, and it's not up to distrobox to do something this invasive (plus it would not work with sudoless installations)

This is an incompatibility that has to be resolved upstram (on the distro) or downstream (on the package maintainer), but not here that would affect all distributions

I'm closing this as I cannot accept this, sorry

PR #238, opened 2022-04-18
Update run_latest_gnome_kde_on_distrobox.md

bam80:
fix broken links

PR #239, opened 2022-04-18
Update README.md

bam80:
fix broken links


89luca89:
Cool thanks @bam80 
On #238 you removed the `KDE` name, leaving only plasma, here it remains, can we either have it or not in both parts for consistency?

bam80:
That was intentional - mentioned KDE Plasma once, we don't need to repeat it every time and just call it Plasma further.
That's my idea at least :)

89luca89:
Thanks, merging :+1:

PR #244, opened 2022-04-26
enter: fix unbound variable with --dry-run

vyskocilm:
The --dry-run switch for enter command fails with

```
DISTROBOX_ENTER_PATH=/usr/bin/distrobox-enter"'
++ set +o xtrace
/usr/bin/distrobox-enter: line 279: container_PATH: unbound variable
```

generate_command expects container_PATH to be declared, so move the
variable declaration slightly upper fixes the problem.

### DIFF
--- a/distrobox-enter
+++ b/distrobox-enter
@@ -333,6 +333,7 @@ generate_command() {
 	printf "%s" "${result_command}"
 }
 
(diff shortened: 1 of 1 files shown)


89luca89:
Thanks @vyskocilm for the contribution!

Issue #252, opened 2022-05-03
[Suggestion] Passing unsupported CLI flag should generate error


aral:
## Use case

Passing an unsupported flag currently fails silently.

This can complicate troubleshooting problems. (e.g., when two different versions of Distrobox are at play or if a flag is misspelled). For example, see https://github.com/89luca89/distrobox/issues/249#issuecomment-1116016993

## Proposed solution

Implement an error message, e.g., 

```
Error: Unknown flag --root 
```

89luca89:
Fixed by #255

PR #255, opened 2022-05-03
distrobox: Error out on invalid options

nathanchance:
As suggested in #252.

NOTE: I lightly tested this and it works but it would be nice if others could verify that it does not trigger any false positives.


89luca89:
I'll test this tomorrow, thanks!

89luca89:
Tested a bit and seems to be working as intended, thanks!

PR #259, opened 2022-05-05
Fixed typo in bash option

nogajun:
Fixed a bash option that would result in "bashl" instead of "bash -l".


89luca89:
Thanks a lot!

Issue #260, opened 2022-05-05
[Issue] Repository group permissions


misobarisic:
Ever since git added [safe directories](https://git-scm.com/docs/git-config/2.35.2#Documentation/git-config.txt-safedirectory), cloning the repo ("sudoless") and running `sudo ./install` does not work properly since the repo is owned by the user cloning it and not root which will be the one executing the files in this scenario.

This behaviour makes the current readme outdated.

nathanchance:
Would this fix it?

```diff
diff --git a/install b/install
index e5b9ac8..221ea63 100755
--- a/install
+++ b/install
@@ -104,10 +104,6 @@ cd "${curr_dir}" || exit 1
 # if files are available, install files in dest directory
 # else download targz and uncompress it
 if [ -e "${curr_dir}/distrobox-enter" ]; then
-       release_ver="github-zip-$(date +%F)"
-       if [ -d "${curr_dir}/.git" ] && command -v git > /dev/null; then
-               release_ver="git-$(git rev-parse HEAD)"
-       fi
        for file in distrobox*; do
                if ! install -D -m 0755 -t "${dest_path}" "${file}"; then
                        printf >&2 "Do you have permission to write to %s?\n" "${dest_path}"
```

It looks like this variable hasn't been used in this block since 68df9d2216465c4e29f43fae78d526c7f00ccd2.

misobarisic:
Yes, it does fix the issue.

nathanchance:
#261

89luca89:
Merged, thanks guys!

PR #261, opened 2022-05-05
install: Stop calling git

nathanchance:
As pointed out by Mišo Barišić in #260, this call might fail when
running `sudo ./install` because of the fix for CVE-2022-24765, which is
present in git 2.36.0 and certain maintenence releases from 2.30 to
2.35. This variable has been unused in this block since 68df9d2, where a
proper version value was used, rather than being dynamically updated by
`./install`.

Link: https://lore.kernel.org/xmqqv8veb5i6.fsf@gitster.g/


89luca89:
Indeed this was a leftoverfrom ho I previously managed versions
Safe to remove!
Thanks!

PR #262, opened 2022-05-06
Some openSUSE improvement

dfaggioli:
This pull request does the following:
1) fix openSUSE systemd (i.e., --init) enabled distrobox
2) customize openSUSE images during init, for enabling recommended packages
3) make it possible to keep recommended packages disabled --if one really wants that-- and do it in a way that is generic and can be used by other distroboxes and for other things

About 2 and 3, the first `enter` in a distrobox with is now quite a bit slower (because zypper is quite slow itself :-( ), but I think it's worth it. In fact, the resulting distrobox is now actually useful. E.g., GUI apps (but also non-GUI apps, as even just vim has problem withou recommended packages!) now works pretty much out of the box, once installed.

For comparison, I've created a Tumbleweed based distrobox, installed neofetch, vim and gedit, and it has ~600 packages. I've done the same on a rawhide one, and it has a little less than 500. So I think, all in all, it's fine, and well worth it.

### DIFF
--- a/distrobox-init
+++ b/distrobox-init
@@ -410,6 +410,17 @@ if ! command -v find || ! command -v mount || ! command -v passwd ||
 			vte-profile
 
 	elif command -v zypper; then
+		# In openSUSE official images, zypper is configured to ignore recommended
+		# packages (i.e., weak dependencies). This however, results in a rather
+		# poor out-of-the-box experience (e.g., when trying to run GUI apps).
+		# So, let's enable them. For the same reason, we make dure we install
+		# docs.
+		sed -i 's/.*solver.onlyRequires.*/solver.onlyRequires = false/g' /etc/zypp/zypp.conf
+		sed -i 's/.*rpm.install.excludedocs.*/rpm.install.excludedocs = no/g' /etc/zypp/zypp.conf
+		# With recommended packages, something might try to pull in
+		# parallel-printer-support which can't be installed in rootless podman.
+		# Since we very much likely never need it, just lock it
+		zypper al parallel-printer-support
 		# Check if shell_pkg is available in distro's repo. If not we
 		# fall back to bash, and we set the SHELL variable to bash so
 		# that it is set up correctly for the user.
--- a/distrobox-init
+++ b/distrobox-init
@@ -427,6 +427,7 @@ if ! command -v find || ! command -v mount || ! command -v passwd ||
 			procps \
 			shadow \
 			sudo \
+			systemd \
 			util-linux
 
 	else


89luca89:
Hi @dfaggioli  thanks for the PR!

I was thinking, would it be better to default to something more generic like 

`${DBX_CONTAINER_UNMINIMIZE}`

The upside would be easier documentation (one var for all distros) and we can use it to unminimize also Ubuntu Debian and Fedora (which I can add later after this PR ) :+1:

distrobox-create
```diff
@@ -505,6 +507,13 @@ generate_command() {
 	# Add additional flags
 	result_command="${result_command} ${container_manager_additional_flags}"
 
+	# Add init-time environment variables. The only requirement is that
+	# their name begins with DBX_CONTAINER_INIT_. They'll (likely!) be
+	# consumed by container-init, or ingeneral inside of the distrobox.
+	for i in $(printenv | grep '=' | grep -Ev ' |"' | grep '^DBX_CONTAINER_INIT_'); do
+		result_command="${result_command} --env \"$i\""
```
89luca89, line 514: We will need curly braces around the variable or it won't pass the linting

89luca89 (changes requested):

distrobox-create
dfaggioli: Yep, forgot to re-run shellcheck locally before pushing! Sorry :-)

dfaggioli:
Mmm... interesting, on openSUSE, something is now trying to install `parallel-printer-support` (really?!?!), which is trying to create stuff like /dev/lp0, and failing. This only happens with the toolbox image (which is why I didn't see, as I focused my testing on Tumbleweed) xxx.

The most of the dependencies come from vte, and the packages are installed before marking /dev as a net-share. But we can't move all the package handling after that, because we need some of those packages there very early.

I can try to split packages installation in two steps. The second one (which for know would only contain vte packages, but I guess other packages, in future, may need the same treatment) will happen later, after mounts have been marked.

dfaggioli:
> I can try to split packages installation in two steps. The second one (which for know would only contain vte packages, but I guess other packages, in future, may need the same treatment) will happen later, after mounts have been marked.

Nope, I've just tested this, and it still does not work.

The only solution I'm seeing right now, is to lock out the `parallel-printer-support` package. Not being able to install such package inside a distrobox does not seem a big deal at all to me... So I'm updating the pull request to that effect.

If anyone see a better solution, I'm all ears :-D

dfaggioli:
> Hi @dfaggioli thanks for the PR!
> 
> I was thinking, would it be better to default to something more generic like
> 
> `${DBX_CONTAINER_UNMINIMIZE}`
> 
Ok, I can do that.

One question: do I keep the DBX_CONTAINER_INIT_ prefix, for the variable name, and then only define in the environment the variables that have that prefix (as I'm doing right now)?

Or shall I just drop the _INIT_ part, and just put in the env all the DBX_CONTAINER_ variables? If we're making this generic, I'm thinking this second solution would be just fine, but I'm happy to do it the way you like it better. :-)

89luca89:
> > Hi @dfaggioli thanks for the PR!
> > I was thinking, would it be better to default to something more generic like
> > `${DBX_CONTAINER_UNMINIMIZE}`
> 
> Ok, I can do that.
> 
> One question: do I keep the `DBX_CONTAINER_INIT_` prefix, for the variable name, and then only define in the environment the variables that have that prefix (as I'm doing right now)?
> 
> Or shall I just drop the `_INIT_` part, and just put in the env all the DBX_CONTAINER_ variables? If we're making this generic, I'm thinking this second solution would be just fine, but I'm happy to do it the way you like it better. :-)

I think droppint _INIT_ is fine, after this I think I'll make it also a command line flag (`--unminimize` something like that) so that it can also be configured in the conf files and so on :+1:

89luca89:
> > I can try to split packages installation in two steps. The second one (which for know would only contain vte packages, but I guess other packages, in future, may need the same treatment) will happen later, after mounts have been marked.
> 
> Nope, I've just tested this, and it still does not work.
> 
> The only solution I'm seeing right now, is to lock out the `parallel-printer-support` package. Not being able to install such package inside a distrobox does not seem a big deal at all to me... So I'm updating the pull request to that effect.
> 
> If anyone see a better solution, I'm all ears :-D

Doubt that we need parallel printer in a container :thinking: so maybe blocking it is ok I guess

distrobox-init
```diff
@@ -196,6 +196,22 @@ mount_bind() (
...
+# which may sound good, but results if a much worse out-of-the-box experience
+# for users that, for instance, want to run GUI apps in their distroboxes.
+# In fact, we expect that most users would want recommended packages (which
+# are enabled by default in "regular" openSUSE/SUSE installations) in their
+# boxes, so let's re-enable them by default.
+#
+# If wanting to keep them disabled, the user can define to false the
+# DBX_INIT_ZYPPER_NO_RECOMMENDS env variable.
+[ "${DBX_CONTAINER_INIT_ZYPPER_NORECOMMENDS:-}" = "" ] && DBX_CONTAINER_INIT_ZYPPER_NORECOMMENDS=false
+if [ -f /etc/zypp/zypp.conf ]; then
```
89luca89, line 210: Was also thinking, instead of this check, we could move the `sed` inside the pkg manager part, so we can standardize when I'll add the other distros unminimization

89luca89 (changes requested):

distrobox-init
dfaggioli: Sounds fine... The only thing i'm not sure is when you say "instead of this check". I can move the check and the sed in the package management "section", but I still need the check, don't I?

89luca89: Yes in the "zypper" section let's call it 😄 
In was referring to the file existence check, I think if you have zypper you mist have zypper.conf right?
Just a thought, we can leave the check as it is no problem 

If was just referring to the file check, obv the variable check remains :+1:

dfaggioli: Ah, sure! Sorry, it was actually even quite obvious that you were... But for some reason I thought about the variable check. Sorry for the noise! Will do :-)

dfaggioli:
> I was thinking, would it be better to default to something more generic like
> 
> `${DBX_CONTAINER_UNMINIMIZE}`
> 
> The upside would be easier documentation (one var for all distros) and we can use it to unminimize also Ubuntu Debian and Fedora (which I can add later after this PR ) +1

Hey, me again, and sorry for bothering. I was back working on this, and I can't convince myself to like "unminimize", mostly because of the "inverted logic".

How about we call it something like DBX_CONTAINER_MINIMIZE or DBX_CONTAINER_MINIMAL_IMAGE ? Then, for openSUSE, I'd just set `onlyRequires` to `true`, **unless** I find that the variable is set to `true` (or `1`).

It seems to me that this could work well also for other distros. In fact, AFAICT, Fedora installs soft dependency by default already, so one could just handle it in a similar way...

That said, if doing things this way is, for some reason that I don't see, unideal, or if you don't like, I'll go for UNMINIMIZE... Just let me know. :-)

89luca89:
> > I was thinking, would it be better to default to something more generic like
> > `${DBX_CONTAINER_UNMINIMIZE}`
> > The upside would be easier documentation (one var for all distros) and we can use it to unminimize also Ubuntu Debian and Fedora (which I can add later after this PR ) +1
> 
> Hey, me again, and sorry for bothering. I was back working on this, and I can't convince myself to like "unminimize", mostly because of the "inverted logic".
> 
> How about we call it something like DBX_CONTAINER_MINIMIZE or DBX_CONTAINER_MINIMAL_IMAGE ? Then, for openSUSE, I'd just set `onlyRequires` to `true`, **unless** I find that the variable is set to `true` (or `1`).
> 
> It seems to me that this could work well also for other distros. In fact, AFAICT, Fedora installs soft dependency by default already, so one could just handle it in a similar way...
> 
> That said, if doing things this way is, for some reason that I don't see, unideal, or if you don't like, I'll go for UNMINIMIZE... Just let me know. :-)

I was thinking the same to be honest :smile: 
Generally speaking I was thinking that UNMINIMIZE should be default, distrobox use cases is to have a usable userland, so having minimized containers, does not make a lot of sense

So maybe, without bothering to check variables, just unminimize it?

dfaggioli:
> > How about we call it something like DBX_CONTAINER_MINIMIZE or DBX_CONTAINER_MINIMAL_IMAGE ? Then, for openSUSE, I'd just set `onlyRequires` to `true`, **unless** I find that the variable is set to `true` (or `1`).
> > It seems to me that this could work well also for other distros. In fact, AFAICT, Fedora installs soft dependency by default already, so one could just handle it in a similar way...
> > That said, if doing things this way is, for some reason that I don't see, unideal, or if you don't like, I'll go for UNMINIMIZE... Just let me know. :-)
> 
> I was thinking the same to be honest smile 

:-)

> Generally speaking I was thinking that UNMINIMIZE should be default, distrobox use cases is to have a usable userland, so having minimized containers, does not make a lot of sense

Totally agree!
 
> So maybe, without bothering to check variables, just unminimize it?

You mean we just get rid of it entirely?

89luca89:
I'd put the conf manipulation just before the `shell_pkg` installation, so we ensure that all the base dependencies are installed with recommends and docs :+1:

89luca89 (approved):

89luca89:
@dfaggioli 
I'd also check for:

`rpm.install.excludedocs=yes`

That also is really useful to unminimize :+1:

PR #273, opened 2022-05-16
Couple of small docs & comments fixes

dfaggioli:


89luca89:
Thanks a lot @dfaggioli !

PR #283, opened 2022-05-21
distrobox-host-exec: new tool for host command execution

dfaggioli:
Executing commands on the host, while inside of a container, can be done
either with flatpak-spawn, of playing tricks with chroot. Let's provide
a new script, which will be available inside of the container, that does
exactly that.

This tries to improve (although maybe not completely fix):
https://github.com/89luca89/distrobox/issues/177

Signed-off-by: Dario Faggioli <dfaggioli@suse.com>

### DIFF
--- a/distrobox-host-exec
+++ b/distrobox-host-exec
@@ -1,0 +1,142 @@
+#!/bin/sh
+# SPDX-License-Identifier: GPL-3.0-only
+#
+# This file is part of the distrobox project:
+#    https://github.com/89luca89/distrobox
+#
+# Copyright (C) 2022 distrobox contributors
+#
+# distrobox is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License version 3
+# as published by the Free Software Foundation.
+#
+# distrobox is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with distrobox; if not, see <http://www.gnu.org/licenses/>.
+
+trap '[ "$?" -ne 0 ] && printf "\nAn error occurred\n"' EXIT
+
+# Defaults
+distrobox_host_exec_default_command="/bin/sh"
+
+verbose=0
+version="1.2.16"
+
+# Print usage to stdout.
+# Arguments:
+#   None
+# Outputs:
+#   print usage with examples.
+show_help() {
+	cat << EOF
+distrobox version: ${version}
+
+Usage:
+
+	distrobox-host-exec [command [arguments]]
+	distrobox-host-exec ls
+	distrobox-host-exec bash -l
+	distrobox-host-exec flatpak run org.mozilla.firefox
+	distrobox-host-exec podman ps -a
+
+
+Options:
+
+	--help/-h:		show this message
+	--verbose/-v:		show more verbosity
+	--version/-V:		show version
+EOF
+}
+
+# Parse arguments
+while :; do
+	case $1 in
+		-h | --help)
+			# Call a "show_help" function to display a synopsis, then exit.
+			show_help
+			exit 0
+			;;
+		-v | --verbose)
+			verbose=1
+			shift
+			;;
+		-V | --version)
+			printf "distrobox: %s\n" "${version}"
+			exit 0
+			;;
+		--) # End of all options.
+			shift
+			;;
+		-*) # Invalid options.
+			printf >&2 "ERROR: Invalid flag '%s'\n\n" "$1"
+			show_help
+			exit 1
+			;;
+		*)
+			command=${distrobox_host_exec_default_command}
+			if [ -n "$1" ]; then
+				command=$1
+				shift
+			fi
+			break
+			;;
+	esac
+done
+
+set -o errexit
+set -o nounset
+# set verbosity
+if [ "${verbose}" -ne 0 ]; then
+	set -o xtrace
+fi
+
+# Check we're running inside a container and not on the host
+if [ ! -f /run/.containerenv ] && [ ! -f /.dockerenv ]; then
+	printf >&2 "You must run %s inside a container!\n" " $(basename "$0")"
+	exit 126
+fi
+
+mode=flatpak
+result_command="flatpak-spawn --host --forward-fd=1 --forward-fd=2 "
+
+if ! command -v flatpak-spawn > /dev/null; then
+	printf "WARNING: flatpak-spawn not found!\n"
+	printf "We recommend installing it and then trying distrobox-host-exec again.\n\n"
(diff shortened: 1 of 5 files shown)


dfaggioli:
I think we should also try to install flatpak-spawn automatcally.

I can do that in another patch. Shall we do that at init time, or the first time the user uses `distrobox-host-exec`?

I kind of like the latter, but it'd mean copying the package handling logic in that script, which is not really nice...

misobarisic:
Adding a new command requires updating the readme's "What it does" to reflect the addition.

dfaggioli:
Hey, @misobarisic , @89luca89 ! Pull request updated; I think I've done what you've asked

89luca89 (changes requested):

89luca89:
@dfaggioli thanks Dario! Left a couple of comments

Testing today the host-exec, I've encountered quite a few problems on the `chroot` approach sadly
I see that not only you cannot use `podman` or `docker`, but you cannot even use `sudo` 

Here are examples:

```console
luca-linux@distrobox:~/Desktop$ ./distrobox-host-exec podman ps
Warning: flatpak-spawn not found! We recommend installing it.
In the meantime, we'll try an alternative (but less powerfull) approach
WARN[0000] "/" is not a shared mount, this could cause issues or missing mounts with rootless containers 
cannot clone: Operation not permitted
WARN[0000] "/" is not a shared mount, this could cause issues or missing mounts with rootless containers 
cannot clone: Operation not permitted
WARN[0000] "/" is not a shared mount, this could cause issues or missing mounts with rootless containers 
cannot clone: Operation not permitted
ERRO[0000] invalid internal status, try resetting the pause process with "podman system migrate": cannot re-exec process 
```

And for sudo:

```console
luca-linux@distrobox:~/Desktop$ ./distrobox-host-exec bash -l
Warning: flatpak-spawn not found! We recommend installing it.
In the meantime, we'll try an alternative (but less powerfull) approach
[luca-linux@microos Desktop]$ sudo su -
sudo: /etc/sudo.conf is owned by uid 65534, should be 0
sudo: /bin/sudo must be owned by uid 0 and have the setuid bit set
[luca-linux@microos Desktop]$ 
```

This is only for rootless containers, rootful containers don't have the second problem.

In case of rootful containers we could use nsenter to escape the container:

`nsenter -t MY_PID -m -u -n -i sh`

Where MY_PID is whatever pid from the $USER

This is obviously not possible as a general approach, where `flatpak-spawn` excels!

So either:

- we found another way to escape the container
- we inform quite more the user of the chroot limitation

distrobox-host-exec
```diff
@@ -0,0 +1,115 @@
...
+	set -o xtrace
+fi
+
+# Check we're running inside a container and not on the host
+if [ ! -f /run/.containerenv ] && [ ! -f /.dockerenv ]; then
+	printf >&2 "You must run %s inside a container!\n" " $(basename "$0")"
+	exit 126
+fi
+
+if ! command -v flatpak-spawn; then
```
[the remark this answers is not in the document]
dfaggioli: Ok. I've seen other places in the code base where that is not done, so I wasn't sure. But, looking again, I think I understand why it's that, in those, and I agree that it makes sense to do it there.

distrobox-host-exec
```diff
@@ -0,0 +1,115 @@
...
+	printf "In the meantime, we'll try an alternative (but less powerfull) approach\n"
+
+	result_command="sudo -E chroot --userspec=$(id -u):$(id -g) /run/host/ /usr/bin/env "
+	for i in $(printenv | grep "=" | grep -Ev ' |"' | grep -Ev "^(_)"); do
+		result_command="${result_command} ${i}"
+	done
+
+	exec ${result_command} sh -c " cd ${PWD} && ${command} $*"
+else
+	flatpak-spawn --host "${command}" "$@"
```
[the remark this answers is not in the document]
dfaggioli: Good point. I'll rework the code toward that

dfaggioli:
> So either:
>     * we found another way to escape the container
>     * we inform quite more the user of the chroot limitation
> 
Ok... Let me put something together for option 2, and let's see what you think.

dfaggioli:
> > So either:
> > * we found another way to escape the container
> > * we inform quite more the user of the chroot limitation
> 
> Ok... Let me put something together for option 2, and let's see what you think.
> 
So, @89luca89 , how about now?

If you like it, there are those two shellcheck warnings, that I'm not sure I fully understand what they're trying to tell me, and how to make them go away... :-( Any pointer/suggestion would be much appreciated :-)

89luca89:
> > > So either:
> > > 
> > > * we found another way to escape the container
> > > * we inform quite more the user of the chroot limitation
> > 
> > 
> > Ok... Let me put something together for option 2, and let's see what you think.
> 
> So, @89luca89 , how about now?
> 
> If you like it, there are those two shellcheck warnings, that I'm not sure I fully understand what they're trying to tell me, and how to make them go away... :-( Any pointer/suggestion would be much appreciated :-)

Yep! Much cleaner that the chroot is really there just in case

For the shellcheck, I've found the problems, I'll comment on the code :+1:

distrobox-host-exec
```diff
@@ -0,0 +1,142 @@
...
+
+mode=flatpak
+result_command="flatpak-spawn --host --forward-fd=1 --forward-fd=2 "
+
+if ! command -v flatpak-spawn > /dev/null; then
+	printf "WARNING: flatpak-spawn not found!\n"
+	printf "We recommend installing it and then trying distrobox-host-exec again.\n\n"
+	printf "Alternatively, we can try an different (chroot-based) approach, but please\n"
+	printf "be aware that it has severe limitations and some commands will not work!\n\n"
+	printf "Do you really want to continue without installing flatpak-spawn? [N/y]\n"
```
89luca89, line 111: I'd replace the final '\n' with just a space here, so that the response to N/y is in line with the question

distrobox-host-exec
```diff
@@ -0,0 +1,142 @@
...
+			;;
+	esac
+fi
+
+# Let's also pass back the environment
+for i in $(printenv | grep "=" | grep -Ev ' |"' | grep -Ev "^(_)"); do
+	if [ "${mode}" = "chroot" ]; then
+		result_command="${result_command} ${i}"
+	else
+		result_command="${result_command} --env=\"${i}\" "
```
89luca89, line 138: problem is here, I've also tested a bit and seems like flatpak-spawn does not want the quotes around the --env argument, like instead podman wants

so just using

`result_command="${result_command} --env=${i} "`

And also the shellcheck stuff should disappear

89luca89 (changes requested):

distrobox-host-exec
dfaggioli: Done

distrobox-host-exec
dfaggioli: Ah, that's quite convenient then! :-D

Thanks for checking this

89luca89 (approved):

89luca89:
Thanks @dfaggioli LGTM, for now the chroot is explicitly shown as an emergency fallback thanks!

This should Fix #177 

Thanks!

PR #284, opened 2022-05-21
Fix gentoo support

nullpointerarray:
this patch includes multiple fixes:
- the local repository no longer gets automatically updated. Other
package managers dont do this either and the output can include
something about an error, that isnt actually a problem, which causes
distrobox-init to fail.
- there are multiple curl packages for gentoo. net-misc/curl is the web
request tool.
- vte-profile doesnt exist for gentoo. The closest package is probably
vte.
- some packages compilation will report errors, which get ignored by the
build tools, so dont matter, but the mention will cause distrobox-init
to abort. To prevent this --quiet-build is used to supress to messages
from compilation

### DIFF
--- a/distrobox-init
+++ b/distrobox-init
@@ -38,7 +38,7 @@ version="1.2.16"
 # Outputs:
 #   print usage with examples.
 show_help() {
-	cat <<EOF
+	cat << EOF
 distrobox version: ${version}
 
 Usage:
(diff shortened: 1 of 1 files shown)


nullpointerarray:
CONTRIBUNTING.md recommends using shfmt -s but the ci uses shfmt -s -ci -sr -kp. Should i update CONTRIBUTING.md?

89luca89:
LGTM, thanks for the contribution!

PR #285, opened 2022-05-21
docs: signal a problem on OpenSUSE Leap 15.3 and 15.4

massimo-zaniboni:
I added these lines in the documentation, for OpenSUSE Leap 15.3 and 15.4, used as host

``` md
podman v3.4.4`` cannot [initialize correctly](https://bugzilla.opensuse.org/show_bug.cgi?id=1199790) the containers managed by ``distrobox``. ``docker`` is [less secure](https://github.com/89luca89/distrobox/issues/223), but it works correctly.|
```

The problem is subtle because the container can work, but it is not completely initializated: some parts of the initialization are interrupted, and there can be problems later. With ``docker`` is instead all ok.


massimo-zaniboni:
> is this issue related to the problem we're facing in #286 ?

Yes. 

The scenario is this:
* I use OpenSUSE Leap as host system (email, browsing, etc..) because it is extremely stable;
* I use Distrobox + OpenSUSE Tumbleweed for developing, because it is cutting edge: nearly comparable to Arch Linux or NixOS;
* I like very much this combo: so many thanks for your wonderful work on Distrobox. I used in the past OpenSUSE Tumbleweed also as host system, but there can be transient problems during upgrades, because it is a rolling distro;
* I use Podman because it is safer and cheaper than Docker;
* there is a fail in the initialization of the container, but it is a non blocking problem, because I can enter in the container in the second attempt, and I don't notice nothing of wrong;
* then after some usage of the container and/or updates of the distro inside it, there can be wrong things, because the container is not in 100% correct state: initialization was interrupted. In my case there were problems with Latex upgrade, because there were problems with the user "nobody";
* I spent some time trying to find the error, and I started thinking that Tumbleweed is not so much robust or that Distrobox has stability problems. But this is wrong: the container was not in a good state from the beginning, and I ignored the errors;

All this will not happen, if I start with Docker from the beginning. So until the problem is not fixed, it is useful warning other users.

**Probably** a good companion solution can be also this: Distrobox should flag containers with bad initialization, and warn the user about this fact, after every ``distrobox-enter``. So, the user is informed that something can be wrong in the container, and that he is using it at his own risk.

89luca89:
Let's see if we manage to fix via the other issue

Anyway the warning of initialization was issued, the enter errored out :smile:

massimo-zaniboni:
Hi @89luca89, I updated the warning message. IMHO it is important insert it into the documentation, because it saves time to users of Distrobox.

89luca89:
Thanks @massimo-zaniboni seems good now, linking to the bug tracker is the correct thing 

Merging

Issue #288, opened 2022-05-24
[Suggestion] create should offer to pull the latest image by default


juhp:
**Is your feature request related to a problem? Please describe.**

It is quite common to have old outdated container images cached locally,
but I would certainly always prefer to pull the newest available image by default
when creating a new toolbox container. There could be an option to disable this of course.

eg recently I created a distrobox toolbox and it used an out of date c9s container image
from months ago, with outdated repo files. It is a bit tedious/error prone to pull images by hand
(eg one might accidentally pull from the wrong tag or name).

**Describe the solution you'd like**

`distrobox create` should offer to pull the latest image by default whenever creating a toolbox container.

(An option like `--no-pull` could be added if necessary to revert to the current behavior of only offering to pull if there is no local image.)

**Describe alternatives you've considered**

An alternative would be to add a `--pull` option but it is easy to forget to use it.

PR #289, opened 2022-05-24
create: prompt to download latest image by default

juhp:
This is a potential solution to #288, which gives
the user the chance to always create a new toolbox with
the latest most current image, instead of an old out-dated one.

### DIFF
--- a/distrobox-create
+++ b/distrobox-create
@@ -23,6 +23,7 @@
 #	HOME
 #	USER
 # Optional env variables:
+#	DBX_CONTAINER_ALWAYS_PULL
 #	DBX_CONTAINER_CUSTOM_HOME
 #	DBX_CONTAINER_IMAGE
 #	DBX_CONTAINER_MANAGER
@@ -40,6 +41,7 @@ if [ "$(id -u)" -eq 0 ]; then
 fi
 
 # Defaults
+container_always_pull=0
 container_clone=""
 container_image=""
 container_image_default="registry.fedoraproject.org/fedora-toolbox:35"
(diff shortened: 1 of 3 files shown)


89luca89:
Hi @juhp thanks!
Will test a bit the changes :+1:

89luca89:
Using it a bit, we might want to change a bit the behavior on this

Right now, every time I `create` it asks `you want to pull? y/N`
This does not make a lot of sense in the optics of having always the latest image
And more confusing is, if I say No, it creates the container if the image is present, which is a bit confusing

So either we could (as you proposed in #289 ) to have a `--pull` flag, which can be forgetful
or we could implement an ENV variable + config for `container_always_pull` for this

That would also set the non_interactive to 1, so that it does not ask questions

What do you think?

juhp:
Thanks for testing - I only really tried the default interactive case myself so far.

> Right now, every time I `create` it asks `you want to pull? [Y/n]`

Correct

> This does not make a lot of sense in the optics of having always the latest image 

(I didn't understand your point here.)

> And more confusing is, if I say No, it creates the container if the image is present, which is a bit confusing

So now "No" meant don't pull a new image (it is pretty similar to the old behavior with an existing image).

> So either we could (as you proposed in #289 ) to have a `--pull` flag, which can be forgetful or we could implement an ENV variable + config for `container_always_pull` for this

Okay, thanks, sounds good enough to me

89luca89:
> Thanks for testing - I only really tried the default interactive case myself so far.

Thanks to you for the PR :smile: 

> Okay, thanks, sounds good enough to me, let's try that then. Then one doesn't even have to answer the prompt which is even better really (once setup).

Yep should be less confusing, maintaining the current behavior and bypass it only in case of config and/or ENV variable

juhp:
How about this then?

89luca89:
> How about this then?

Great exactly what I was thinking :smile: 

There is one error for undefined variable in the code
Also we should update the usage in `docs/usage/distrobox-create.md` to reflect this change, and add also a couple of examples

Man pages are automatically generated from that change, so no need to recreate them yourself :+1:

juhp:
Ah yes thanks - hopefully better now and added more docs coverage.

89luca89 (approved):

89luca89:
Thanks a lot for the contribution!

juhp:
Thanks for your help too :-)

PR #302, opened 2022-06-03
Automatically create containers during 'distrobox-enter' (if they don't exist!)

dfaggioli:
When someone tries to enter a container that does not exist, let's offer the option of just creating it ourselves. The main (only?) issue is whether to accept and, if yes, how to handle 'distrobox-create' options in 'distrobox-enter'.

As a first step, let's ingore the problem entirely and just invoke 'distrobox-create $container_name', in this case.

An improvement could be making 'distrobox-enter' support all distrobox-create parameter. Of course, that would be "just" the parsing of them, to stash them somewhere and, if we get to creating the container, pass them to 'distrobox-create'. That's easy enough to do, but it then means that every time we add/remove/modify a parameter of 'distrobox-create', we also need to do the same change in 'distrobox-enter' parsing code, which is not really nice...

### DIFF
--- a/distrobox-enter
+++ b/distrobox-enter
@@ -364,24 +364,25 @@ if [ "${dryrun}" -ne 0 ]; then
 	exit 0
 fi
 
-# Inspect the container we're working with.
-container_status="unknown"
-eval "$(${container_manager} inspect --type container "${container_name}" --format \
-	'container_status={{.State.Status}};
-	{{range .Config.Env}}{{if slice . 0 5 | eq "HOME="}}container_home={{slice . 5 | printf "%q"}};{{end}}{{end}}
-	{{range .Config.Env}}{{if slice . 0 6 | eq "SHELL="}}container_shell={{slice . 6 | printf "%q"}};{{end}}{{end}}
-	{{range .Config.Env}}{{if slice . 0 5 | eq "PATH="}}container_path={{slice . 5 | printf "%q"}}{{end}}{{end}}')"
-container_exists="$?"
-# Set SHELL as a login shell
-container_shell="$(basename "${container_shell}") -l"
-# Does the container exists? check if inspect reported errors
-if [ "${container_exists}" -gt 0 ]; then
+# Check if the container is even there
+container_exists=$(${container_manager} ps -aq -f name=^"${container_name}"$)
+if [ "${container_exists}" = "" ]; then
 	# If not, prompt to create it first
 	printf >&2 "Cannot find container %s, does it exist?\n" "${container_name}"
 	printf >&2 "\nTry running first:\n"
 	printf >&2 "\tdistrobox-create <name-of-container> --image <remote>/<docker>:<tag>\n"
 	exit 1
 fi
+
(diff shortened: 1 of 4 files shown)


89luca89:
I really like the idea! :smile: 

Tested and it's working as intended, but at this point I have a doubt/question to solve

Contrary to toolbox implementations (like Microos, and Fedora), the point of distrobox is that you choose your distro
So right now, the create command defaults to using a simple `distrobox create $NAME` which will default to the
default container image (configured or not) and without any other option

Maybe we could add to the question, that we're going to use the default image (showing it if possible) so that the user
would know which image it is going to be used?

dfaggioli:
> Contrary to toolbox implementations (like Microos, and Fedora), the point of distrobox is that you choose your distro So right now, the create command defaults to using a simple `distrobox create $NAME` which will default to the default container image (configured or not) and without any other option
> 
> Maybe we could add to the question, that we're going to use the default image (showing it if possible) so that the user would know which image it is going to be used?

Yes, I can do that. I think I can also try to make it work somehow like this:

```
dario@Wayrath:~/src/toolbox/distrobox.git> distrobox-enter test-test-test
Cannot find container test-test-test
Create it now, out of  registry.opensuse.org/opensuse/tumbleweed:latest ?
[Yes/no/image]:
```

And, if the user answers `i`, `I` or `image`, we let him/her type the image to be used.

Also, if the image is the main concern, I think (but that's only my opinion) it's fine to have `distrobox-enter` take and handle `--image/-i`, just for the purpose of passing it to `distrobox-create`, in case the container does not exist. I'm not sure I'd do the same for *all* `distrobox-create` parameters, but if it's just one or two, it's probably ok...

I can implement the latter in a separate patch, so we'll see how it looks like, and we can keep it out if not liking it. :-)

89luca89:
Generally speaking I think just the question  

```
Create it now, out of  registry.opensuse.org/opensuse/tumbleweed:latest ?
[Y/n]:
```

Is more than enough for a couple of reasons
1 - we keep codebase simple 
2 - if an user does not want to use defaults, well they can use `create` first :-) 

But I like the question with the image name because it alerts the user on what's happening

distrobox-enter
```diff
@@ -364,24 +377,55 @@ if [ "${dryrun}" -ne 0 ]; then
 	exit 0
 fi
 
-# Inspect the container we're working with.
+# Check if the container is even there
+container_exists=$(${container_manager} ps -aq -f name=^"${container_name}"$)
```
89luca89, line 381: Could we use the previous `$container_exists` ?
This would let us do all the checks with just 1 `podman/docker inspect` instead of 2, to optimize
`enter` performance

89luca89 (changes requested):

distrobox-enter
89luca89: If we move all this block after the `eval ...` where `container_exists` was declared before, it should need no other changes

89luca89:
I'm merging this :+1:  I can take care of that small optimization, the rest is brilliant thanks!

PR #304, opened 2022-06-03
host exec docs: 755 -> 644 

misobarisic:
I changed the environment variable prefix in `export`. `DISTROBOX` -> `DBX` to match other project files.

Also removed `x` permissions wise from host-exec docs.


89luca89:
Thanks @misobarisic 

Not sure how I feel about those changes of variable names
For two main reasons:

1 - those are not meant to be configurable, those are stuff inected in the env by `create`, so keeping them separate from the rest could be useful
2 - this is a breaking change, to take effect people will need to re-create their containers

misobarisic:
Alright then, sounds good. Thanks for the heads up.

misobarisic:
Repurposed the PR to just be a permission change for the host exec docs

89luca89:
Cool thanks!

PR #330, opened 2022-06-20
docs: updated script in a post

misobarisic:
Removed redundant lines


89luca89:
Are they redundant?
Seems like the dbus one is needed, and the `/etc/systemd/system` one is good to ensure there are no leftovers in general

misobarisic:
>  `/run/systemd/system` one is good to ensure there are no leftovers in general

Alright, makes sense

89luca89:
No problems at all :+1:

PR #333, opened 2022-06-21
enter: if flagless container doesn't exist, try it as an image

juhp:
This is done with a new `container` variable which can be used as
fallback `container_name` if it exists, otherwise can attempt using it
as image of a new toolbox.

Without a flagless arg, enter tries `container_name_default` (still my-distrobox)
and then falls back to `container_image_default` (fedora-toolbox:36).

How does this look? It is possible I am missing some edge or corner cases,
but my from basic testing this seems to work okay to me.

### DIFF
--- a/distrobox-enter
+++ b/distrobox-enter
@@ -48,7 +48,9 @@ container_shell="$(basename "${container_shell}") -l"
 container_image=""
 container_image_default="registry.fedoraproject.org/fedora-toolbox:36"
 container_manager="autodetect"
-container_name="my-distrobox"
+container=""
+container_name=""
+container_name_default="my-distrobox"
 container_manager_additional_flags=""
 non_interactive=0
 
@@ -181,10 +183,10 @@ while :; do
 			exit 1
 			;;
 		*) # Default case: If no more options then break out of the loop.
-			# If we have a flagless option and container_name is not specified
-			# then let's accept argument as container_name
+			# If we have a flagless option
+			# then let's accept argument as container
 			if [ -n "$1" ]; then
-				container_name="$1"
+				container="$1"
 				shift
 			else
 				break
@@ -192,6 +194,14 @@ while :; do
 			;;
 	esac
 done
+
+if [ -z "${container_name}" ]; then
+	if [ -n "${container}" ]; then
+		container_name="${container}"
+	else
+		container_name="${container_name_default}"
+	fi
+fi
 
 set -o errexit
 set -o nounset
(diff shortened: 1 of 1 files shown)


89luca89:
And as conseguence, `enter` has to have the same defaults as `create` to have it work flagless

If you want to use `fedora-toolbox-36` with create, enter and others without having to specify flags, you can simply use the config file to change the defaults:

~/.distroboxrc
```sh
container_name="fedora-toolbox-36"
container_image="fedora-toolbox:36"
```

89luca89:
@juhp 
ah I see now! you want to do something like

`distrobox enter alpine:latest`

and make `alpine:latest` be the image for the prompt in the creation right?

juhp:
Right that is the idea!

Not sure I could have explained it any worse, lol ;o)
I was trying to describe the new logic - but my explanation was too terse and not so coherent.

juhp:
Also to add a little more background: I would like to be able to run ephemeral distroboxes:
ie a throwaway toolbox which is deleted when one quits it: for this a behavior like the one
here would be desirable - so that is why I started this PR.

89luca89:
> Also to add a little more background: I would like to be able to run ephemeral distroboxes:

Mhh you mean like `podman run --rm`? 
Would you like to open a discussion on this? Maybe we can get feedback from other people also, I'm not super-convinced about this as this seems pretty much the job for a traditional podman/docker? let's discuss this

> Right that is the idea!

Cool, will try and have a look, I'm not too keen to keep it flagless, it's confusing that you can specify 2 different things without flags (mutually exclusive)

My main train of thought is:

  - you want something simple? -> `distrobox enter` (flagless) will prompt for all defaults
  - you want something simple, but custom? -> use config file
  - you want something more? -> use `distrobox create` first

I don't want for `enter` to inherit too much of `create` here, as those are pretty distinct phases
Right now the question (does not exist, create?) makes sense if we want to just use the defaults, but it gets confusing quickly if we start to add `IMAGE` stuff to enter && create

Question should be, why do I have (or want to) pass an IMAGE to an enter command?

89luca89:
@juhp 

Regarding:

> Also to add a little more background: I would like to be able to run ephemeral distroboxes

I was thinking, wouldn't be easier to add a command to the `distrobox` command?
In the switch case:

```sh
case "${distrobox_command}" in
...
...
	run-and-rm)
		enter=$("${distrobox_path}"/distrobox-create "$@" 2>/dev/null)
		eval ${enter}
		eval $(echo "${enter}" | sed 's|enter|rm -f|g')
		;;
...
...
esac
```

And running it seems to be what you're looking for:

image

juhp:
You are right, a separate command could indeed work just fine for that, thanks!

Dunno if there's a better name - maybe `temporary` could be another option?

89luca89:
> You are right, a separate command could indeed work just fine for that, thanks!
> 
> Dunno if there's a better name - maybe `temporary` could be another option?

Like `run-temporary`? Could work yes
Also `run-spot` could work maybe

Wanna take this? 
If not i can work on it separately, just open an issue about it and we colse this PR no worries 👍

juhp:
I could but I may not have time soon (that's also why I was being lazy thought this could be a quick hack piggybacked onto enter, but you are right a separate command will be cleaner and makes more sense), so let's start with a ticket.

Issue #335, opened 2022-06-23
[rfe] command to run a temporary (throwaway) toolbox


juhp:
This is a follow-on from #333 where we had some initial discussion.

**Is your feature request related to a problem?**
Containers can be great for quickly testing things in a pristine environment.
(In fact I have a small [tool](https://github.com/juhp/fhcontainer) just for that.)
Even better sometimes one wants to test something in a short-lived toolbox.
This issue to suggest to add a new command to distrobox to allow running
a toolbox generated from some image and removing it as soon as it finishes
(ie an extension of `podman run --rm [-it] ...` with added toolbox goodness).

**Describe the solution you'd like**
I guess we can bike-shed the command name ;-P - but with shell completions
the exact name doesn't worry me so much: some early ideas include `run-temp`, `run-and-rm`, `temporary`, `ephemeral`.
I am more keen to have the functionality.

PR #336, opened 2022-06-23
'temporary' ephemeral distrobox (#335)

juhp:
Initial poc implementation of temporary toolbox command.

Well installing the basic packages takes a little time but
this is already quite useful to me.
Though it could surely be improved upon...
I needed this just now so I thought i would start this off.

In the longer term the right thing may well be to move
this out to a separate file like the other cmds.


misobarisic:
`distrobox` is just a wrapper script. "distrobox-temp" should be a separate script

89luca89:
If we move it to it's own file, I think we can at this point do some flag management
Specifically we can pass all arguments except `--exec and --` to create, and the command to execute to `enter` so oe could do temporary containers that execute specific commands

89luca89:
> If we move it to it's own file, I think we can at this point do some flag management Specifically we can pass all arguments except `--exec and --` to create, and the command to execute to `enter` so oe could do temporary containers that execute specific commands

I'd say we could start from the smallest script we have, `stop` or `rm`, and just add the `--/--exec` flag, and use the same code @juhp put in the command above, and we should be ok to start testing

89luca89:
@juhp created a dedicated file for it, please do some tests, seems like it's working as intended

accepts all flags from `create` plus a `--/--exec` to pass to enter

if all is ok, we tidy up the help, create the usage file before merging

juhp:
Okay, `distrobox-ephemeral -i ...` seems to work :+1: Thanks!

(I also tried `distrobox-ephemeral alpine` - that didn't go so well, but I can live without it.)

89luca89:
> (I also tried `distrobox-ephemeral alpine` - that didn't go so well, but I can live without it.)

well makes sense, the name is always random in ephemeral, and I like the idea of having them this way to be honest

I'll create the usage pages for it, do some testing and merge :+1:

juhp:
Thanks!

> > (I also tried `distrobox-ephemeral alpine` - that didn't go so well, but I can live without it.)

Well `ephemeral` always needs an image, so in my opinion being able to run just `distrobox-ephemeral alpine:latest` would be useful.

89luca89:
> Thanks!
> 
> > > (I also tried `distrobox-ephemeral alpine` - that didn't go so well, but I can live without it.)
> 
> Well `ephemeral` always needs an image, so in my opinion being able to run just `distrobox-ephemeral alpine:latest` would be useful.

But that would be in contrast with the rest of the ands that take the name as flagless argument

PR #342, opened 2022-06-28
feat: Replace call to 'rev | cut' with 'grep'

adam-gaia:
This modifies the parsing of $release_name from $release_ver by
grepping for a semantic version with a '.tar.gz' extension.

The prior method used 'rev | cut | rev' to isolate $release_name
as the last field of $release_ver. The rev tool is not installed
by default on some machines.


89luca89:
That's a nice catch thanks! :smile: 

I don't plan to use `-suffixes`, If I do we'll fix this no problem :+1:

PR #346, opened 2022-07-03
Implement options for sudo program customization

### LINKED ISSUE #312: [Suggestion] Prefer GUI authentication over sudo for rootful containers
I prefer my toolbox to use rootful podman so it has full access to the machine and /dev. The podman container is set up to start at boot, and I would like to open a terminal directly into the toolbox.

The problem is that `distrobox enter --root toolbox` uses `sudo` for authentication, so it has to be run from a shell or there's no way for me to enter my admin password.

My suggestion is to prefer the default polkit authentication UI whenever possible (see screenshot), instead of sudo. This can be done by replacing the `sudo` calls in distrobox with `pkexec`.

image

PgBiel:
Hello,
This PR aims to fix #312 by allowing the user to specify a different sudo command to enter rootful containers (as well as when exporting apps/services/binaries with sudo), as described below in detail. **(There are some caveats, as described more to the end.)**

# Changes
I separated the changes in three groups of (sub)commands:

## `distrobox-create`, `-enter`, `-list`, `-rm`, `-stop`

For these distrobox commands, if either `DBX_SUDO_PROGRAM` env var or `distrobox_sudo_program` config var is specified, then the specified sudo program will be used in place of `sudo` when calling podman/docker with root privileges to work with rootful containers.

~~Also, I have added a fix for `distrobox-enter` auto-creation of containers (when trying to enter a container that does not exist) - it used to always create non-rootful containers (thus ignoring `--root`). I made it now consider `--root` - if specified, it will create a rootful container instead (which should be the expected behavior).~~ (**EDIT:** The fix was reversed by request, as it will be addressed separately later.)

## `distrobox-ephemeral`

The distrobox commands run by `distrobox-ephemeral` will all already consider `DBX_SUDO_PROGRAM` and `distrobox_sudo_program` by themselves, thus no change in `distrobox-ephemeral`'s code was needed.

## `distrobox-export`

Apps/services/binaries exported from rootful containers invoke `distrobox-enter --root` in the host, which will thus use `DBX_SUDO_PROGRAM` env or `distrobox_sudo_program` config var as the sudo program (if either is present). In this sense, no change in `distrobox-export`'s code was needed as well.

# Docs & Tests

Documentation was updated for every affected command. Additionally, I added a section to `useful-tips.md` showing the most common usages of sudo program customization. I also wrote a section in `distrobox-export.md` for insight on how to run graphical apps exported from rootful containers when using the default `sudo`.

For every commit, I have run tests in an EndeavourOS VM, and everything seems to be working fine.

# Caveats

The given sudo program will be **called multiple times** throughout most commands. This isn't an issue with `sudo`, as it implements persistence by default (i.e., not having to write the root password multiple times during a certain interval), but could be an issue with `doas` and `pkexec`, since they aren´t configured to persist by default.

The solution is to either configure the given sudo program appropriately, or, if a graphical root authentication prompt is needed along with persistence, use `sudo` with `SUDO_ASKPASS`, as I explain in `useful-tips.md`.

I'll remain available to answer any questions that might remain regarding my contribution. If anything is missing or should be improved, please let me know.


89luca89:
Hi @PgBiel 
thanks for this contribution!

I'll go step by step and comment some parts of the implementation

1 - I feel like having a custom sudo program belongs more to a config (or ENV) than a flag, so I'd remove the `--sudo-program` flag. If one uses `doas` or `pkexec`, they can just export the ENV in their rc file, or use a distrobox.conf file.
The idea is that flags are for thing that I might want to have per-container

2 - the `distrobox_invoke_container_manager` function, you say "use functions to avoid injection", can you explain this a little bit?

PgBiel:
> Hi @PgBiel thanks for this contribution!
Hi, no problem! Please forgive me for the huge amount of text lol, I wanted to give every detail of what I was doing, but I recognize the final result might have gotten a bit... excessive lol

> 1 - I feel like having a custom sudo program belongs more to a config (or ENV) than a flag, so I'd remove the `--sudo-program` flag. If one uses `doas` or `pkexec`, they can just export the ENV in their rc file, or use a distrobox.conf file. The idea is that flags are for thing that I might want to have per-container

Ah, alright! This actually makes sense. I'll work on this later today, or perhaps tomorrow.

By the way, should this ENV stuff be the case for the `distrobox_export` flags as well? I.e. for `--sudo-program` (which defines the container's sudo) and `--host-sudo-program` (the host's sudo), should those be ENV (/config)-only variables inside the container? 

> 2 - the `distrobox_invoke_container_manager` function, you say "use functions to avoid injection", can you explain this a little bit?

Okay so, at first, I was basically thinking how like, if we simply replaced `sudo` in the `container_manager` string by whatever `sudo_program` was, this could enable globbing and splitting abuse, given that, when you expand a variable in a shell command call, it performs globbing and splitting on the content, so, if the sudo program string is maliciously changed, this could enable stuff such as passing arbitrary parameters to the command call, or globs that may crash the program

The use of a function protects against this, by passing the sudo program between double quotes instead of inserting it directly into the container_manager string, i.e.,

"${distrobox_sudo_program}" ${container_manager}

That way, if distrobox_sudo_program is something like "/* /* /* /*" or "bad_command --bad-parameter" or something like that, those will be inserted literally, as a command name, without expanding globs or parameters

However, recently I was re-thinking this, and, maybe if the sudo_program string can be maliciously changed, then it could probably also just point to a valid executable which happens to be malicious (against which the function doesn't protect). So maybe this might not be an actual concern, and the function is rather useless in that sense. 

And, also, given that we will remove the `--sudo-program` flag, checking if the program exists is also unneeded, since we wouldn't be aiming to be user-friendly anyways (and, if the user wants to use a custom program, it's their responsibility)

With that in mind, I believe I will be submitting a commit to reverse the function implementation to the older, simpler one, as I believe that it might not give the protection I was thinking it would. Sorry for that!

PgBiel:
Wow, what a nice reply! Thank you very much @89luca89 for understanding! The best part about being in the FOSS world is finding nice maintainers hahaha
I'll be releasing those commits pretty soon :+1:

89luca89:
Ah also @PgBiel 

I've noticed you're including this choice into `distrobox-export`
You should just check for host's sudo program (with the usual methods we discussed) as for **inside** the container, sudo will always be presend and configured, so it does not make sense to have the option to use another sudo program inside the container :+1:

PgBiel:
> as for **inside** the container, sudo will always be presend and configured, so it does not make sense to have the option to use another sudo program inside the container :+1:

Oh thank you, I was actually going to ask you this, because I was failing to make `--sudo-program` work with local ENV variables in the `distrobox-export` code lol
I had thought of making exported apps, services and binaries always attempt to use an ENV variable from inside the container instead of sudo (but defaulting to sudo if it is not present), but it seems that adding a global ENV variable inside the container _after_ it was created isn't trivial (I at first had thought of modifying `/etc/environment`, or adding a profile to `/etc/profile.d`, but it seemed that using `env=` podman options were the only way to make them truly global after all)

Anyway, but none of that matters, since `sudo` is guaranteed to be present (and with trivial usage), so this feature would be rather useless..
I will commit the new changes to distrobox-export in a bit :+1:

PgBiel:
> You should just check for host's sudo program (with the usual methods we discussed)
@89luca89 by the way, at first I was thinking of, by default, just keeping `distrobox-export` as it is currently (i.e. not changing the file at all in this PR), as the exported app/service/binary would likely already pick up the host's `DBX_SUDO_PROGRAM` env var or `distrobox_sudo_program` config

However, I was thinking that, maybe that wouldn't be desired for exporting graphical apps, because some users might prefer using `pkexec` (for example) for graphical apps and `sudo` for everything else

This is because, at least on my tests, `distrobox-export --app` on rootful containers will export a `.desktop` file that cannot be (easily) run by default, given that it will try to use sudo without a terminal
(Of course, this is configurable with `SUDO_ASKPASS`, but not everyone will want to - or be able to - use that) 

So perhaps I should keep host sudo program configurable on a per-item by detecting an env variable inside the container (let's say `DISTROBOX_EXPORT_HOST_SUDO_PROGRAM`), and, if it is present, it overrides the host's `DBX_SUDO_PROGRAM` env var on the app / binary  line

Usable like this, for example (given we're in a rootful container):

```shell
DISTROBOX_EXPORT_HOST_SUDO_PROGRAM="pkexec" distrobox-export --app simplescreenrecorder
```

And then, on the exported `.desktop`, the distrobox call would look something like this:

```shell
DBX_SUDO_PROGRAM="pkexec" distrobox-enter -n container -- simplescreenrecorder
```

Thus being possible to override the host sudo program depending on the item type

(Perhaps keeping `--host-sudo-program` would also be convenient here? As this would be per-item, basically)

What do you think?

PgBiel:
(Also, by the way: I have greatly simplified the PR description (!!), with the exception of the `distrobox-export` part, as I haven't updated its implementation with the latest changes we've discussed yet)

distrobox-enter
```diff
@@ -408,14 +412,25 @@ if [ "${container_status}" = "unknown" ]; then
...
-				"$(dirname "${0}")"/distrobox-create -i "${container_image}" -n "${container_name}"
+			if [ "${rootful}" -ne 0 ]; then # pass rootful parameter to the create subcommand
+				printf >&2 "  distrobox create -r -i %s %s\n" "${container_image}" "${container_name}"
+				if [ "${dryrun}" -ne 1 ]; then
+					"$(dirname "${0}")"/distrobox-create -r -i "${container_image}" -n "${container_name}"
+				fi
+			else # rootless container => no -r required
+				printf >&2 "  distrobox create -i %s %s\n" "${container_image}" "${container_name}"
+				if [ "${dryrun}" -ne 1 ]; then
+					"$(dirname "${0}")"/distrobox-create -i "${container_image}" -n "${container_name}"
```
89luca89, line 423: Can you revert this? I want to address this separately :+1: thanks!

distrobox-enter
```diff
@@ -408,14 +412,25 @@ if [ "${container_status}" = "unknown" ]; then
...
+				fi
 			fi
 			;;
 		n | N | No | no | NO)
 			printf >&2 "Ok. For creating it, run this command:\n"
-			printf >&2 "\tdistrobox create <name-of-container> --image <remote>/<docker>:<tag>\n"
+			if [ "${rootful}" -ne 0 ]; then # display -r if it was also given to distrobox enter
```
89luca89, line 429: Can you revert this? I want to address this separately :+1: thanks!

docs/usage/distrobox-export.md
```diff
@@ -104,6 +104,29 @@ The option "--delete" will un-export an app, binary, or service.
 
 The option "--sudo" will launch the exported item as root inside the distrobox.
 
+**Exporting items from rootful containers**
```
89luca89, line 107: This should be removed as the work on export has been reverted

docs/useful_tips.md
```diff
@@ -3,6 +3,8 @@
   - [Create a distrobox with a custom HOME directory](#create-a-distrobox-with-a-custom-home-directory)
   - [Mount additional volumes in a distrobox](#mount-additional-volumes-in-a-distrobox)
   - [Use a different shell than the host](#use-a-different-shell-than-the-host)
+  - [Run the container with real root](#run-the-container-with-real-root)
```
89luca89, line 6: What does this refer to?
`   - [Run the container with real root](#run-the-container-with-real-root)`

89luca89 (changes requested):

docs/usage/distrobox-export.md
PgBiel: Whoops, sorry.

docs/useful_tips.md
PgBiel: It refers to this section (which was apparently missing in the list at the top):

https://github.com/89luca89/distrobox/blob/1e4d7b1cc914d05cb00ae8d7df5e5828c9cf43f6/docs/useful_tips.md?plain=1#L70-L83

PgBiel:
Alright, well, the base feature should be working properly. Regarding the discussion of having a separate, graphical sudo program when exporting apps from rootful containers, perhaps that would be better left for a future issue/PR, as to not block this PR's merging any further (and also because it can already be solved using `SUDO_ASKPASS`, for example). So I believe things should be OK for merging now. Make sure to tell me if anything else should be changed. :+1:

docs/usage/distrobox-export.md
PgBiel: Update: I added a new section to replace this one with more useful instructions regarding the new implementation. Tell me if I should change anything :+1:

docs/useful_tips.md
89luca89: I would also add a clear warning that `root` inside a container that is running as `--root` or with `docker`, it's also root on the machine! It must be clear as it could be a dangerous thing

89luca89 (approved):

89luca89:
That's wonderful thanks! I've reviewed it and seems all good! Running the pipeline now, when it's all done, I'll merge :+1:

PgBiel:
Woo-hoo!!! Thank you very much for the merge! Glad I could contribute :)
 Let's hope for more in the future!! :smile: :rocket:

PR #349, opened 2022-07-11
host-exec: replace chroot with host-spawn

### LINKED ISSUE #321: [Suggestion] Provide a better alternative to flatpak-spawn --host

`flatpak-spawn --host` is a nice convenience to be able to run programs on the host machine from inside the toolbox. But it has some major shortcoming, such as not correctly passing through the tty devices, so not all programs can be run cleanly using this tool.

An example of these shortcomings is if you try to open a shell on the host, and run sudo.

```
toolbox:~ % flatpak-spawn --host bash
bash: cannot set terminal process group (-1): Inappropriate ioctl for device
bash: no job control in this shell
[sph@host ~]$ sudo -i
sudo: a terminal is required to read the password; either use the -S option to read from standard input or configure an askpass helper
^Csudo: a password is required
```

AFAIK it is possible to write an alternative to `flatpak-spawn --host` that directly interacts with Flatpak DBus portal to avoid all this shortcomings. Though this alternative does not exist yet, I would like Distrobox to ship with such an alternative to provide a seamless experience. 

Here's the Tilix terminal emulator spawning a command on the host through DBus: https://github.com/gnunn1/tilix/blob/5008e73a278a97871ca3628ee782fbad445917e7/source/gx/tilix/terminal/terminal.d#L2963

I know this is quite complicated, but I think distrobox would benefit if people can access their host machine without any compromise. 

**Additional context**

A collection of issues about flatpak-spawn --host: 

* https://github.com/flatpak/flatpak-xdg-utils/issues/57
* https://github.com/flatpak/flatpak/issues/3697
* https://github.com/flatpak/flatpak/issues/3285

89luca89:
Thanks to @1player, we now have a single-binary solution to replace
flatpak-spawn when not present.

Right now the solution is to ask to download the matching-version binary
from github releases.

This PR also implements the use of symlinks to make `distrobox-host-exec` execute as that command:

```console
~$: ln -s /usr/bin/distrobox-host-exec /usr/local/bin/podman
~$: ls -l /usr/local/bin/podman
lrwxrwxrwx. 1 root root 51 Jul 11 19:26 /usr/local/bin/podman -> /usr/bin/distrobox-host-exec
~$: podman version
...this is executed on host...
```

Fix #321

Signed-off-by: Luca Di Maio <luca.dimaio1@gmail.com>

### DIFF
--- a/distrobox-host-exec
+++ b/distrobox-host-exec
@@ -21,8 +21,9 @@
 trap '[ "$?" -ne 0 ] && printf "\nAn error occurred\n"' EXIT
 
 # Defaults
-distrobox_host_exec_default_command="/bin/sh"
-
+host_command=""
+distrobox_host_exec_default_command="${SHELL:-/bin/sh}"
+host_spawn_version="1.0.1"
 verbose=0
 version="1.3.2"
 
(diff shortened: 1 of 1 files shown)


1player:
Two questions:

1. Can I choose to prefer `host-spawn` instead of `flatpak-spawn --host`? I'd like to keep both installed, but I prefer the former, of course.
2. This requires `host-spawn` to be installed in /usr/bin. Anywhere in $PATH should be fine and preferable (I keep it in `~/.local/bin`) than only using it if it's in that exact location.

Issue #352, opened 2022-07-16
[Suggestion] distrobox-list should suppress colour when attached to a pipe


juhp:
**Is your feature request related to a problem? Please describe.**
I love the colored output of distrobox-list.
But if I run `distrobox-list | ...` then the colour control codes are still passed out.

**Describe the solution you'd like**
How about suppressing coloured output when stdout of distrobox-list is attached to a pipe?

**Additional context**
It seems this can be detected like this, for example:
```shell
if [ -t 1 ] ; then echo terminal; else echo "not a terminal"; fi
```

89luca89:
@juhp great idea! Implemented :+1:

juhp:
Wow thank you @89luca89, so fast!

PR #353, opened 2022-07-16
Add no-color flag to list

misobarisic:
I believe explicitly specifying a `no-color` mode should be available as well.

Extends #352

### DIFF
--- a/distrobox-list
+++ b/distrobox-list
@@ -63,7 +63,7 @@ done
 # Outputs:
 #   print usage with examples.
 show_help() {
-	cat <<EOF
+	cat << EOF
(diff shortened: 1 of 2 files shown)


89luca89:
Thanks @misobarisic 
I think it's not passing lint because of formatting, refer to contributing docs to use shfmt with the right parameters 👍

misobarisic:
> I think it's not passing lint because of formatting, refer to contributing docs to use shfmt with the right parameters

Yeah, I was missing a flag. That's why it wasn't passing

89luca89:
Thabks a lot @misobarisic

PR #355, opened 2022-07-17
Adds Crystal Linux to the distro list and adds a disclaimer about using it on Distrobox

orowith2os:


docs/compatibility.md
```diff
@@ -144,6 +145,8 @@ NixOS is not a supported container distro, and there are currently no plans to
 bring support to it. If you are looking for unprivlaged NixOS environments,
 we suggest you look into [nix-shell](https://nixos.org/manual/nix/unstable/command-ref/nix-shell.html).
 
+Crystal Linux may first require you to open the container with `distrobox enter boxnamehere -- bash` to install the shell that your host uses, should you not be using Bash. Otherwise, the `distrobox enter` command may fail.
+
```
89luca89, line 149: A comment on this, 
when $SHELL is set to something non-bash, distrobox-init should take care of this. If it does not we should check distrobox-init for it, from my test, using a different SHELL did not cause any problem, so I think we can remove this comment

89luca89 (changes requested):

docs/compatibility.md
orowith2os: I've removed the warning now that it works.

89luca89 (approved):

PR #361, opened 2022-07-26
docs: fix a mistake for gentoo dockerfile

top1Ashier:
There is an error message when using current dockerfile of gentoo.

> invalid from flag value portage: pull access denied for portage, repository does not exist or may require 'docker login': denied: requested access to the resource is denied

Obviously, the "portage" in the last line doesn't exist. Indicate gentoo/portage as portage to fix it.


89luca89:
Thanks a lot!

Issue #362, opened 2022-07-27
[Suggestion] Add a flag to `distrobox-host-exec` to skip the flatpak-spawn confirmation message.


emmetoneillpdx:
**Is your feature request related to a problem? Please describe.**

I have a setup where a script inside of my distrobox container is using `distrobox-host-exec` to open a browser on the host (with `gio open [url]`). I understand that it's better to use `flatpak-spawn` to achieve what I'm going for here, but it isn't available in the slightly outdated version of Ubuntu LTS that my container is based on. This container is designed to be shared with other people however, and I feel that the forced flatpak-spawn confirmation message is a bit annoying in that context.

**Describe the solution you'd like**

I'd like to be able to pass a CLI flag to `distrobox-host-exec` (for example, "-y" or whatever) to have it automatically skip past the confirmation message. 

**Describe alternatives you've considered**

For now I'll just have my users deal with the message, but it'd be nicer if they didn't have to.

**Additional context**

We've been putting together a distrobox-based development container for the FOSS painting and animation application, Krita. :smile: 

https://invent.kde.org/eoinoneill/krita-devbox

PS: Distrobox is excellent! :+1:

codexMechanicus:
One way to get around the prompt is to pipline a yes.
Like this:
 `yes | distrobox-host-exec`

codexMechanicus:
Update I've implemented the changes discussed here by [emmetoneillpdx](https://github.com/emmetoneillpdx) in my fork and submitted a merge request.

I've tried to keep the code high quality, and confirming to distrobox's conventions with comments and submitted a pull request with the changes. 
I added a --no-confirm/-nc flag. it doesn't force using host-spawn it just skips the prompt and tries to install host-spawn if flatpak and host-spawn are not installed.

I updated the feedback and added information about the --no-confirm flag to the --help flags feedback.

codexMechanicus:
Update, @89luca89 approves of the flag. The pull request could be merged. :smile:

89luca89:
Solved by #363

PR #363, opened 2022-07-27
An implementation of the suggestion from issue #362 "Add a flag to distrobox-host-exec to skip the flatpak-spawn prompt"

codexMechanicus:
See more details of the suggestion here. https://github.com/89luca89/distrobox/issues/362

Add a flag to distrobox-host-exec to skip the flatpak-spawn prompt. 

I've tried to keep the code high quality, and confirming to distrobox's conventions with comments and submitted a pull request with the changes. 
I added a --no-confirm/-nc flag. it doesn't force using host-spawn it just skips the prompt and tries to install host-spawn if flatpak and host-spawn are not installed.

I've made some minor changes to the wording of the prompt, including a line that directs people to the docs and added information about new flag to the options listed by the --help flag.

### DIFF
--- a/distrobox-host-exec
+++ b/distrobox-host-exec
@@ -22,9 +22,9 @@ trap '[ "$?" -ne 0 ] && printf "\nAn error occurred\n"' EXIT
 
 # Defaults
 host_command=""
-install_host_spawn=""
+non_interactive=0
 distrobox_host_exec_default_command="${SHELL:-/bin/sh}"
-host_spawn_version="1.0.3"
+host_spawn_version="1.1.0"
 verbose=0
 version="1.3.2"
 
@@ -51,9 +51,9 @@ Options:
 	--help/-h:		show this message
 	--verbose/-v:		show more verbosity
 	--version/-V:		show version
-    --no-confirm/-nc:   skip prompt so the host-spawn will be installed
-                        on guest system automatically, if neither
-                        flatpak-build or host-spawn are detected
+    --yes/-Y:      Automatically answer yes to prompt: 
+                        host-spawn will be installed on the guest system
+                        if neither flatpak-build or host-spawn are detected
 EOF
 }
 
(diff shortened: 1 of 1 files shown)


distrobox-host-exec
```diff
@@ -22,8 +22,9 @@ trap '[ "$?" -ne 0 ] && printf "\nAn error occurred\n"' EXIT
 
 # Defaults
 host_command=""
+install_host_spawn=""
 distrobox_host_exec_default_command="${SHELL:-/bin/sh}"
-host_spawn_version="1.0.2"
+host_spawn_version="1.0.3"
```
89luca89, line 27: This should be 1.1.0 seeing here: https://github.com/1player/host-spawn/releases

distrobox-host-exec
```diff
@@ -75,6 +79,10 @@ if [ -z "${host_command}" ]; then
 				printf "distrobox: %s\n" "${version}"
 				exit 0
 				;;
+			-nc | --no-confirm)
```
89luca89, line 82: I'm ok with the idea, can you use `-Y/--yes` instead like in other parts of distrobox? check `distrobox-rm` for a reference
also using `0` and `1` instead of yes and no :+1:

89luca89 (changes requested):

89luca89:
Thanks @codexMechanicus left a couple of comments but it's ok having this type of flag for me :+1:

distrobox-host-exec
codexMechanicus: >I'm ok with the idea, can you use -Y/--yes instead like in other parts of distrobox? 
Why/--yes I can. :smile: 

I've implemented and tested the changes locally and will update the pull request.

89luca89 (approved):

89luca89:
Brilliant thanks!

PR #364, opened 2022-07-27
enter: fix automatic container creation when '-r' is used

dfaggioli:
When calling 'distrobox-create', from inside 'distrobox-enter' (in cause
automatic creation of the container is enabled) we were not "forwarding"
the '-r' flag properly. Therefore, when 'distrobox enter -r' was used,
we were trying to create a rootless container, which is obviously wrong!

Fixes: 8b195e3328d0a6fb19564555dec067c607fa8116
Signed-off-by: Dario Faggioli <dfaggioli@suse.com>


89luca89:
Brilliant thanks!

PR #365, opened 2022-07-28
Add some checks for fixing issues due to dirs or manpages missing (on openSUSE MicroOS)

dfaggioli:
Check a bit better the existence of a couple of things, such as directories and manpage entries, to avoid errors on distros where they may not exist (e.g., openSUSE MicroOS).

### DIFF
--- a/distrobox
+++ b/distrobox
@@ -82,7 +82,7 @@ case "${distrobox_command}" in
 		;;
 	help | --help | -h)
 		if command -v man > /dev/null; then
-			man distrobox
+			man distrobox || show_help
 			exit 0
 		fi
 		show_help
--- a/distrobox-export
+++ b/distrobox-export
@@ -303,6 +303,11 @@ export_binary() {
 #	or error code.
 export_application() {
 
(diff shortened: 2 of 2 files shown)


distrobox
```diff
@@ -51,6 +49,19 @@ if [ $# -eq 0 ]; then
 	exit
 fi
 
+# Handle 'help' here, before setting 'errexit', so we have a chance
+# to show our help if the man command is there but fails.
```
89luca89, line 53: Instead of moving it before, how about using

`man distrobox || show_help`

This will ensure the show help after to be executed if man fails

89luca89:
Very cool thanks 😄 
Just a comment on the help, while it makes sense to not exit if man fails, and at least display the help, probably it's better to leave the flag management where it is

distrobox
dfaggioli: Sure. And done. :-)

89luca89 (approved):

89luca89:
Thanks!

PR #368, opened 2022-08-02
Add Nix System Path and Initialize a Nix Flake

FedX-sudo:
I added a new NixOS path to `distrobox-create` as it was mounting all the system bin files into the container, and for easier development purposes with Nix, I added a basic flake, which still needs a bit of work, but can be very helpful in debugging.


FedX-sudo:
Sorry for the bad formatting, I cannot re-produce the error on my system...

89luca89:
Thanks @FedX-sudo for the PR!

I'm not convinced about including the flake in the main repo
I'm more keen to keep the official installation process to be either the curl/install or the maintained packages

Regarding the additional mountpoint:

I see you use `realpath` to resolve the symlink, this should not be needed as podman already solves it during start, and avoiding using realpath should ensure in future that the container will start even if the symlink source changes

FedX-sudo:
The flake is something I wrote because it's difficult to test on NixOS  as Nix does not let you execute arbitrary binaries, and you must use something sanctioned by NixOS. Essentialy, I worte it because I needed it to for development, and figured it would be good to include. 

As far as usign `realpath` I did that because I noticed it was used here: https://github.com/89luca89/distrobox/blob/main/distrobox-create#L454-L457

89luca89:
> The flake is something I wrote because it's difficult to test on NixOS as Nix does not let you execute arbitrary binaries, and you must use something sanctioned by NixOS. Essentialy, I worte it because I needed it to for development, and figured it would be good to include because someone else may want it, and/or have to re-write it latter down the road.

Understood, maybe we can mention this as a snippet (gist) in the documentation, but I'd prefer to keep the repo distro-agnostic

> As far as usign `realpath` I did that because I noticed it was used here: https://github.com/89luca89/distrobox/blob/main/distrobox-create#L454-L457

Yep because that's a mountpoint that won't change
I was under the impression that nix uses symlinks for stuff that will change (like the stores, versions and such), if that's the case, I'd let podman resolve it so that we won't need to re-build the container when the link will change

89luca89:
@FedX-sudo cool thanks, do you know if/when `/run/current-system/sw` changes during Nix lifecycle?
Can we test if the container survives the change of that directory's link? thanks!

FedX-sudo:
It changes every system update, and I have been testing this PR for a bit without issue, but I did not check to see if it updates with a full NixOS update...

89luca89:
Thanks a lot @FedX-sudo  :heart:

PR #376, opened 2022-08-14
install: use either wget or curl

gcoakes:
Some environments such as Ubuntu 22.04 don't have curl by default. Use
wget if curl is unavailable.


89luca89:
That's a good idea thanks for the contribution @gcoakes!

PR #383, opened 2022-08-23
docs - Clarify smallness of common cloud images

castrojo:
Desktop users might not know that their favorite distro's cloud image is tiny. Clarified post-creation instructions.


89luca89:
Thanks a lot ❤️

PR #388, opened 2022-08-29
Update README.md

profetik-777:
Given that a major value of distrobox is leveraging various distro/environments, I recommended including the command on how to install non-default distroboxes along w/ a link to review a list of images a user can try out other than the one listed in the example.


profetik-777:
Created a new one moving this entry up to a few lines.

PR #394, opened 2022-09-03
create: mount the store directory of Guix.

cireu:


89luca89:
Looks good, thanks for contributing!

PR #404, opened 2022-09-09
create: add --no-entry

misobarisic:
I have found myself not always needing the automatic application entry generation so I implemented a way to disable it


89luca89:
@misobarisic 
Looks good to me!

Just wondering if maybe `no-icon` would be more explanatory for the user? Not sure on this 🤔

misobarisic:
A `.desktop` file does not need an icon to be specified. Wouldn't calling it `no-icon` be misleading in that context?

89luca89:
LGTM thanks a lot @misobarisic

PR #414, opened 2022-09-13
Redundant "else" in readme

misobarisic:
The placement of `else` does not make sense with the flow of the chapters. `Alternative methods` does not contain a prior statement which should be "else-d".


89luca89:
Thanks!

PR #419, opened 2022-09-17
Fix using --home with space in path.

bsvh:
When you try to use `distrobox-create --home` with a home directory that has a space in it, it errors out. This fixes that.


89luca89:
Nice catch :smile: 

Thanks for your contribution!

PR #422, opened 2022-09-18
Add home prefix

misobarisic:
I've added the ability to use a home prefix. An equivalent to defining a custom `/home` partition. Does not have precedence over custom homes set with `--home`.

If a home directory, either created with `--home` or `--prefix`, does not exist, it will be created.


89luca89:
Hey @misobarisic 

I'm not sure of this, in which way this is different than simply specifying `--home/-H`?

misobarisic:
It is a directory where all container homes would live. Instead of specifying `--home` for every box explicitly, it would automatically be bound to one if the environment variable was set. Think of it as `/home/user` where `home` is the prefix and `user` the container name.

misobarisic:
@89luca89 I agree. Now that I think about it, having a flag for a specific prefix is redundant. It should depend on environment variables instead.

89luca89:
Cool

At this point I'd document this a bit better inside the `usage` of `create` and maybe an entry in `useful_tips` so we've covered the feature :smile:

misobarisic:
Should these autogenerated home directories be removed when the container gets deleted by `distrobox-rm`?

89luca89:
> Should these autogenerated home directories be removed when the container gets deleted by `distrobox-rm`?

Yea was thinking of it
I'd say, we should ask confirmation (following the behaviour with `-f` and `--yes` flags obviously)

misobarisic:
> > Should these autogenerated home directories be removed when the container gets deleted by `distrobox-rm`?
> 
> Yea was thinking of it I'd say, we should ask confirmation (following the behaviour with `-f` and `--yes` flags obviously)

I'm also thinking about prompting for confirmation to delete the home dir of containers which use `--home`. Easier implementation too. Just checking if the container home differs from the host.

89luca89:
> > > Should these autogenerated home directories be removed when the container gets deleted by `distrobox-rm`?
> > 
> > 
> > Yea was thinking of it I'd say, we should ask confirmation (following the behaviour with `-f` and `--yes` flags obviously)
> 
> I'm also thinking about prompting for confirmation to delete the home dir of containers which use `--home`. Easier implementation too. Just checking if the container home differs from the host.

Agree!

misobarisic:
I've added an environment variable `DBX_CONTAINER_RM_CUSTOM_HOME` which is responsible for autodeleting the custom home of containers.

Interactive sessions ask for confirmation when `DBX_CONTAINER_RM_CUSTOM_HOME` or `--rm-home` haven't been specified.
Non interactive sessions delete only when either of the two have been specified and are _truthy_.

misobarisic:
@89luca89 is something blocking the merge as of now?

PR #430, opened 2022-09-21
Experiment/library

89luca89:
This PR wants to implement some changes about how this script is structured:

1 - separate common logic in importable functions
2 - create as many functions as possible to make it more modular
3 - implement some rudimentary unit tests in order to ensure some basic solidity

The changes lead to quite a lot of code deduplication, improve legibility and modularity

Separating the container creation, cloning, entering in functions, is also useful if someone wants in the future create another interface/manager based on distrobox


FedX-sudo:
I am having a hard time executing the `/lib` files. This may be a Nix issue, but it appears to be looking for lib in a reletive `../` from the current dirrectory, as opposed to the absolute `lib` path.

89luca89:
> I am having a hard time executing the `/lib` files. This may be a Nix issue, but it appears to be looking for lib in a reletive `../` from the current dirrectory, as opposed to the absolute `lib` path.

Yea that's one of my main concerns

Generally speaking I was thinking about discovering those files as they usually are in `$PREFIX/bin` (be it ~/.local or /usr/local)
So doing a ../lib, should lead to $PREFIX/lib

Hardcoding it is not an option obviously,
What path does it try to search in your case?

FedX-sudo:
In my case, it's in `/nix/store/r236jfc07c2rrqvd42hdi8a38cq8mp3w-distrobox-20220921/lib`. Typicaly in Nix this path is hard-coded into the installed script...

89luca89:
> In my case, it's in `/nix/store/r236jfc07c2rrqvd42hdi8a38cq8mp3w-distrobox-20220921/lib`. Typicaly in Nix this path is hard-coded into the installed script durring install...

Was assuming that you wouod have something like

/nix/store/r236jfc07c2rrqvd42hdi8a38cq8mp3w-distrobox-20220921/lib
/nix/store/r236jfc07c2rrqvd42hdi8a38cq8mp3w-distrobox-20220921/bin

Is it not right?

FedX-sudo:
That is correct, Nix does `/nix/store/<hash>-<pname>-<version>` and outputs bin to `$out/bin` and lib to `$out/bin`

89luca89:
> That is correct, Nix does `/nix/store/<hash>-<pname>-<version>` and outputs bin to `$out/bin` and lib to `$out/lib`

Ok, in the import I do 

`. "$(dirname "$(realpath "${0}")")/`

which should give:

`/nix/store/r236jfc07c2rrqvd42hdi8a38cq8mp3w-distrobox-20220921/bin`

and from there, `../lib`, should point to `/nix/store/r236jfc07c2rrqvd42hdi8a38cq8mp3w-distrobox-20220921/lib` 
and the libraries are not there?

FedX-sudo:
Unfortunatly that does not work either. It's now looking in `/nix/store/7769xj12nhgb2hffpyn5lydbvlwlf3zd-distrobox-20220921/bin/lib/config-load`. This is probably an issue for nixpkgs to solve...

89luca89:
> Unfortunatly that does not work either. It's now looking in `/nix/store/7769xj12nhgb2hffpyn5lydbvlwlf3zd-distrobox-20220921/bin/lib/config-load`. This is probably an issue for nixpkgs to solve...

mmmh can you `tree /nix/store/7769xj12nhgb2hffpyn5lydbvlwlf3zd-distrobox-20220921/` so we can see how the dirs are lay out?

89luca89:
> Well this is most interesting. After having reverted to `HEAD` and using the updated installer, it's now working perfectly for me...

Not sure what happened, but it works now?

FedX-sudo:
Yes. I had changed it to `. "$(dirname "$(realpath "${0}")")/` in the codebase, I reverted to `. "$(dirname "$(realpath "${0}")")/../`, and changed the `flake.nix` I was using to overwrite the `lib` paths to the `/nix/store` path...

PR #431, opened 2022-09-22
Remove shebangs from bash completions

michel-slm:
These files start with the #! sequence that marks interpreted scripts, but they are bash completion scripts that is merely intended to be sourced.

See https://lintian.debian.org/tags/bash-completion-with-hashbang

Signed-off-by: Michel Alexandre Salim <michel@michel-slm.name>


89luca89:
Thanks as always :heart:

PR #441, opened 2022-09-26
Do not open manpager automatically

misobarisic:
`distrobox --help` checks if man is present and acts as if `man distrobox` was executed by the user. Other commands do not exhibit this behaviour and I believe the wrapper should not be an outlier.
My belief is if a user wanted to open a manpage they would have called `man` themselves.


89luca89:
Makes sense yes :+1:

PR #449, opened 2022-09-30
compatibility: add gentoo docker image

89luca89:
Turns out that we can use the `stage3` image as a normal container. This (with the addition of `emerge --sync`) can solve the problem of providing and testing gentoo in the CI.

Signed-off-by: Luca Di Maio <luca.dimaio1@gmail.com>

### DIFF
--- a/distrobox-init
+++ b/distrobox-init
@@ -379,10 +379,20 @@ if [ "${upgrade}" -ne 0 ] || ! command -v find || ! command -v mount || ! comman
 		fi
 
 	elif command -v emerge; then
+		# Check if the container we are using has a ::gentoo repo defined,
+		# if it is defined and it is empty, then synchroznize it.
+		gentoo_repo="$(portageq get_repo_path / gentoo)"
+		if [ -n "${gentoo_repo}" ] && [ ! -e "${gentoo_repo}" ] ; then
+			emerge-webrsync
+		fi
+		# If we need to upgrade, do it and exit, no further action required.
+		if [ "${upgrade}" -ne 0 ] ; then
+			emerge --sync
+			exit
+		fi
 		# Check if shell_pkg is available in distro's repo. If not we
 		# fall back to bash, and we set the SHELL variable to bash so
 		# that it is set up correctly for the user.
-		emerge --sync --quiet
 		if ! emerge --ask=n --autounmask-continue --noreplace --quiet-build "${shell_pkg}"; then
 			shell_pkg="bash"
 		fi
--- a/distrobox-init
+++ b/distrobox-init
@@ -382,11 +382,11 @@ if [ "${upgrade}" -ne 0 ] || ! command -v find || ! command -v mount || ! comman
 		# Check if the container we are using has a ::gentoo repo defined,
 		# if it is defined and it is empty, then synchroznize it.
 		gentoo_repo="$(portageq get_repo_path / gentoo)"
-		if [ -n "${gentoo_repo}" ] && [ ! -e "${gentoo_repo}" ] ; then
+		if [ -n "${gentoo_repo}" ] && [ ! -e "${gentoo_repo}" ]; then
 			emerge-webrsync
 		fi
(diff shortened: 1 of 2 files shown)


xgqt:
`emerge --sync` without `--quiet` produces a lot of noise (it will list out all ~20000 synced package names)
I wonder if we could use `emaint sync` (a different command for system maintenance provided by portage) for a not-compeltely-slitent sync

89luca89:
@xgqt we can track here the support

We need some missing things:

- a way to provide vte-profile
- lsof
- gnupg2

Nice to have:

- mesa/vulkan libraries

We can set mesa to be optional, but the first three have to work to provide an on-par experience with other offerings

89luca89:
> `emerge --sync` without `--quiet` produces a lot of noise (it will list out all ~20000 synced package names) I wonder if we could use `emaint sync` (a different command for system maintenance provided by portage) for a not-compeltely-slitent sync

I think we can use `--sync`, the logs of `init` use `set -x` so in the logs a failed sync should be detectable anyway

xgqt:
> lsof

in `sys-process/lsof` it is a small pkg so definitely can be added

> gnupg2

This would probably fit within the time limit

`time emerge -av --quiet-build app-crypt/gnupg` on clean image

```
real    1m17.138s
user    3m11.037s
sys     0m39.295s
```

> a way to provide vte-profile

unfortunately on Gentoo vte-profile (`/etc/profile.d/vte-2.91.sh`) is included inside the vte package which will pull in a the Xorg stack which will take probable so long (~2h mainly because of rust) that it will time-out the GitHub CI.

89luca89:
> > lsof
> 
> in `sys-process/lsof` it is a small pkg so definitely can be added
> 
> > gnupg2
> 
> This would probably fit within the time limit
> 
> `time emerge -av --quiet-build app-crypt/gnupg` on clean image
> 
> ```
> real    1m17.138s
> user    3m11.037s
> sys     0m39.295s
> ```
> 
> > a way to provide vte-profile
> 
> unfortunately on Gentoo vte-profile (`/etc/profile.d/vte-2.91.sh`) is included inside the vte package which will pull in a the Xorg stack which will take probable so long (~2h mainly because of rust) that it will time-out the GitHub CI.

Ok added lsof and gnupg2 thanks :smile: 

So for now I'd exclude vte, do you happen to know how much does it take for mesa/vulkan support?

xgqt:
> > > lsof
> > 
> > 
> > in `sys-process/lsof` it is a small pkg so definitely can be added
> > > gnupg2
> > 
> > 
> > This would probably fit within the time limit
> > `time emerge -av --quiet-build app-crypt/gnupg` on clean image
> > ```
> > real    1m17.138s
> > user    3m11.037s
> > sys     0m39.295s
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > > a way to provide vte-profile
> > 
> > 
> > unfortunately on Gentoo vte-profile (`/etc/profile.d/vte-2.91.sh`) is included inside the vte package which will pull in a the Xorg stack which will take probable so long (~2h mainly because of rust) that it will time-out the GitHub CI.
> 
> Ok added lsof and gnupg2 thanks smile
> 
> So for now I'd exclude vte, do you happen to know how much does it take for mesa/vulkan support?

might take more than 2 hours mainly because of llvm (which is pulled in for the AMD driver)

xgqt:
The PR https://github.com/89luca89/distrobox/pull/450 replaces running `emerge --sync` with running it only when `/var/db/repos/gentoo` has no contents or there is a forced update (from distrobox CLI)

xgqt:
Are we missing anything essential?

89luca89:
Seems like everything is ok now :smile: 

Merging, thanks @xgqt for the help on this!

xgqt:
Thanks a lot for help with testing and listening to my suggestions!

PR #460, opened 2022-10-05
Complete rewrite of exec argument handling to handle spaces in quoted args.

### LINKED ISSUE #439: [Error] Commands generated with distrobox-export do not support whitespace and special characters in arguments
**Describe the bug**
Commands generated with `distrobox-export` do not pass arguments with whitespaces or special characters to containers properly.
This also happens with `.desktop` files generated by distrobox.

**To Reproduce**
I wrote simple python script that prints received arguments
```python
#!/usr/bin/env python

import sys

print(f'Received {len(sys.argv)}:')
for i, arg in enumerate(sys.argv):
    print(f'{i}: {arg}')
```
Then exported with `distrobox-export --bin`. I tried few inputs, but didn't receive expected results
```
$ pathtest "arg with spaces"
Received 4:
0: /home/deck/dev/path-test/pathtest
1: arg
2: with
3: spaces
```
```
$ pathtest 'special "!@#'
/home/deck/.local/bin/distrobox-enter: eval: line 501: unexpected EOF while looking for matching `"'

An error occurred
```

**Expected behavior**
It works fine when calling binary directly, or calling generated script from inside container
```
dev/path-test/pathtest "arg with spaces"
Received 2:
0: dev/path-test/pathtest
1: arg with spaces
```
```
$ dev/path-test/pathtest 'special "!@#'
Received 2:
0: dev/path-test/pathtest
1: special "!@#
```
```
(deck@steamdeck ~)$ distrobox enter arch
[deck@arch ~]$ pathtest 'special "!@#'
Received 2:
0: /home/deck/dev/path-test/pathtest
1: special "!@#
```

 - Are you using podman or docker? podman
 - Which version or podman or docker? 4.2.1
 - Which version of distrobox? 1.4.1 on SteamDeck

**Additional context**
I tried playing around with various escaping strategies, but couldn't find anything that would cover all cases. This one is probably the closest
```shell
#!/bin/sh
# distrobox_binary
# name: arch
if [ ! -f /run/.containerenv ] && [ ! -f /.dockerenv ]; then
    command="/home/deck/.local/bin/distrobox-enter  -n arch -- \" /home/deck/dev/path-test/pathtest "
    for i in "$@"
    do
        command="$command $(printf "%q " "$i")"
    done
    command="$command \""
    echo $command
    eval "$command"
else
    /home/deck/dev/path-test/pathtest "$@"
fi
```
but it still fails the `pathtest 'special "!@#'` test. Maybe best solution would be to just base64 encode input args and have some program inside container that would decode and exec binary directly, skipping the whole shell escaping mess.

HugLifeTiZ:
Fixes #439. Probably also fixes some other things, too; I consistently got errors after entering distrobox environments, and now I don't anymore.

* Moves the entirety of the `generate_command` function out of a function, and over to the only place where it is actually called.
* Completely drops the `container_command` variable, and instead relies on `$@` for preserving spaces.
* Renames `result_command` to `command_prefix`.
* Moving it out of a function means no more `printf` into stdout into command substitution; that was redundant and possibly harmful.
* Uses `eval` to interpret manual quoting inside the `command_prefix` variable when prepending it to `$@`.
* Drops the newlines and tabs from the variable assignments because newlines in `eval` mean "command args are over, here is a new command"
* Simplifies looking up the shell from `/etc/passwd` as a fallback command.

Simple test case for #439:

```
echo 'hiya!' > "/tmp/with-spaces"
distrobox enter $your_fave_container -- cat "/tmp/with spaces"
```

Without this fix, it tries to cat two separate files: "/tmp/with" and "spaces". With this fix, quoting and spacing is preserved. You can call your `distrobox enter` commands naturally, as you would with any other command, without any additional quoting or spacing acrobatics.

### DIFF
--- a/distrobox-enter
+++ b/distrobox-enter
@@ -39,7 +39,6 @@ if [ "$(id -u)" -eq 0 ]; then
 fi
 
 # Defaults
-container_command=""
 container_image=""
 container_image_default="registry.fedoraproject.org/fedora-toolbox:36"
 container_manager="autodetect"
@@ -173,7 +172,8 @@ while :; do
 			;;
 		-e | --exec | --)
 			shift
-			container_command=$*
+			# Do nothing except consume the argument and break. We will
+			# use "$@" to preserve spaces in the remaining args.
 			break
 			;;
 		-*) # Invalid options.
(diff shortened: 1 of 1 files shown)


89luca89:
> Moves the entirety of the generate_command function out of a function, and over to the only place where it is actually called.

Sorry but I cannot do this, having a function is important, and in future everything will be moved to a different "library" file, so this is not possible to merge

89luca89:
> It is of near-crucial importance to be able to send arguments with spaces to distrobox enter the same way you would with any other command. The README in https://github.com/89luca89/distrobox/pull/448 suggests you would have to quote quoted arguments: distrobox enter test -- bash -l -c '"echo \$HOME && whoami"' It's not acceptable to ask users to do that. Accomplishing the same thing outside of the container is done like so: bash -l -c "echo \$HOME && whoami" and it's important for it to work the same way for distrobox enter.

    Though I do think it is acceptable to have to handle arguments to the container manager itself in a specialized way.
    
In that commit I **removed** that as it's now handled :smile: 

> There is one more thing I would like you to think about. The fact that you want to refactor your code into a library suggests that Distrobox has severely outgrown POSIX sh. There is something to be said for using the right tool for the job, and I think that wanting to put functions into a library is a good indicator that you have long-since crossed the line where sh is no longer a good fit. sh is great for single-file scripts that can run anywhere its requisite programs are at, but I think Distrobox is too complex at this point--it has an installation process anyways--and sticking with sh is just making your life harder. I think Python 3 would be a good fit.

The choice of POSIX sh was not for the ease of use (as you can see gives a lot of headaches :sweat_smile: ) but for compatibility. This thing has to run (especially the export, init, host-exec parts) anywhere, without particular dependencies, and spanning across almost 10 years of distros

Anyway, if you take a look at #448 , I've updated it with a "simple" logic to handle special chars, and it's working as intended with the current layout, as you can see here:

image

89luca89:
I will close this, we will continue on #448

HugLifeTiZ:
> and spanning across almost **10 years of distros**

Ah. Right. Yeah, there's the sticking point. Okay, POSIX sh makes even more sense with that particular requirement. There's really no other scripting language that satisfies it; they all have too much churn while POSIX sh doesn't change, ever. The only other way to do it is to write a program with a compiled language and distribute binaries statically linked with musl, and the jump from sh to any compiled language is huge compared to the jump from sh to any other scripting language. That sort of migration is not feasible at all; someone's better off writing a different program compatible with distrobox's CLI.

Well, this was a good talk. Even though this PR wasn't merged, I'm glad it could be of some use to you. :)

89luca89:
@HugLifeTiZ 

Yea but even statically compiled binaries have limits about architectures (32bit? aarch?) and maybe some syscall get deprecated and such
And if you think of it, apart from some minor logic, I'm just assembling and launching either podman/docker commands and/or package managers :joy:  In the end I'd just create a `C-script` if you will lol

Anyway thanks for the talk, it inspired me to explain these things in the docs, because it is effectively a bizarre/unusual choice and it deserves to be explained :+1:

PR #461, opened 2022-10-07
Update handler function to support bash scripts

gagallo7:
`command_not_found_handler` only checks for `$ZSH_VERSION` environment variable. To support bash as well, we need to check for `$BASH_VERSION` environment variable.

Plus, give a tip to ensure that the change has an effect as we do in the `fish` section.

### DIFF
--- a/docs/posts/execute_commands_on_host.md
+++ b/docs/posts/execute_commands_on_host.md
@@ -82,7 +82,7 @@ command_not_found_handle() {
   
   distrobox-host-exec "${@}"
 }
-if [ -n "${ZSH_VERSION-}${BASH_VERSION-}" ]; then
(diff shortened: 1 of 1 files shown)


89luca89:
Hi @gagallo7 

Bash uses `command_not_found_handle` while Zsh uses `command_not_found_handler`, notice the final `r` :sweat_smile: 
as specified here: https://www.gnu.org/software/bash/manual/html_node/Command-Search-and-Execution.html

So bash was already covered by declaring the first function, the zsh check is just to bind that `command_not_found_handle` function to `command_not_found_handler` so that also zsh uses it

So this check is not necessary
But we can keep the reload tip :+1:

PR #464, opened 2022-10-08
export: look at global user services while trying to export a service

bellegarde-c:


89luca89:
Thanks a lot @bellegarde-c very useful :smile:

PR #474, opened 2022-10-11
Fix cni_plugin_dirs

descention:
I was having an issue with networking and found the configuration for cni_plugin_dirs wasn't being referenced. The configuration document suggests it needed to be in the [network] section. After moving it there, cni was able to find the plugins it said were missing.

https://github.com/containers/common/blob/main/docs/containers.conf.5.md#network-table


89luca89:
Nice catch, thanks! @descention

PR #475, opened 2022-10-12
Small clarifications

damianoognissanti:
When trying the GNOME instructions (on Debian 11 Stable) I had to:
- manually create the folder /run/dbus in fedora-rawhide.
- add the flag -f to chown (otherwise it gave an `Operation not permitted` error) in fedora-rawhide.
- add the .desktop file to `/usr/share/xsessions` instead of `/usr/local/share/wayland-sessions/` for it to show in the login window. 

I also made the word host bold when talking about adding the .desktop entry, even if that should be obvious. Since all other instructions are inside fedora-rawhide, one might forget to exit before that step.

(Written from GNOME 43 on Fedora-Rawhide on Debian 11 Stable)

### DIFF
--- a/docs/posts/run_latest_gnome_kde_on_distrobox.md
+++ b/docs/posts/run_latest_gnome_kde_on_distrobox.md
@@ -49,6 +49,8 @@ user@fedora-rawhide:~$ sudo ln -s /run/host/run/systemd/system /run/systemd
 user@fedora-rawhide:~$ sudo ln -s /run/host/run/dbus/system_bus_socket /run/dbus/
 ```
 
+If you get an error message, you might have to manually create the folders `/run/systemd` and `/run/dbus` with `sudo mkdir`.
+
 Then we can proceed to install GNOME in the container:
 
 ```shell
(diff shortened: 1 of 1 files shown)


89luca89:
Thanks a lot @damianoognissanti 
That's really useful, thanks :smile:

docs/posts/run_latest_gnome_kde_on_distrobox.md
```diff
@@ -92,6 +94,7 @@ X-GDM-SessionRegisters=true
 ```
 
 This file should be placed under `/usr/local/share/wayland-sessions/distrobox-gnome.desktop`
+(If it doesn't show up, you can place it under `/usr/share/xsessions/distrobox-gnome.desktop`)
```
bam80, line 97: Hmm, may I guess - you have SDDM on host?

bam80: @damianoognissanti then if it's GDM it should be no problem to put it by `/usr/local/...` path.
Did you try to investigate why it's not for your case?

damianoognissanti: @bam80 Yeah, I know, that was very odd. Since it worked with this fix I let it go and thought the comment could help someone where the standard path wouldn't work.

bam80: Maybe your GDM runs on X11 and thus doesn't pick up the Wayland session files?

PR #478, opened 2022-10-14
Add --network option

Cloudwalk9:
This patch experimentally implements the --network option to use custom networking other than the default "host". This generally works fine for networks created using "podman network create".

It's otherwise a relatively crude, experimental implementation, with limitations:

* There is no bash completion yet. The obvious solution is to check for docker or podman in the completion script itself and print the output of `podman network ls --format "{{.Name}}"`, but I wasn't sure about it.
* It requires --init. I could not get it to work properly without --init. The container has no Internet access after setup, probably because /etc/resolv.conf and /etc/hosts are mounted. I wasn't able to find a clean way to not mount those when --init is absent.

I have not tested this with rootful containers, and other networking modes such as "bridge" and "ns:". For the latter, I'd expect those to work fine.

### DIFF
--- a/distrobox-create
+++ b/distrobox-create
@@ -58,6 +58,8 @@ container_user_gid="$(id -rg)"
 container_user_home="${HOME:-"/"}"
 container_user_name="${USER}"
 container_user_uid="$(id -ru)"
+container_network=""
+container_network_default="host"
 distrobox_sudo_program="sudo"
 dryrun=0
 init=0
@@ -143,6 +145,7 @@ Options:
 				of the same environment.
 	--home/-H:		select a custom HOME directory for the container. Useful to avoid host's home littering with temp files.
 	--volume:		additional volumes to add to the container
+	--network:		select a network namespace for the container
 	--additional-flags/-a:	additional flags to pass to the container manager command
 	--init-hooks:		additional commands to execute during container initialization
 	--pre-init-hooks:	additional commands to execute prior to container initialization
@@ -258,6 +261,13 @@ while :; do
 				shift
 			fi
 			;;
+		--network)
+			if [ -n "$2" ]; then
+				container_network="${2}"
+				shift
+				shift
+			fi
+			;;
 		--) # End of all options.
 			shift
 			break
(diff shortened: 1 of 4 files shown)


distrobox-init
```diff
@@ -1006,11 +1015,14 @@ eval ${init_hook}
 if [ "${init}" -eq 0 ]; then
 	printf "container_setup_done\n"
 	# Keepalive loop
-	HOST_WATCH="
-		/etc/hosts
-		/etc/localtime
-		/etc/resolv.conf
-	"
+	HOST_WATCH="/etc/localtime"
+	if [ "${custom_networking}" -eq 0 ]; then
```
89luca89, line 1019: Technically we can leave these untouched
If it's a custom network, they will not be mountpoints, so the subsequent check will not pick them up for sync

89luca89:
Thanks a lot @Cloudwalk9 

Seems very in line with what I would like to implement for some **very basic** decoupling/sandboxing from the host

I'd like to test it more so I'll keep this for a bit as an experiment if you don't mind :smile: 

This would be a nice starting point for the network :+1:

Cloudwalk9:
> Thanks a lot @Cloudwalk9
> 
> Seems very in line with what I would like to implement for some **very basic** decoupling/sandboxing from the host
> 
> I'd like to test it more so I'll keep this for a bit as an experiment if you don't mind smile
> 
> This would be a nice starting point for the network +1

Thanks, that means a lot. Happy to help c:

And no problem, take your time.

thinkingBanana:
Using it right now and it seems to be working, thank you so much. Two potential improvement suggestions:
1. Currently if I start a new box without the "--network" flag, the following error would occur:
> Network "host" does not exist.
> for a list of networks, run:
>
>podman network ls
>

2. If I force remove the distrobox without properly shutting it down, the network device will not be removed correctly and therefore remained in the system with the following message:
>ERRO[0000] Unable to clean up network for container xxxxxx: "error unmounting network namespace for container xxxxxx: fail to remove ns path /run/user/1000/netns/yyyyyy: remove /run/user/1000/netns/yyyyyy: device or resource busy"

distrobox-init
Cloudwalk9: If I recall, I also expected they wouldn't be mount points but networking still broke for some reason. Should be a temporary workaround until that's looked into.

89luca89:
Hi @Cloudwalk9 there are some updates, speaking on another PR #609 
Can we check if we can work together on a solution?

Cloudwalk9:
The part of my brain responsible for git merges atrophied apparently.

Cloudwalk9:
That's more like it.

bubuntux:
just FYI i have been testing this feature and works great !

bubuntux:
are extra changes need it to cover this new parameters for distrobox-assemble?

PR #501, opened 2022-11-05
Fixed toolbox URL

thegamerhat:
Syntax mistake caused the "https://github.com/containers/toolbox" URL to have an extra `),` at the end, which makes the URL incorrect, and leads to a 404 page on Github instead of the real project page.


89luca89:
Thanks a lot!

Issue #476, opened 2022-10-13
[Discussion] Any way to pass flags to `podman run` command?


FSund:
I'm trying to figure out if it's possible to pass additional flags along to the `podman run` command when starting a distrobox. Specifically I'm looking to use the `--cpuset-cpus` [flag](https://docs.podman.io/en/latest/markdown/podman-run.1.html#cpuset-cpus-number) to limit the number of cpus the container has access to. Is this possible?

My current attempts got this far

```
$ distrobox enter -a "--cpuset-cpus='0-1'" --name testing
Container testing is not running.
Starting container testing
run this command to follow along:

 podman logs -f testing

 Starting container...                  	 [ OK ]
 Installing basic packages...           	 [ OK ]
 Setting up read-only mounts...         	 [ OK ]
 Setting up read-write mounts...        	 [ OK ]
 Setting up host's sockets integration...	 [ OK ]
 Integrating host's themes, icons, fonts...	 [ OK ]
 Setting up package manager exceptions...	 [ OK ]
 Setting up dpkg exceptions...          	 [ OK ]
 Setting up apt hooks...                	 [ OK ]
 Setting up sudo...                     	 [ OK ]
 Setting up groups...                   	 [ OK ]
 Setting up users...                    	 [ OK ]
 Executing init hooks...                	 [ OK ]

Container Setup Complete!
Error: unknown flag: --cpuset-cpus
See 'podman exec --help'

An error occurred

An error occurred
```

thjderjktyrjkt:
I guess you need to pass that argument when creating distrobox (`distrobox-create -a`), since OCI container doesn't allow changing its environment after being created. Alternatively you may use SystemD resource control to limit the amount of CPUs it can use on the go (check `man systemd.resource-control`), you'll need to use unit override file.

FSund:
> I guess you need to pass that argument when creating distrobox (`distrobox-create -a`), since OCI container doesn't allow changing its environment after being created. 

I was suspecting that I had to do it on creation. That's not compatible with the workflow I'm using, unfortunately.

> Alternatively you may use SystemD resource control to limit the amount of CPUs it can use on the go (check `man systemd.resource-control`), you'll need to use unit override file.

That sounds promising. Thanks!

89luca89:
@FSund 
As @thjderjktyrjkt suggested, there is a dedicated flag for it, but it needs a container re-creation

Neither podman nor docker support on-the-fly resource assignation/restriction so distrobox cannot provide it sadly

PR #515, opened 2022-11-15
[Docs] Add resource limitation tips

thjderjktyrjkt:
Inspired by this issue: https://github.com/89luca89/distrobox/issues/476#issuecomment-1277486593

Why not add some tutorials for it?

---

Problems:

- Only works with Podman. I don't use Docker so I don't know how Docker works with SystemD.
- Only works with SystemD. I don't use other init systems, I don't even know if alternate init systems provide similar functionality or not.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -88,6 +88,7 @@ graphical apps (X11/Wayland), and audio.
   - [Run latest GNOME and KDE Plasma using distrobox](posts/run_latest_gnome_kde_on_distrobox.md)
   - [Integrate VSCode and Distrobox](posts/integrate_vscode_distrobox.md)
   - [Execute a command on the Host](posts/execute_commands_on_host.md)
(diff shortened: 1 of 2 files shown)


thjderjktyrjkt:
@FSund Not sure if you still need it but anyway tutorial is here.

thjderjktyrjkt:
@89luca89 maybe something similar could be baked into distrobox directly? these are just a bunch of commands after all, and helping users doing it would be a QoL improvement. Not sure how to properly represent all the resource control functionalities though.

89luca89:
@thjderjktyrjkt 

Thanks a lot for this!
I've pushed some fixes to pass the CI.

> @89luca89 maybe something similar could be baked into distrobox directly? these are just a bunch of commands after all, and helping users doing it would be a QoL improvement. Not sure how to properly represent all the resource control functionalities though.

Can we open a feature request for this? We can discuss there :+1:

PR #541, opened 2022-11-29
use --entrypoint to override entrypoint defined by image

tomeichlersmith:
This is a small change but is helpful for users with specific images that have their own entry point defined for another purpose but building a distrobox using the same image would be helpful for debugging purposes. 

I've tested this on Ubuntu 22.04 running docker and I've checked that [podman-create](https://docs.podman.io/en/latest/markdown/podman-create.1.html) has the same option.


89luca89:
Thanks for the contribution! @tomeichlersmith

Issue #493, opened 2022-11-01
[Suggestion] Better support cases where an HTTPS Proxy is present


mtalexan:
**Is your feature request related to a problem? Please describe.**
When entering a newly created distrobox instance, the `distrobox-init` is run to install the minimum necessary packages on top of the standard public distro images.  However many users in corporate environments have to pass thru an HTTPS proxy to reach these servers, and require custom CA certificates to do so.

**Describe the solution you'd like**
Add support for a custom certificates folder to the `distrobox.conf`.  Mount this read-only into all containers, and during `distrobox-init` add the certificates to the trust store/anchors prior to running any package manager commands.

**Describe alternatives you've considered**
The `--pre-init-hook` option could be used to do this with a custom script that uses the `mount_bind` function in `distrobox-init`.  However writing this to cover every supported distro isn't trivial and would do better if it were implemented consistently within distrobox itself given how common a use case this is likely to be.  Additionally it requires the user manually adding the proper `--pre-init-hook` to every `distrobox-create` call.

Creating custom distro container images that include the extra certificates is technically an alternative.  
This runs into the same issue that `distrobox-init` solves however.  Now custom images for every possible distro need to be maintained so the extra proxy CA certificates can be added.  

--- 

EDIT: I discovered the minimally documented `--pre-init-hook` and `--init-hook` options in `distrobox-create` and `distrobox-init`, which are very similar to what I originally described as one of the alternative options to add.

mtalexan:
I just realized from reading the `distrobox-init` code that there's a minimally documented option `--pre-init-hook` that allows you to specify a custom script to run prior to the distrobox-init running.  
However the user workflow is expected to be:
1. `distrobox create`
2. `distrobox enter`
`distrobox-init` is usually called automatically (I can't even find where), and the `--pre-init-hook` option is not available from the `distrobox-enter` that will normally invoke `distrobox-init`.  Also, the `--pre-init-hook` is expected to run on the host system and is responsible for starting the container and executing commands within it if that's part of what's necessary, making it extremely involved for the simple and common case of needing to include extra CAs in the container.

mtalexan:
From playing around with it, it appears at least Debian and Ubuntu images are cloud images that don't have even ca-certificates or basic tools installed in them when downloaded, but also do unauthenticated installation of packages (at least on first run).  For those the pre-init-hook can't do anything to help, but the init-hook could.

[...]

PR #542, opened 2022-11-29
Add init and pre-init hook defaults as supported config options

### LINKED ISSUE #497: [Error] --pre-init-hooks doesn't match --init-hooks screwing up .distroboxrc config
**Describe the bug**
Using `--pre-init-hooks` on the `distrobox create` command line sets the `container_pre_init_hook` variable to `--pre-init-hooks=${2}`, while using `--init-hooks` sets `container_init_hook` to `${2}`.
Both of these hooks can be set in the distroboxrc file with their appropriate variables, but one requires to you manually add the docker/podman command-line switch to it as well and the other doesn't.

**To Reproduce**
Examine https://github.com/89luca89/distrobox/blob/main/distrobox-create#L296

**Expected behavior**
The `--pre-init-hook=` part should be prepended in code right before it's used on [line 633](https://github.com/89luca89/distrobox/blob/main/distrobox-create#L633) rather than added during command-line argument parsing.

mtalexan:
Since the `--init-hook` and `--pre-init-hook` are the current method for resolving custom TLS certificates being needed by corporate proxies (see #493), being able to configure fixed hooks that don't need to be manually added on the command-line each time are useful.  

The variables already existed and could be overridden by the config files (which are just sourced into the scripts), but the `--pre-init-hook` was processed strangely and was appending the arguments needed when calling docker at argument parsing-time instead (see #497).  The command-line options override the config file options already, so no changes were needed for that.

Unrelated, but there was a typo in the `bashate` command in the CONTRIBUTING guide as well. 

/closes #497

### DIFF
--- a/.github/workflows/compatibility.yml
+++ b/.github/workflows/compatibility.yml
@@ -30,7 +30,7 @@ jobs:
       distrobox_changed: ${{ steps.check_file_changed.outputs.distrobox_changed }}
     steps:
 
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         with:
           # Checkout as many commits as needed for the diff
           fetch-depth: 2
@@ -55,7 +55,7 @@ jobs:
       github.event.inputs.run_always == 'True'
     steps:
 
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
       # Fetch from compatibility table all the distros supported
       - id: set-matrix
@@ -82,7 +82,7 @@ jobs:
 
     steps:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
       # Ensure distrobox create works:
       - name: Distrobox create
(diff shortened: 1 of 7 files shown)


89luca89:
Hi @mtalexan 

seems all ok for me, did a couple of changes so we can handle the variable as a normal one, but seems all ok
Will wait for the CI to finish and will merge if all ok :+1: 

Thanks a lot!

mtalexan:
Seems reasonable to me, supporting blank argument for the `pre-init-hook` in `distrobox-init` so that it can be blindly passed as part of `distrobox-create`s call to it.  

I don't follow what the GitHub workflow changes are, it looks like there's some outside generation step you ran and then reference as the new config?

89luca89:
> Seems reasonable to me, supporting blank argument for the `pre-init-hook` in `distrobox-init` so that it can be blindly passed as part of `distrobox-create`s call to it.

Yep :+1: 

> I don't follow what the GitHub workflow changes are, it looks like there's some outside generation step you ran and then reference as the new config?

No it's just GH has deprecated some stuff and the CI was not going ahead so I just fixed this now :shrug:

89luca89:
Merged, thanks a lot!

PR #543, opened 2022-11-29
Automatically create docker group matching docker.sock group GID when present

### LINKED ISSUE #494: [Error] docker.sock permissions not mapped into distrobox
Please, before opening a bug:

- [x] make sure you've read the documentation.
- [x] Ensure there isn't already an open issue about this.
- [x] Ensure there isn't already a closed/resolved issue about this.

**Describe the bug**
According to the documentation on how to run docker from within a distrobox, the `docker.socket` is mapped from the host into the distrobox container if distrobox is using docker, and all you need to do is install docker inside the distrobox and add yourself to the `docker` group.  
This appears to be incorrect, the `docker.socket` is mapped into the container but the `docker` group is not.  The `docker.socket` is therefore owned by an unknown GID as far as the distrobox can tell (the GID that matches the host's `docker` group).

Running thru these steps is how I discovered it:
```
host$ distrobox create -i ubuntu:20.04 -n ubuntu
host$ distrobox enter ubuntu
# install docker per https://docs.docker.com/engine/install/ubuntu/
ubuntu$ sudo apt install ca-certificates curl gnupg lsb-release
ubuntu$ sudo mkdir -p /etc/apt/keyrings
ubuntu$ curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo gpg --dearmor -o /etc/apt/keyrings/docker.gpg
ubuntu$ echo "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.gpg] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable" | sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
ubuntu$ sudo apt-get update
ubuntu$ sudo apt-get install docker-ce docker-ce-cli containerd.io docker-compose-plugin
# add to docker group
ubuntu$ usermod -a -G docker $(id -n)
ubuntu$ exit
# make sure the group ownership took effect
host$ distrobox stop ubuntu
host$ distrobox enter ubuntu
ubuntu$ docker login MYURL
```

The error is always:
```
Authenticating with existing credentials...
Login did not succeed, error: Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Post "http://%2Fvar%2Frun%2Fdocker.sock/v1.24/auth": dial unix /var/run/docker.sock: connect: permission denied
```
With some examination:
```
ubuntu$ groups
MYGRP docker

ubuntu$ ls -la /var/run/docker.sock
lrwxrwxrwx 1 root root 25 Nov  2 12:03 /var/run/docker.sock -> /run/host/run/docker.sock

ubuntu$ ls -la /run/host/run/docker.sock
srw-rw---- 1 root 997 0 Oct 31 08:20 /run/host/run/docker.sock

ubuntu$ getent group docker
docker:x:998:MYUSER
```
```
host$ ls -la /run/docker.socket
srw-rw---- 1 root docker 0 Oct 31 08:20 /run/docker.sock

host$ getent group docker
docker:x:997:MYUSER
```

**To Reproduce**

```
host$ distrobox create -i ubuntu:20.04 -n ubuntu
host$ distrobox enter ubuntu
ubuntu$ getent group docker
ubuntu$ ls -la /run/host/run/docker.sock
```
The `docker` group should be present in the container, and the `docker.sock` should be owned by it.  But neither is true.

**Expected behavior**
The group owner of the docker.socket on the host should be mapped into the distrobox (probably as group name `docker`).  
This allows the instructions to function as described, and adding the user within the container to the `docker` group would grant them the permissions to access the GID of the group-owner of the `docker.socket`.

**Logs**
Attached is the output of the `docker-create -i ubuntu:20.04 -n ubuntu` command 
[distrobox-create.log](https://github.com/89luca89/distrobox/files/9925129/distrobox-create.log)  
And the censored output of the `docker logs -f ubuntu` while running `distrobox enter ubuntu` 
[distrobox-enter.log](https://github.com/89luca89/distrobox/files/9925135/distrobox-enter.log)

 - Are you using podman or docker?
     docker
 - Which version or podman or docker?
     `Docker version 20.10.21, build baeda1f`
 - Which version of distrobox?
     `distrobox: 1.4.1`

**Additional context**
N/A

mtalexan:
To ensure optional manual installation of docker in the distrobox can use the docker socket mapped from the host, it's a necessity that the docker.sock group GID be associated with the container group name 'docker'. Detect if the socket is present after host socket mapping in `distrobox-init`, and create the 'docker' named group to own it. 

Unrelated: CONTRIBUTING.md for pre-checks could do with some minor clarifications so they were thrown in.

/closes #494

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -85,13 +85,15 @@ done
 - use `shellcheck` to check for posix compliance and bashisms using:
   - install from: [HERE](https://github.com/koalaman/shellcheck)
     following [this](https://github.com/koalaman/shellcheck#installing)
-  - `shellcheck -s sh -a -o all -Sstyle -Calways -x -e SC2310,SC2311,SC2312`
+  - `shellcheck -s sh -a -o all -Sstyle -Calways -x -e SC2310,SC2311,SC2312 {changed_files}`
 - use `shfmt` to style the code using:
   - install from [HERE](https://github.com/mvdan/sh) using `go install mvdan.cc/sh/v3/cmd/shfmt@latest`
-  - `shfmt -d -s -ci -sr -kp`
+    - `shfmt -d -s -ci -sr -kp {changed_files}`
+  - or via Docker/Podman without installation:
+    - `docker run --rm -u "$(id -u):$(id -g)" -v "$PWD:/mnt" -w /mnt mvdan/shfmt:latest -d -s -ci -sr -kp {changed_files}`
 - use `bashate` to check the code:
   - install using `pip3 install bashate`
-  - `bashate -i E002,E003,E010,E011 --max-line-length 12`
+  - `bashate -i E002,E003,E010,E011 --max-line-length 120 {changed_files}`
 - use `markdownlint`
   - install using `npm -i -g markdownlint-cli`
   - run `markdownlint $(find . -name '*.md' | grep -vF './.git')`
--- a/distrobox-init
+++ b/distrobox-init
@@ -775,6 +775,8 @@ host_sockets="$(find /run/host/run -name 'user' \
 	-prune -o -name 'system_bus_socket' \
 	-prune -o -type s -print \
 	2> /dev/null || :)"
+
+docker_socket=""
 # we're excluding system dbus socket and nscd socket here. Including them will
 # create many problems with package managers thinking they have access to
 # system dbus or user auth cache misused.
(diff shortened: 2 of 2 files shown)


89luca89:
Hi @mtalexan 

why instead of installing `docker` inside distrobox to connect to host's docker daemon
you don't just do
`sudo ln -s /usr/bin/distrobox-host-exec /usr/bin/docker` inside the container?

This will give you a `docker` command in PATH that calls directly host's docker trasparently

mtalexan:
That's a great question. I don't understand why the main README lists this as one of the few documented uses (https://github.com/89luca89/distrobox/blob/main/docs/useful_tips.md#using-docker-inside-a-distrobox).

Docker in particular gains nothing from running it in Distrobox since it's actually executed in the host via the docker socket.  I was just fixing a big I don't when I did a quick test on it, even though I realized there was no benefit versus running it directly on the host.

89luca89:
@mtalexan 

 that's right, the docs are behind
 Will update the docs with this information you're right :+1:

89luca89:
I've updated the docs to reflect this, resolved the issue and will close this PR
Thanks anyway :smile:

PR #558, opened 2022-12-17
make podman-host executable

mirkobrombin:


89luca89:
:heart:

mingyc:
I tried to follow [the second approach](https://github.com/89luca89/distrobox/blob/main/docs/posts/integrate_vscode_distrobox.md#the-not-so-easy-one) and found that it doesn't work. There are two errors

1. When trying to `Attach to Running Container...`, it gives  
```
+ '[' version == exec ']'
+ exec flatpak-spawn --host podman version --format '{{.Server.APIVersion}}'
Portal call failed: Failed to start command: Failed to execute child process “.local/bin/podman” (No such file or directory)
```
I guess it is because that the environment of the spawned host/external application is not inherited from the host shell that launched the VSCode Flatpak. I don't know how to fix this.

2. Even with hardcoded path to podman within `podman-remote`, it still gives the following error

```
+ '[' version == exec ']'
+ exec flatpak-spawn --host /path/to/home/.local/podman/bin version --format '{{.Server.APIVersion}}'
Portal call failed: Failed to start command: Failed to execute child process “/path/to/home/.local/podman/bin” (Permission denied)
```

PR #569, opened 2022-12-25
Add yay and paru support to distrobox-init and distrobox-upgrade

cyrv6737:
This PR adds the paru and yay AUR helpers to distrobox-init to be preferred over pacman when they exist on the container.

 I've added a `--aur` argument to `distrobox-upgrade` which will check for either `yay` or `paru`, and if detected will run the entrypoint without sudo.  `distrobox-upgrade --all --aur` will also handle containers using standard package managers along with arch containers with yay or paru installed. These checks are needed as makepkg disallows itself to run as root.

Please let me know if there are any issues and I'll try my best to resolve ASAP.

### DIFF
--- a/distrobox-init
+++ b/distrobox-init
@@ -452,6 +452,96 @@ if [ "${upgrade}" -ne 0 ] || ! command -v find || ! command -v mount || ! comman
 			mesa-vulkan-drivers \
 			vulkan || :
 
+	elif command -v yay; then
+		# If we need to upgrade, do it and exit, no further action required.
+		if [ "${upgrade}" -ne 0 ]; then
+			yay -Syyu --noconfirm
+			exit
+		fi
+		# In archlinux  official images, pacman is configured to ignore locale and docs
+		# This however, results in a rather poor out-of-the-box experience
+		# So, let's enable them.
+		sed -i "s|NoExtract.*||g" /etc/pacman.conf
+		sed -i "s|NoProgressBar.*||g" /etc/pacman.conf
+
+		# Check if shell_pkg is available in distro's repo. If not we
+		# fall back to bash, and we set the SHELL variable to bash so
+		# that it is set up correctly for the user.
+		yay --noconfirm -Syyu
+		if ! yay -Sy --noconfirm "${shell_pkg}"; then
+			shell_pkg="bash"
+		fi
+		yay -Sy --noconfirm \
+			"${shell_pkg}" \
+			bc \
+			curl \
+			diffutils \
+			findutils \
+			gnupg \
+			less \
+			lsof \
+			ncurses \
+			pinentry \
+			procps-ng \
+			shadow \
+			sudo \
+			time \
+			util-linux \
+			wget \
+			vte-common
+
+		# These are graphics drivers for 3d applications
+		yay -Sy --noconfirm \
+			mesa \
+			opengl-driver \
+			vulkan-intel \
+			vulkan-radeon
+
+	elif command -v paru; then
+		# If we need to upgrade, do it and exit, no further action required.
+		if [ "${upgrade}" -ne 0 ]; then
+			paru -Syyu --noconfirm
+			exit
+		fi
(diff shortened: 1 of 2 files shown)


cyrv6737:
After submitting this PR it has come to my attention that `distrobox-upgrade` automatically runs the entrypoint as sudo which conflicts with makepkg disallowing itself to be run as sudo. Looking into a way to sanely implement running as user

cyrv6737:
Previous comment should be resolved. Apologies for not preparing that before PR.

89luca89:
Hi @cyrv6737 

I'm really glad for your interest in distrobox
But I'm sorry to say that this feature is something I'm not really keen or interested to add to distrobox

Simply put there are tons and tons of aur helpers, and one cannot possibly be able to maintain and ensure everything is always working correctly, and it's a maintenance burden that I'm not interested in having

So I prefer distrobox to stick to the official package managers for init/upgrade

cyrv6737:
Thanks for your quick response @89luca89 . I understand your perspective, however I would urge you to reconsider for the following reasons:

- This is a desired feature for distrobox to be able to optionally upgrade containers running yay/paru. It's not just myself wanting this functionality. It's pretty annoying to have to worry about AUR packages when those of us running Fedora Silverblue with automatic updates on everything don't have to worry about updates in any other instance.

- In my opinion, it is a reasonable assumption that yay and paru would cover the vast majority of arch users that use an AUR helper. In the same vein, I don't think it's unreasonable to not support other AUR helpers. Additionally, both of these AUR helpers have a great history with stability are well maintained. **Even just one of these would be better than none at all**, if that's a compromise you'd entertain.

- The last commit I made in this PR removes yay/paru from being called in container initialization. Granted, yay/paru would only be called during initialization if distrobox is pulling from a custom image, but in that case pacman would still handle initialization.

It would seem to me that very little additional maintenance burden would fall on you because of this PR, especially since the user would have to manually set up an AUR helper and manually invoke distrobox-upgrade with `--aur` to handle it. However, that being said, I am not the maintainer and you would have a better understanding of this than I would.

I appreciate your consideration for this feature regardless of your decision.

89luca89:
@cyrv6737 will reply by steps

> This is a desired feature for distrobox to be able to optionally upgrade containers running yay/paru. It's not just myself wanting this functionality. It's pretty annoying to have to worry about AUR packages when those of us running Fedora Silverblue with automatic updates on everything don't have to worry about updates in any other instance

For this cases, there are tools like topgrade (or even just a simple script in the container) to do that 

> In my opinion, it is a reasonable assumption that yay and paru would cover the vast majority of arch users that use an AUR helper. In the same vein, I don't think it's unreasonable to not support other AUR helpers. Additionally, both of these AUR helpers have a great history with stability are well maintained. Even just one of these would be better than none at all, if that's a compromise you'd entertain.

I argue that it's either better offering something properly supported, testable and maintainable in the future, than an additional feature, likely to cause breakage/incompatibilities in the future due to it's unofficial nature (aur is still not part of main arch)

> It would seem to me that very little additional maintenance burden

As stated before, it's not (mainly) about the added code, but how this fits in the future maintainability of the software, how likely is to create problems and breakage

> especially since the user would have to manually set up an AUR helper and manually invoke distrobox-upgrade with --aur to handle it.

Point is, distrobox is *not* about arch at all, not about aur, but is a generic tool
At this point one could argue that some users would want to add Packman (opensuse aur), Copr, pacstall, deb-get etc etc 
It will quickly go out of control

Gecked-Deck:
I'd love to see this implemented, it'd greatly reduce the hassle of updating arch packages!

cyrv6737:
@89luca89 
All very good points, thank you for the thought-out reply. Above all I respect sticking to your design philosophy of this project. I'll go ahead and close this PR. Thanks for your time again!

89luca89:
> @89luca89 All very good points, thank you for the thought-out reply. Above all I respect sticking to your design philosophy of this project. I'll go ahead and close this PR. Thanks for your time again!

Thanks to you for understanding :)

PR #577, opened 2022-12-31
clearer root error message

misobarisic:
Signed-off-by: Mišo Barišić <me@misobarisic.com>


89luca89:
Hi @misobarisic 
I'll keep this pending for now, there is #582 that would change things a bit so this PR could be affected

89luca89:
@misobarisic sorry, I'm closing this one, as I've merged @luc14n0 's PR which fixes the problem of running as real root

PR #582, opened 2023-01-09
Allow root user to run Distrobox

### LINKED ISSUE #575: DIstrobox fails to run as root user with or without '--root' flag
Hello all, 
I maintain a single-user mode OS that uses an ubuntu base (work purposes). There are no other accounts besides root and various service accounts. 
I was looking at integrating Distrobox into the OS and during my experimentation I haven't been able to get the latest tip of main to work. 

I do apologize for opening what appears to be the same issue as #572 but I hope this post serves as guidance for users on how to properly interact with a community. Thank you for your time. 

```
root@rtbuntu ➜ ~ distrobox-create --name test
Running distrobox-create as sudo is not supported.
 try instead running:
        distrobox-create --root --name test
root@rtbuntu ➜ ~ distrobox-create --root --name test
Running distrobox-create as sudo is not supported.
 try instead running:
        distrobox-create --root --root --name test
root@rtbuntu ➜ ~ distrobox-create --root --root --name test
Running distrobox-create as sudo is not supported.
 try instead running:
        distrobox-create --root --root --root --name test
root@rtbuntu ➜ ~ distrobox -v
Error: invalid command
distrobox version: 1.4.2.1
```

luc14n0:
Up until now, we weren't able to distinguish between normal users (running the scripts via `sudo`) and **root** user (running the script in a login shell). This intruduces a check for the **SUDO_USER** variable (created in unmodified environments, by `sudo`), among other mechanisms, to allow the **root** user to manage _rootful_ containers.

This fixes #575.

### DIFF
--- a/distrobox-create
+++ b/distrobox-create
@@ -33,11 +33,15 @@
 #	DBX_NON_INTERACTIVE
 #	DBX_SUDO_PROGRAM
 
-# Dont' run this command as sudo.
-if [ "$(id -u)" -eq 0 ]; then
-	printf >&2 "Running %s as sudo is not supported.\n" "$(basename "${0}")"
-	printf >&2 " try instead running:\n"
-	printf >&2 "	%s --root %s\n" "$(basename "${0}")" "$*"
+# Despite of running this script via SUDO/DOAS being not supported (the
+# script itself will call the appropriate tool when necessary), we still want
+# to allow people to run it as root, logged in in a shell, and create rootful
+# containers.
+#
+# SUDO_USER is a variable set by SUDO and can be used to check whether the script was called by it. Same thing for DOAS_USER, set by DOAS.
+if [ -n "${SUDO_USER}" ] || [ -n "${DOAS_USER}" ]; then
+	printf >&2 "Running %s via SUDO/DOAS is not supported. Instead, please try running:\n" "$(basename "${0}")"
+	printf >&2 "  %s --root %s\n" "$(basename "${0}")" "$*"
 	exit 1
 fi
 
@@ -59,7 +63,6 @@ container_user_gid="$(id -rg)"
 container_user_home="${HOME:-"/"}"
 container_user_name="${USER}"
 container_user_uid="$(id -ru)"
-distrobox_sudo_program="sudo"
 dryrun=0
 init=0
 non_interactive=0
@@ -75,7 +78,9 @@ distrobox_hostexec_path="$(cd "$(dirname "${0}")" && pwd)/distrobox-host-exec"
 [ ! -e "${distrobox_entrypoint_path}" ] && distrobox_entrypoint_path="$(command -v distrobox-init)"
 [ ! -e "${distrobox_export_path}" ] && distrobox_export_path="$(command -v distrobox-export)"
 [ ! -e "${distrobox_hostexec_path}" ] && distrobox_hostexec_path="$(command -v distrobox-hostexec)"
-rootful=0
+# If the user runs this script as root in a login shell, set rootful=1.
+# There's no need for them to pass the --root flag option in such cases.
+[ "${container_user_uid}" -eq 0 ] && rootful=1 || rootful=0
 verbose=0
 version="1.4.2.1"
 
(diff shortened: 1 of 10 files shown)


distrobox-create
```diff
@@ -33,8 +33,10 @@
 #	DBX_NON_INTERACTIVE
 #	DBX_SUDO_PROGRAM
 
-# Dont' run this command as sudo.
-if [ "$(id -u)" -eq 0 ]; then
+# Although running this script via sudo is not supported, we still want
+# to allow people to run it as root, using a login shell, and get rootful
+# containers, of course.
+if [ -n "${SUDO_USER}" ]; then
```
89luca89, line 39: Is this valid also for other "sudo" implementations like `doas`?

distrobox-create
```diff
@@ -59,7 +61,15 @@ container_user_gid="$(id -rg)"
...
-distrobox_sudo_program="sudo"
+# If we're running distrobox-create as root - as in, logged in the shell as
+# root user, and not via sudo -, we don't want to set distrobox_sudo_program
+# (it's meaningless for our usecase).
+if [ "${container_user_uid}" -ne 0 ]; then
+	# If the DBX_SUDO_PROGRAM/distrobox_sudo_program variable was set by the
+	# user, use its value instead of "sudo". But only if not runnig the script
+	# as root (UID 0).
+	distrobox_sudo_program=${DBX_SUDO_PROGRAM:-${distrobox_sudo_program:-"sudo"}}
+fi
```
89luca89, line 72: If we move this block after the "source" command, it will continue to work as before, in the sense that a non-root user can set the sudo program in the config files also

distrobox-create
```diff
@@ -664,7 +675,7 @@ fi
 if ${container_manager} inspect --type container "${container_name}" > /dev/null 2>&1; then
 	printf "Distrobox named '%s' already exists.\n" "${container_name}"
 	printf "To enter, run:\n\n"
-	if [ "${rootful}" -ne 0 ]; then
+	if [ "${container_user_uid}" -ne 0 ] || [ "${rootful}" -eq 0 ]; then
```
89luca89, line 678: Put this way, seems like if `rootful=0` or `container_user_uid!=0` will always say to use `--root`

So If I use `distrobox create` as a normal user (no root, nor --root) I will receive this message, which is not correct

89luca89: You do the right check at 734

89luca89 (changes requested):

89luca89:
Hi @luc14n0 

I've left some comments, thanks a lot for working on this! :smile:

distrobox-create
luc14n0: I can't tell first hand but if I would guess I'd say no. But I'm willing to take a look at a couple of DOAS Linux ports/implementation and see if there's a similar mechanism in use, otherwise I think I know how we could go about being DOAS-aware.

distrobox-create
luc14n0: This bit is a tricky one to get right. I'm gonna revise this carefully, though I remember this working as expected in my testings.

distrobox-create
89luca89: Seems like `doas` does something similar to sudo, injecting a `DOAS_USER` in the env

luc14n0: Yes, it does. Which's going to make our life easier.

distrobox-create
luc14n0: You were right. I forgot to use the solution I've found, used at line 734, here.

luc14n0: The documentation need changes as well, I almost forgot about them.

luc14n0:
Do we need anything else to be updated (from `docs/`/`man/`)?

89luca89:
Thanks a lot @luc14n0 seems all is working as intended now! :smile: 
Great stuff!

senorsmile:
Any idea when this will be in a release?  I'd like to test on NixOS, but cannot just `git clone` and `./install` since /usr/lib (and really almost all dirs) are immutable.

senorsmile:
Never mind.  I was able to git clone and add to path and confirmed it's running as single user root on NixOS.  Thanks!

PR #585, opened 2023-01-15
Fix typo and incorrect key for generated desktop entry `Categories`

ian-h-chamberlain:
The proper key is "Categories", per https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html

The trailing quote seems to just be a copy-paste error from moving the lines around in 5d77405e5ffc89479cdc8e9b2d482e018c469885

I hope it's okay, I didn't open an issue for this as the fix seemed trivial I wasn't sure it was worth filing separately.


89luca89:
Thanks a lot @ian-h-chamberlain !

PR #605, opened 2023-01-31
Install reflector(1) in arch containers

axtloss:
This installs reflector(1) in arch containers (if found in the repos) and runs a sort on the 40 latest synchronized http/https mirrors to make sure that the rest of the container setup is way faster

### DIFF
--- a/distrobox-init
+++ b/distrobox-init
@@ -471,6 +471,16 @@ if [ "${upgrade}" -ne 0 ] || ! command -v find || ! command -v mount || ! comman
 		if ! pacman -Sy --noconfirm "${shell_pkg}"; then
 			shell_pkg="bash"
 		fi
+
+		# Install reflector(1) and make it sort the 40 most recently
+        # synchronized http/https mirrors by speed and overwrite the mirrorlist
+        # This greatly increases the download speed of installed packages
+        if ! pacman -Sy --noconfirm reflector; then
+			pacman -Sy --noconfirm reflector
+	        reflector --latest 40 --protocol http,https --sort rate --save /etc/pacman.d/mirrorlist
+		fi
+
+
 		pacman -Sy --noconfirm \
 			"${shell_pkg}" \
 			bc \


89luca89:
Hi @axtloss 

testing this, but instead of speeding up the setup, it increased greatly the time now, because:

image

It literally tries every mirror and waits for timeouts

axtloss:
that's, weird, it's I dictating that it scans every mirror, even though it's supposed to only scare the latest 40

89luca89:
> that's, weird, it's I dictating that it scans every mirror, even though it's supposed to only scare the latest 40

Not familiar with this tool
But even if it scans the latest 40, wouldn't this be slow to check all of them?

If it's any similar to `fastmirror` of dnf/yum, it's not a net gain in a one-shot action like the first init

axtloss:
When I test it it actually is significantly faster than without it, without reflector I get ~500kb/s download, with reflector it's ~5mb/s. It's also worth noting that these changes stay after the distrobox is properly initialized, making any user installation of extra packages by the user faster too.

I guess I could change the amount of mirrors it compares, which could decrease the startup time

89luca89:
Probably this is something more suitable to an init-hook
One could set init-hooks globally (for example if you use exclusively arch distroboxes) in distrobox configs:

file: ~/.config/distrobox/distrobox.conf or ~/.distroboxrc
```toml
...
container_pre_init_hook="pacman -Syu --noconfirm reflector && reflector [--options...]"
...
```

axtloss:
Yeah I guess that would work, is it possible to create pre init hooks for specific container types?

89luca89:
> Yeah I guess that would work, is it possible to create pre init hooks for specific container types?

You can specify scripts, and then in that script you can manage stuff like:

```toml
...
container_pre_init_hook="/path/to/script"
...
```

then in the script

```
#!/bin/sh

if command -v zypper; then
...

```

axtloss:
oh alright, I guess this PR isn't needed then

PR #609, opened 2023-02-03
distrobox-create: allow the ability to "unshare" netns and ipc

### LINKED ISSUE #608: [Discussion] Is is currently possible to not share a particular namespace
Is your feature request related to a problem? Please describe.

Currently, I do not necessarily see a way to run a distrobox container without sharing a specific namespace. I'm wondering if it possible. 

I'm using distrobox to develop eBPF programs which touch the network stack. It's very helpful to develop these applications in a "network sandbox". Meaning, I can create devs, remove them, mess with routing and routing policy, and then just "blow it away" when I'm done. 

However, I can't seem to find a way to have distrobox create my container without sharing the net namespace. I imagine the desire to do this extends further then just network namespaces. 

Describe the solution you'd like

Curious if there is currently a way to not share the network namespace when creating a distrobox container (based on fedora toolbox).

ldelossa:
this commit adds two new flags to "distrobox-create":
--unshare-netns
--unshare-ipc

both can be used to no longer share the respective namespace with the host.

Closes #608 

Signed-off-by: Louis DeLosSantos <louis.delos@gmail.com>

### DIFF
--- a/distrobox-create
+++ b/distrobox-create
@@ -64,8 +64,10 @@ container_user_gid="$(id -rg)"
 container_user_home="${HOME:-"/"}"
 container_user_name="${USER}"
 container_user_uid="$(id -ru)"
-nvidia=0
 dryrun=0
+nvidia=0
+unshare_ipc=0
+unshare_netns=0
 init=0
 non_interactive=0
 # Use cd + dirname + pwd so that we do not have relative paths in mount points
@@ -494,12 +496,12 @@ generate_command() {
 		--security-opt label=disable
 		--user root:root"
 
-	if  ! [ -v unshare_ipc ]; then
+	if [ "${unshare_ipc}" -ne 0 ]; then
 		result_command="${result_command}
 			--ipc host"
 	fi
 
-	if  ! [ -v unshare_netns ]; then
+	if [ "${unshare_netns}" -ne 0 ]; then
 		result_command="${result_command}
 			--network host"
 	fi
--- a/distrobox-create
+++ b/distrobox-create
@@ -171,6 +171,8 @@ Options:
 	--init/-I:		use init system (like systemd) inside the container.
 				this will make host's processes not visible from within the container.
 	--nvidia:		try to integrate host's nVidia drivers in the guest
+	--unshare-netns:        do not share the net namespace with host
+	--unshare-ipc:          do not share ipc namemspace with host
 	--compatibility/-C:	show list of compatible images
 	--help/-h:		show this message
 	--no-entry:		do not generate a container entry in the application list
@@ -249,6 +251,14 @@ while :; do
 		-I | --init)
 			shift
 			init=1
+			;;
+		--unshare-netns)
+			shift
+			unshare_netns=1
+			;;
+		--unshare-ipc)
+			shift
+			unshare_ipc=1
 			;;
 		-C | --compatibility)
 			show_compatibility
@@ -479,12 +489,20 @@ generate_command() {
 	# use the host's namespace for ipc, network, pid, ulimit
 	result_command="${result_command}
 		--hostname \"${container_name}.$(uname -n)\"
-		--ipc host
 		--name \"${container_name}\"
-		--network host
 		--privileged
 		--security-opt label=disable
 		--user root:root"
+
+	if  ! [ -v unshare_ipc ]; then
+		result_command="${result_command}
+			--ipc host"
+	fi
+
+	if  ! [ -v unshare_netns ]; then
+		result_command="${result_command}
+			--network host"
+	fi
 
 	if [ "${init}" -eq 0 ]; then
 		result_command="${result_command}


89luca89:
Hi @ldelossa thanks for the PR!

Maybe we can try to work together with the PR for network isolation? #478 
This way we can deduplicate some work

ldelossa:
Hey there. 

Im opened to this, however it did feel that for the use case of distrobox, plumbing thru the various network types seems maybe like overkill? 

I did brush over that pr and it seems like it got a bit complicated with cleaning up. Had assumed since it didnt move forward, it maybe wasnt the best idea. 

What you thinking?

89luca89:
In the optics of #28 probably yes that one is overkill
@Cloudwalk9  are you still working on that network PR?

Seems like just unsharing is enough for basic decoupling instead of creating private networks, this is more or less what bwrap does in flatpaks I guess

ldelossa:
Thats my personal opinion. For the "user centric" use case, i imagine simply saying "share the ns, or not. 

If im getting into the realm of creating net bridges and bridging containers together, then i think we have left the general use case of using distrobox for app/dev containers. 

Happy to go either way tho. Both options accomplish my goal.

89luca89:
> Thats my personal opinion. For the "user centric" use case, i imagine simply saying "share the ns, or not" is enough.

Yes I agree on this

Probably it would be better to implement something like this (maybe also a --no-network after) and leave the advanced bridge/networking to pure podman

Will wait a bit for a feedback from @Cloudwalk9

Cloudwalk9:
> > Thats my personal opinion. For the "user centric" use case, i imagine simply saying "share the ns, or not" is enough.
> 
> Yes I agree on this
> 
> Probably it would be better to implement something like this (maybe also a --no-network after) and leave the advanced bridge/networking to pure podman
> 
> Will wait a bit for a feedback from @Cloudwalk9

I haven't been able to work much more on it because of mental health and I'm probably gonna be unable to continue for a while, unfortunately. The only thing I was stuck on was tab autocomplete I think.

But for background, I created that to mimic Android's work profile feature that allows normal Internet connection on your main profile and a VPN on your work profile. This patch would only unshare to a single namespace for all containers created with the flag. My use case required multiple containers with separate VPN connections. I'd say both patches complement each other.

89luca89:
> I haven't been able to work much more on it because of mental health and I'm probably gonna be unable to continue for a while, unfortunately

I'm very sorry to hear! I wish you a speedy recovery!

> But for background, I created that to mimic Android's work profile feature that allows normal Internet connection on your main profile and a VPN on your work profile. This patch would only unshare to a single namespace for all containers created with the flag without touching docker/podman network. My use case required multiple containers with separate VPN connections. I'd say both patches complement each other.

Mh I see, probably we can implement the multi-network stuff in a second step, first implementing some light unsharing for #28 and if the feature is used/of-interest we can go deeper
I'll keep your PR anyway for future reference to it!

ldelossa:
@89luca89 do you know if you'd like to merge this one yet?

89luca89:
> @89luca89 do you know if you'd like to merge this one yet?

Yes sorry, I was quite busy :/

I was just thinking, in the optics of having it easy to use for a regular user, not familiar with containers in general,
would be useful to have `--unshare-network` instead of `netns`? Maybe the concept of network namespaces is not as widespread as simply saying "unshare the network please"

ldelossa:
> > @89luca89 do you know if you'd like to merge this one yet?
> 
> Yes sorry, I was quite busy :/
> 
> I was just thinking, in the optics of having it easy to use for a regular user, not familiar with containers in general, would be useful to have `--unshare-network` instead of `netns`? Maybe the concept of network namespaces is not as widespread as simply saying "unshare the network please"

Sounds good to me. What about "unshare-ipc" ? Are you okay with this naming?

89luca89:
> > > @89luca89 do you know if you'd like to merge this one yet?
> > 
> > 
> > Yes sorry, I was quite busy :/
> > I was just thinking, in the optics of having it easy to use for a regular user, not familiar with containers in general, would be useful to have `--unshare-network` instead of `netns`? Maybe the concept of network namespaces is not as widespread as simply saying "unshare the network please"
> 
> Sounds good to me. What about "unshare-ipc" ? Are you okay with this naming?

Yea I was thinking "unshare network" is something even more regular users would search for
Ipc is more advanced, so it can stay that way

Cloudwalk9:
Riding a burst of energy:

I propose `--unshare-netns` (or `--unshare-network`) creates a new network namespace with `podman network` and a `--network` option could attach to an existing network namespace instead of also creating a new one, which `--unshare-network` could do instead. Perhaps an optional argument for `--unshare-network` to specify a name, with the default being "dbx-<container name>" or something else if left blank.

89luca89:
@ldelossa there are a couple of comments to check for some bashisms :)

ldelossa:
@89luca89 - Thanks i'm trying to deal with some of the linting errors with running "shfmt" locally. However, it creates quite a large diff. If there a way for me to run the same linters as your CI does so I can make it look the same?

89luca89:
> @89luca89 - Thanks i'm trying to deal with some of the linting errors with running "shfmt" locally. However, it creates quite a large diff. If there a way for me to run the same linters as your CI does so I can make it look the same?

Sure, if you check the CONTRIBUTING.md all the commands for shellcheck and shfmt are there :+1:

ldelossa:
Aha, thanks, totally missed that.

89luca89:
Thanks a lot @ldelossa for the contribution
I'm merging this

PR #616, opened 2023-02-07
fix oversight in #422

misobarisic:
I recently discovered how `$DISTROBOX_HOST_HOME` was not being set properly due to a fault in #422


89luca89:
Thanks!

PR #620, opened 2023-02-11
Can't start container using podman

s3rgeym:
podman logs outputs to stderr instead docker to stdout => forever loop because .pipe file is empty => container not starting


89luca89:
Hi @s3rgeym 
this shouldn't be the case

I have podman in several systems and it's always using stderr

PR #635, opened 2023-02-25
init: Reuse locked flags when mounting read-only

### LINKED ISSUE #545: [Error] `distrobox enter` fails with failed bind mount for /var/lib/flatpak
**Describe the bug**

When a container is created, and then entered for the first time, it fails and bails out with the following output:
```
Container ubuntu is not running.
Starting container ubuntu

 podman logs -f ubuntu

 Starting container...                  	 [ OK ]
 Installing basic packages...           	 [ OK ]
 Setting up read-only mounts...         	
 Warning: failed to bind mount /run/host/var/lib/flatpak to /var/lib/flatpak Error: An error occurred

An error occurred

An error occurred
```

 Subsequent attempts to enter the container also fail.

I tried both the default fedora image and with ubuntu:22.04 and had the same issue. My host system is NixOS.

**To Reproduce**

1. Run `distrobox create -i ubuntu:22.04 ubuntu
2. Run `distrobox enter ubuntu`

There is then the warning and exit.

**Expected behavior**

I expect to enter the container.

**Logs**

[distrobox create -i ubuntu:22.04 ubuntu --verbose](https://github.com/89luca89/distrobox/files/10145499/log_create.txt)

[distrobox enter ubuntu --verbose](https://github.com/89luca89/distrobox/files/10145500/log_enter.txt)

[podman logs ubuntu](https://github.com/89luca89/distrobox/files/10145504/log_podman.txt)

 - Using podman
 - podman version 4.3.1
 - distrobox version 1.4.1

**Additional context**

morganhjk:
Fix #545.

In a user namespace (e.g. when using rootless Podman), certain mount flags are considered "locked" and cannot be cleared.

The check for these flags is triggered (in this case) when bind-mounting directories from the host as read-only.

These flags must be explicitly reused when a host directory, or its nearest mountpoint, uses them.

For example, if /var/lib/flatpak on the host is on a rootfs mounted using `nodev` (one of the locked flags), it must be reused.


morganhjk:
I did some testing on the images that failed in CI and found that the `findmnt` version in them did not check parent directories until it encountered a mountpoint. Added a workaround.

89luca89:
> I did some testing on the images that failed in CI and found that the `findmnt` version in them did not check parent directories until it encountered a mountpoint. Added a workaround.

Cool running CI again let's see :)

89luca89:
Seems all green! Merging :)
Thanks a lot for the PR!

PR #647, opened 2023-03-04
host-exec: bump host-spawn to 1.4.1 and do proper version comparison

pavelxdd:
Download `host-spawn` only if it is missing or when the installed version is actually older.

Prior to this change `distrobox-host-exec` will download `host-spawn` even when the locally installed version is greater than expected.

For example `distrobox-host-exec` expected `host-spawn` version `1.2.1` and I have version `1.4.1` installed locally. Every time I call `distrobox-host-exec` it won't work unless I download the version `1.2.1`.

This shouldn't be happenning, considering that `distrobox-host-exec` prints that my version is older, when it is actually newer.

### DIFF
--- a/distrobox-host-exec
+++ b/distrobox-host-exec
@@ -124,8 +124,8 @@ fi
 
 # Setup host-spawn as a way to execute commands back on the host
 if  ! command -v host-spawn > /dev/null ||
-	[ "$(printf "%s\n%s\n" "${host_spawn_version}" "$(host-spawn --version)" \
-		| sort -V | head -n 1)" != "${host_spawn_version}" ]; then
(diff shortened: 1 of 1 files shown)


89luca89:
Hi @pavelxdd thanks a lot!
Just a problem of formatting, you can run `shfmt -s -ci -sr -kp -w $filename` and should be ok

pavelxdd:
I've run `shfmt -s -ci -sr -kp -w distrobox-host-exec` inside the repo directory and no changes were done by that command.

Are you sure there's a formatting issue?

89luca89:
> I've run `shfmt -s -ci -sr -kp -w distrobox-host-exec` inside the repo directory and no changes were done by that command.
> 
> Are you sure there's a formatting issue?

I've just ran it myself and pushed that probably is why :)

Anyway now is all ok, we can merge :)

PR #651, opened 2023-03-08
distrobox-export: add login shell option

### LINKED ISSUE #639: [Suggestion] distrobox-export: run in login shell
Some binaries may expect the presence of certain environment variables to function properly (in my case: petalinux suite from Xilinx). 
As of now, it is impossible to run such binaries with distrobox, except by specifying each variable in the Containerfile, which may cumbersome. This is a bit annoying because otherwise distrobox is the perfect tool to run that type of cumbersome proprietary software which are supported on a few, potentially end-of-life platforms.

This issue mainly comes from the fact that the binaries are ultimately run with `podman exec  <container> <binary>` which does not provide a login shell as such. I suggest modifying https://github.com/89luca89/distrobox/blob/6af1458efffa5879abc82899be6f2f4d8cdf7d99/distrobox-export#L249 to insert `bash -l -c` in front of `${is_sudo} ${exported_bin} ${extra_flags} "`.

This can be either a new default, or through a new option which could be named `--login` during the export.

Alternatives either involve manually editing the exported scripts, edit the Containerfile, which are a bit cumbersome. The other option would be to find somehow a way to source files for non-interactive, non-login shells, but I am not aware of a universal way to accomplish that.

I would be willing to make a merge request if this idea is well recieved.

Scafir:
Add the option (`-l` or `--login`) to `distrobox-export` so that the exported command runs inside a login shell, giving it access to the variables sourced during login (e.g: /etc/profile).
If `--sudo` is also specified, then care has been taken so that it also runs in a login shell (via -i).

Fixes #639

### DIFF
--- a/distrobox-export
+++ b/distrobox-export
@@ -38,7 +38,9 @@ extra_flags=""
 #	DBX_HOST_HOME is set in case container is created
 #	with custom --home directory
 host_home="${DISTROBOX_HOST_HOME:-"${HOME}"}"
-is_sudo=""
+start_shell=""
+is_login=0
+is_sudo=0
 rootful=""
 verbose=0
 version="1.4.2.1"
@@ -77,6 +79,7 @@ Options:
 				Defaults to (on \$container_name)
 	--export-path/-ep:	path where to export the binary
 	--extra-flags/-ef:	extra flags to add to the command
+    --login/-l      run the exported item in a login shell
 	--sudo/-S:		specify if the exported item should be run as sudo
 	--help/-h:		show this message
 	--verbose/-v:		show more verbosity
@@ -124,8 +127,12 @@ while :; do
 				shift
 			fi
 			;;
+        -l | --login)
+            is_login=1
+            shift
+            ;;
 		-S | --sudo)
-			is_sudo="sudo"
+			is_sudo=1
 			shift
 			;;
 		-el | --export-label)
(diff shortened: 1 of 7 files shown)


89luca89:
Hi @Scafir thanks for the PR!

Just a couple of comments there, they refer to the use of "${variable}" instead of $variable

I think after fixing those, and running CI, we can merge

89luca89:
All green, we can merge :)
Thanks a lot!

distrobox-export
```diff
@@ -228,8 +235,17 @@ if [ -z "${container_name}" ]; then
 	container_name=$(uname -n | cut -d'.' -f1)
 fi
 
+#
+if [ "${is_login}" -ne 0 ] && [ "${is_sudo}" -ne 0 ]; then
+	start_shell="sudo -i"
+elif [ "${is_login}" -ne 0 ]; then
+	start_shell="sh -l -c"
```
fredizzimo, line 242: This does not work for commands taking arguments since `-c` takes a string, and not arbitrary parameters. Also shouldn't it use the user shell rather than `sh`

89luca89: Thanks @fredizzimo I've pushed a fix for it :)

Scafir: Looking a bit further, there is another issue that I don't quite understand. If you run `bash gcc`, it will abort with `cannot execute binary file`, whereas `bash -c "gcc"` successfully runs. It seems to me that having a structure like this:
```
${SHELL} -l -c "exportedCommand $@"
```
is the most general solution.

fredizzimo: In order to work around the issues I had here https://github.com/89luca89/distrobox/issues/656 I made changes to always run the login shell, no matter what. In `distrobox-enter`  

```diff
        result_command="${result_command} ${container_name}"
 
        if [ -n "${container_command}" ]; then
-               result_command="${result_command} ${container_command}"
+                quoted_container_command="\\\"${container_command}\\\""
+               result_command="${result_command} sh -c \"\\\$(getent passwd ${USER} | cut -f 7 -d :) -lc ${quoted_con
tainer_command}\""
        else
                # if no command was specified, let's execute a command that will find
                # and run the default shell for the user
```
So that's basically what's @Scafir is suggesting to do here, allthough the quoting needed special handling there. I'm not sure if it's needed here as well.

By always running the login shell I can guarantee that the rc file is always run and I can unset and set the environment variables I want. This variant has been working well for me, but I'm not sure if that's the best way to solve the issues, so I have not sent any PR .But anyway I think a similar approach could be taken here as well.

PR #655, opened 2023-03-13
Fix microOS install instructions

FarisFiroz:
PR to change 'pkcon' to 'transactional-update pkg'

Pkcon is not on used on microOS anymore it seems. It is not installed by default anymore.

Packages now are installed using 'transactional-update pkg install', the openSuse microOS documentation has also reflects this change.


89luca89:
Thanks for the PR!

Thinking of this, distrobox is installed by default nowadays on MicroOS, maybe we can change the line to specify this?
No need to install it at all

FarisFiroz:
If I recall correctly, MicroOS installs distrobox automatically on most auto-installs but the "minimal install" doesn't have it installed by default. For that reason I don't think removing it completely is the best idea. We could maybe add "if it is not installed by default", but then again that seems a bit unnecessary.

89luca89:
Ah right, no problem then :)

Thanks a lot for the PR!

Issue #654, opened 2023-03-12
[Suggestion] add installation note for EndlessOS


btreut:
**Is your feature request related to a problem? Please describe.**
Endless OS does not have curl installed, so installation of distrobox needs quite some digging before proceeding. But since Endless OS 4.0.0 was mentioned as host system, I tried to overcome that

**Describe the solution you'd like**
add a note that wget might be used instead of curl to download the install script.
 
**Additional context**
EndlessOS has podman and wget installed by default. The install script runs without apparent problems as user and after adding `.local/bin` to the path works without apparent problems.

PS: Molto grazie per distrobox

89luca89:
@btreut maybe we can add curl *and* wget instructions in the readme?

The install.sh itself automatically detects if it has to use wget or curl, so what is missing is just the installation instruction, am I right?

btreut:
thanks a lot, looks good ...

PR #663, opened 2023-03-19
List wget as an alternative install option

mirkobrombin:
closes https://github.com/89luca89/distrobox/issues/654


89luca89:
Thanks a lot @mirkobrombin :heart:

PR #669, opened 2023-03-23
fix(nvidia): mount libcuda.so*, this matches nvidia-container-toolkit

mjlbach:
Not sure if there should also be a check if this exists on host.

### DIFF
--- a/distrobox-init
+++ b/distrobox-init
@@ -947,6 +947,8 @@ if [ "${nvidia}" -eq 1 ]; then
 	# because we need to extract the relative path to mount them in the
 	# correct path based on the guest's setup
 	NVIDIA_LIBS="$(find /run/host/usr/lib* -iname "*nvidia*.so*" || :)"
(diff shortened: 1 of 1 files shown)


distrobox-init
```diff
@@ -947,6 +947,8 @@ if [ "${nvidia}" -eq 1 ]; then
 	# because we need to extract the relative path to mount them in the
 	# correct path based on the guest's setup
 	NVIDIA_LIBS="$(find /run/host/usr/lib* -iname "*nvidia*.so*" || :)"
```
89luca89, line 949: We can just add `-o -iname "*cuda*.so*"` to this command and all will be ok :+1:

89luca89 (changes requested):

distrobox-init
mjlbach: Feel free to go ahead and make the fix, just putting this out there as an example :) Not sure I would link all host libraries into the container though. There might be distro specific differences

PR #675, opened 2023-03-27
enter: fix $PATH when it contains spaces

black-desk:
When the $PATH in host contain spaces, iterating with "for ... in ..."
will produce a wrong result.

refer to https://stackoverflow.com/a/44524309

### DIFF
--- a/distrobox-enter
+++ b/distrobox-enter
@@ -345,11 +345,11 @@ generate_command() {
 		fi
 	done
 	# Ensure the $PATH entries from the host are appended as well
-	for standard_path in $(echo "${PATH}" | tr ':' ' '); do
+	while read -d ':' standard_path; do
 		if [ -n "${container_paths##*:"${standard_path}"*}" ]; then
 			container_paths="${container_paths}:${standard_path}"
 		fi
-	done
+	done <<< "$PATH:"
 	result_command="${result_command} --env \"PATH=${container_paths}\""
 
 	# Ensure the standard FHS program paths are in XDG_DATA_DIRS environment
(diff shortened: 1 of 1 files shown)


89luca89:
Hi @black-desk 

We will need to find another way to implement this, as your implementation is `bash` while distrobox adheres to posix `sh` standards

black-desk:
> Hi @black-desk
> 
> We will need to find another way to implement this, as your implementation is `bash` while distrobox adheres to posix `sh` standards

done.

I not familiar shell programming, you better have a check.

luc14n0:
Hi folks,

The simplest and most elegant and efficient solution here is:

``` diff
--- a/distrobox-enter
+++ b/distrobox-enter
@@ -345,11 +345,13 @@ generate_command() {
                fi
        done
        # Ensure the $PATH entries from the host are appended as well
-       for standard_path in $(echo "${PATH}" | tr ':' ' '); do
+       ORIG_IFS="${IFS}"; IFS=:
+       for standard_path in ${PATH}; do
                if [ -n "${container_paths##*:"${standard_path}"*}" ]; then
                        container_paths="${container_paths}:${standard_path}"
                fi
        done
+       IFS="${ORIG_IFS}"
        result_command="${result_command} --env \"PATH=${container_paths}\""
 
        # Ensure the standard FHS program paths are in XDG_DATA_DIRS environment
```

Just like said in the Stack Overflow post, I can confirm it's POSIX compliant. Note that using `unset IFS` -- in place of backing up the variable -- makes it lose part of its value and being only `$` (a new line) instead of the pristine value ` ^I$` (a space, a tab, and a new line).

black-desk:
@luc14n0 I have update this PR to use your solution.

89luca89:
Thanks @luc14n0 @black-desk for the fix!
Merging

PR #681, opened 2023-03-31
ci: trigger differential-shellcheck workflow on push

jamacku:
Also, update `differential-shellcheck` to the latest version (`v4.0.2`) - https://github.com/redhat-plumbers-in-action/differential-shellcheck/releases Push events are supported since `v4.0.0`.

Fixes: https://github.com/redhat-plumbers-in-action/differential-shellcheck/issues/215

This should get rid of annoying messages from the `github-code-scanning` bot.

Screenshot from 2023-03-22 07-00-32

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -78,13 +78,11 @@ jobs:
           exit $result
 
   differential-shellcheck:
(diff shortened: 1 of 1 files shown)


89luca89:
Thanks a lot @jamacku !!

PR #687, opened 2023-04-04
Fix executable scripts lookup

mkoura:
* The `distrobox-*` executable scripts don't need to be in the same dir as the `distrobox-create` script. The only exception is the `distrobox-generate-entry` script. This change unifies the behavior.
* Fix script name: `distrobox-hostexec` -> `distrobox-host-exec`


89luca89:
Thanks a lot @mkoura !

PR #691, opened 2023-04-07
added nix portable as an option for nix options

dnkmmr69420:


89luca89:
Thanks a lot @dnkmmr69420

PR #696, opened 2023-04-08
More aesthetically pleasing install script

misobarisic:
Makes the install prints more readable by not making it all red. Also shows the manpage path which gets used.
Only warns about PATH if `$dest_path` is not already in it.

image


89luca89:
@misobarisic thanks a lot!

Can you do the same for the uninstall script also?

misobarisic:
Sure thing
image

misobarisic:
Should I also add info about completions and icons, leave it as is or just shoe general info about using `...../share`?

89luca89:
> Should I also add info about completions and icons, leave it as is or just shoe general info about using `$prefix/share`?

I think it's already plenty as is :) thanks a lot!

PR #698, opened 2023-04-09
enter: fix unexpected pathname expansions

akinomyoga:
With the current implementation, pathname expansions are performed for the contents of arguments. For example, with the `git` command (which is a distro_binary script generated by `distrobox-export` and passes the arguments to `distrobox-enter`), the following command results in the expansion of filenames in the current directory (even after the fix #697).

```bash
$ git commit -m ' * '
```

To fix this, the argument of `eval` needs to be quoted properly.

### DIFF
--- a/distrobox-enter
+++ b/distrobox-enter
@@ -275,10 +275,9 @@ fi
 #   prints the podman or docker command to enter the distrobox container
 generate_command() {
 	result_command="${container_manager} exec"
-	result_command="${result_command} \\
-		--interactive \\
-		--detach-keys=\"\" \\
-		--user=\"${USER}\""
+	result_command="${result_command} --interactive"
+	result_command="${result_command} --detach-keys=\"\""
+	result_command="${result_command} --user=\"${USER}\""
 
 	# For some usage, like use in service, or launched by non-terminal
 	# eg. from desktop files, TTY can fail to instantiate, and fail to enter
@@ -287,8 +286,7 @@ generate_command() {
 	# work in tty-less situations.
 	# Disable tty also if we're NOT in a tty (test -t 0).
 	if [ "${headless}" -eq 0 ] && [ -t 0 ]; then
-		result_command="${result_command} \\
-			--tty"
+		result_command="${result_command} --tty"
 	fi
 
 	# Entering container using our user and workdir.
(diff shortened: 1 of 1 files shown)


distrobox-enter
```diff
@@ -317,9 +317,9 @@ generate_command() {
 		# Skipping workdir we just enter $HOME of the container.
 		workdir="${container_home}"
 	fi
-	result_command="${result_command}
-		--workdir=\"${workdir}\"
-		--env \"CONTAINER_ID=${container_name}\"
+	result_command="${result_command} \\
+		--workdir=\"${workdir}\" \\
+		--env \"CONTAINER_ID=${container_name}\" \\
```
89luca89, line 322: Could we do this without doing this double backslash?
I'd like to keep code readability and make it a bit less tricky

distrobox-enter
```diff
@@ -531,5 +531,4 @@ fi
 
 # Generate the exec command and run it
 cmd="$(generate_command)"
```
89luca89, line 533: If we do a `cmd="$(generate_command | tr '\n' ' ')"` we can get rid of the `\\` in the code, and quoted eval will continue to work :+1:

89luca89 (changes requested):

distrobox-enter
akinomyoga: I think that the intended newlines are also unexpectedly converted to spaces in that way. For example, the newlines specified by the user in the command line would be converted to spaces, e.g., in the case `git commit -m $'do something\n\nhello world'`.

89luca89: hmmm you're right, just moving it to line 391:

```sh
	result_command="$(echo "${result_command}" | tr '\n' ' ')"
	if [ -n "${container_command}" ]; then
		result_command="${result_command} ${container_command}"
	else
		# if no command was specified, let's execute a command that will find
		# and run the default shell for the user
		result_command="${result_command} sh -c \"\\\$(getent passwd ${USER} | cut -f 7 -d :) -l"\"
	fi
```

This fixes the issue

distrobox-enter
akinomyoga: One way is to connect all the lines in a single line, which looks somewhat ugly.

```bash
  result_command="${result_command} --workdir=\"${workdir}\" --env \"CONTAINER_ID=${container_name}\" --env \"DISTROBOX_ENTER_PATH=${distrobox_enter_path}\""
```

Another way is to add an argument to `result_command` in each line separately.

```bash
  result_command="${result_command} --workdir=\"${workdir}\""
  result_command="${result_command} --env \"CONTAINER_ID=${container_name}\""
  result_command="${result_command} --env \"DISTROBOX_ENTER_PATH=${distrobox_enter_path}\""
```

Or another way is to use the line continuation (with backslashes outside the quotation), though the readability is bad:

```
  result_command="${result_command} "\
    "--workdir=\"${workdir}\" "\
    "--env \"CONTAINER_ID=${container_name}\" "\
    "--env \"DISTROBOX_ENTER_PATH=${distrobox_enter_path}\""
```

akinomyoga: I added a commit 88d6ac132179ab3cb9e06b3888905bdcf052387b that implements the second option.

89luca89 (approved):

distrobox-enter
89luca89: Perfect!

89luca89:
All set then, thanks a lot for the work! :smile:

89luca89 (approved):

akinomyoga:
Thanks!

PR #702, opened 2023-04-10
init: Only install mesa vulkan packages if available

ericcurtin:
For example, these are not available on aarch64 CentOS Stream 9.


89luca89:
Thanks a lot @ericcurtin !

PR #718, opened 2023-04-23
Draft: install-podman: prefix choosing and system reset fixes

Meister1593:
So while i was making my script ([shameless plug](https://github.com/Meister1593/ALVR-Distrobox-Linux-Guidance/tree/master)) that heavily utilized rootless podman functionality, i noticed that something odd happening with system and non system installations of distrobox. 
So it took me a while to realize that it was modifying my home folder podman containers and actually resetting them (!).

I changed installation script a bit to utilise correct prefix if it was specified and excluded podman system commands from being ran at all as they are probably hardcoded to home folder and i didn't find a way to specify prefix for it. Not using them anyway doesn't matter for prefixed installation as prefix is already known.

### DIFF
--- a/install
+++ b/install
@@ -21,7 +21,7 @@
 
 next=0
 verbose=0
-version=1.6.0.1
+version=1.7.0
 
 # Print usage to stdout.
 # Arguments:


Meister1593:
I will leave this mr as draft as i discover more things attached to this as not being too portable

Meister1593:
Before proceeding, [podman issue](https://github.com/containers/podman/issues/18375) needs to be figured out/adressed

89luca89:
Thanks a lot @Meister1593 for this work!
Let me know when this is done for you :+1:

Meister1593:
> Thanks a lot @Meister1593 for this work! Let me know when this is done for you +1

Yeah this is currently pretty much done, but without rundir fixes are required so that it won't ever interfere with existing containers (and it will, i did find it out), so it's kindof blocked by this:
https://github.com/containers/storage/issues/1587

89luca89:
Interesting, let's keep an eye on that and revisit in the future :) 
Thanks a lot for your work!

PR #733, opened 2023-05-02
Fix spelling of openSUSE

grisu48:
Fix the spelling of openSUSE in the compatibility list


89luca89:
Thanks for the correction @grisu48 !

PR #734, opened 2023-05-02
Add distrobox ls

grisu48:
Adds the `distrobox ls` command, as a shortcut for `distrobox list`.

This PR is a suggestion to add the `ls` command, which is a shortcut to the `list` command to make it easier to list existing distro-boxes. The suggestion was inspired by the existence of the `rm` command, which is the shortcut for remove.

### DIFF
--- a/completions/distrobox
+++ b/completions/distrobox
@@ -2,6 +2,10 @@
 
 _generate_from_help() {
 	command=$1
+	# if command does not exist, try with `distrobox subcommand`
+	if ! $command --help 2>/dev/null ; then
+		command="$(echo $command | tr '-' ' ')"
+	fi
 
 	local list cur prev totalopts opts diropts
 
--- a/completions/distrobox
+++ b/completions/distrobox
@@ -37,7 +37,7 @@ _generate_from_help() {
 		COMPREPLY+=($(compgen -W "${opts}" -- ${cur}))
 		return 0
 	elif [[ ${command} != *"create"* ]] && [[ ${command} != *"ephemeral"* ]] &&
-		[[ ${command} != *"list"* ]] && [[ ${command} != *"assemble"* ]]; then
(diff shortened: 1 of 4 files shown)


89luca89:
Thanks @grisu48 
I think you can simply just keep the entry in `distrobox`
We don't need a dedicated `distrobox-ls` file (and the related change to the CI)

`ls` should just be a shorthand for `list`, so keeping it just in the switch/case of `distrobox` is enough for me

grisu48:
> Thanks @grisu48 I think you can simply just keep the entry in `distrobox` We don't need a dedicated `distrobox-ls` file (and the related change to the CI)
> 
> `ls` should just be a shorthand for `list`, so keeping it just in the switch/case of `distrobox` is enough for me

Hey, thanks for the feedback!

I tried to reduce it to `distrobox` only and am running into issues with the autocompletion. I have no strong opinion about this one and see two ways forward:

1. We don't care about autocompletion of `distrobox ls` - this would look like the following then:

```
phoenix@tumbleweed:~> distrobox ls distrobox-ls: command not found
distrobox-ls: command not found
distrobox-ls: command not found
```

Advantage: Code changes are very small, which is nice :slightly_smiling_face:   
Downside: No autocompletion for `distrobox ls`

2. We keep it as it is, this would mean introducing two symlinks and changes in `distrobox` and in autocompletion as well.

***

I have no preference, what do you like more?

grisu48:
@89luca89 ping? I'd prefer to still have autocorrect possible, but it's your call if simplicity should prevail :slightly_smiling_face:

89luca89:
Hi @grisu48 
yea I've tried a bit and indeed it's a problem for auto-completion

I've now pushed a workaround for it so we won't need more files :)

Thanks a lot!

grisu48:
Thank you for merging! This is great :slightly_smiling_face:

PR #742, opened 2023-05-07
terminal display fixes

cgzones:
* rm: add missing trailing newline

* list: reorder color reset

  The color might not get reset at the end if no actual character is printed and the output has reached the bottom of the terminal.

  Screenshot_20230507_203526


89luca89:
Thanks a lot @cgzones !

PR #747, opened 2023-05-09
rm: pass verbose mode to distrobox-generate-entry

cgzones:


89luca89:
Thanks a lot @cgzones !

PR #748, opened 2023-05-09
generate-entry: drop apparent debug statements

cgzones:
Commit 888d4561 ("generate-entry: add icon auto-download for known distros") added two printf statements printing the bare container name without any context.  This leads to superfluous on diststrobox-rm:

    $ distrobox rm test_alpine
    Do you really want to delete test_alpine? [Y/n]: y
    test_alpine
    test_alpine


89luca89:
Thanks again :)

PR #750, opened 2023-05-15
init: preserve the full SHELL if it points to the Nix store

midchildan:
Use the host `SHELL` if it is installed through Nix or Guix. This would improve consistency between the host and container environments. I made this change because container-provided shells often couldn't properly interpret my rc files, which targeted newer versions of the same shell. Sharing the same shell works because Nix packages are self-contained in the Nix store and usable both inside and outside of the container.

### DIFF
--- a/distrobox-init
+++ b/distrobox-init
@@ -294,7 +294,7 @@ shell_real_path="$(realpath "${SHELL}" 2> /dev/null || :)"
 case "${shell_real_path}" in
 	/nix/store/*|/gnu/store/*)
 		shell_pkg=""
-		if ! grep -q "${SHELL}" /etc/shells; then
+		if ! grep -q "^${SHELL}\$" /etc/shells; then
 			echo "${SHELL}" | tee -a /etc/shells
 		fi
 		;;
--- a/distrobox-init
+++ b/distrobox-init
@@ -290,12 +290,23 @@ if [ "${shell_pkg}" = "ash" ]; then
 	shell_pkg="busybox"
 fi
 
+shell_real_path="$(realpath "${SHELL}" 2> /dev/null || :)"
+case "${shell_real_path}" in
+	/nix/store/*|/gnu/store/*)
+		shell_pkg=""
+		if ! grep -q "${SHELL}" /etc/shells; then
+			echo "${SHELL}" | tee -a /etc/shells
+		fi
+		;;
+	*) ;;
+esac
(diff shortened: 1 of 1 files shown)


89luca89:
Hi @midchildan 
This seems to be a very nixos specific thing, I'm not sure I can maintain this functionality (as I don't use nixos, nor nix)

Maybe this could be achieved using the pre-init-hooks?

Thanks anyway for the PR

midchildan:
@89luca89 Unfortunately, there are currently no hooks that can override the logic for determining the shell.

Would you be open to an alternative approach that isn't Nix-specific? This PR leaves `$SHELL` as-is if it's a path that points to the Nix store. Instead, it might be safe to do the same thing when `$SHELL` has an exact match in the container's `/etc/shells`. Then, I can use the pre-init-hooks to add an entry to `/etc/shells` and get the desired result. Or maybe just provide an option to preserve `$SHELL`.

89luca89:
@midchildan 
pre-init hooks are performed before the package management, so you should be able to use some custom logic there

midchildan:
@89luca89 That's sort of the problem. Even when I use the pre-init hooks to specify `$SHELL`, it gets overridden.

For example, when I do `--pre-init-hooks "SHELL=~/.nix-profiles/bin/zsh"`, it first strips everything before the last path component:

https://github.com/89luca89/distrobox/blob/f594405c64ea5360e7a93f3edac79de9f5776aaf/distrobox-init#L286

Then, it goes on to install a separate copy of the shell:

https://github.com/89luca89/distrobox/blob/f594405c64ea5360e7a93f3edac79de9f5776aaf/distrobox-init#L775

Finally, it overrwrites `$SHELL`:

https://github.com/89luca89/distrobox/blob/f594405c64ea5360e7a93f3edac79de9f5776aaf/distrobox-init#L881

What I'm looking for is a way to

1. keep `$SHELL` untouched
2. skip the installation of the shell (because it's provided by the host through a shared mount)

89luca89:
mmh I get it now, thanks for the explanation :)

Let's do this, let's convert this in a draft
open an issue linked to this, and we can try to discuss a more generic way to do is
less "nixy" 

I'd like to keep the code as generic as possible and introduce quirks only if really strictly necessary (like we did in create for example)

89luca89:
Thanks :)

Issue #708, opened 2023-04-19
[Suggestion] Add flag to distrobox-export to not copy app icons over


VioletRoses:
**Is your feature request related to a problem? Please describe.**
On my arch box, I have the icon theme papirus installed and linked to my ``~/.local/share/icons`` so that I can use the theme system-wide and Fedora Silverblue doesn't have to mess with the files every time the ostree is updated (as the sheer amount of files massively increases commit time). However, this causes ``distrobox-export`` to try to copy the icons to themselves, giving an error like so:
``cp: '/usr/share/icons/Papirus/16x16/apps/godot.svg' and '/run/host//home/rose/.local/share/icons/Papirus/16x16/apps/godot.svg' are the same file``

**Describe the solution you'd like**
A var/flag setting to disable this behavior, or at least make the script continue after this error.

**Describe alternatives you've considered**
I've considered creating separate volumes for each theme folder (Papirus, Papirus-light, Papirus-dark, etc) so they can have different names on the host than on the guest, but it feels kind of hacky and I'm not sure if it would even work correctly because the theme folders have a lot of symlinks to each other.
For now I just manually create .desktop files for integration.

PR #760, opened 2023-05-21
Fix missing icons error, distrobox-export

bubuntux:
Some applications have symbolic link to non existing icons (Zoom for example) and this prevent exporting such apps into the host system, by simply checking the file exist before copying solves this problem

### DIFF
--- a/distrobox-export
+++ b/distrobox-export
@@ -396,7 +396,9 @@ export_application() {
 
 		# we wanto to export the application's icons
 		mkdir -p "${icon_home_directory}"
-		cp -r "$(realpath "${icon_file}")" "${icon_home_directory}"
+		if [ -e "$(realpath "${icon_file}")" ]; then
+			cp -r "$(realpath "${icon_file}")" "${icon_home_directory}"
+		fi
 	done
 
 	# create desktop files for the distrobox


89luca89:
Hi @bubuntux 

This was fixed in one of my latest pushes for another unrelated fix

Thanks anyway for the PR!

bubuntux:
@89luca89  i think the other commit fixes #708 that prevents icon override, this fix is different, in here, is preventing the cp command to fail if the source of the symbolic link doesn't exist, i may have to adjust the mr now that the code changed tho...

89luca89:
Hi @bubuntux 
the icon should exist as the path is found with `find` and if it doesn't exist it shouldn't be listed in the first place

bubuntux:
@89luca89  the problem is that the file exist, but is just a symbolic link when we do the copy we use `realpath` that resolves the link and is pointing into a non-existing file causing the `cp` command to fail and the app not getting exported, 

steps to reproduce is 
```
# create the container
distrobox create --image ubuntu:22.04
distrobox enter ubuntu-22-04

#Inside the container we install zoom
 sudo apt-get update -y && sudo apt-get install -y curl && curl -sSL "https://zoom.us/client/latest/zoom_amd64.deb" -o /tmp/zoom.deb  && sudo  apt-get install -y /tmp/zoom.deb 

#Export the app
distrobox-export --app Zoom
```

it results in  `cp: cannot stat '/usr/share/icons/Humanity/actions/24/zoom-best-fit.svg': No such file or directory`
this is because the file /usr/share/icons/Humanity/stock/24/stock_zoom-page.svg is a symbolic link pointing to /usr/share/icons/Humanity/actions/24/zoom-best-fit.svg that does not exist. 

so the fix still need it

89luca89:
Ok got it, makes sense then
I mean it's cleraly a broken package, but doesn't hurt to guard against it :)

PR #766, opened 2023-05-23
Ignore net files when unshare nets

bubuntux:
mounting the /etc/resolv.conf is problematic using podman and dnsmasq is setup with network manager, since it doesn't resolve the real dns and tries to use the local dnsmasq server, resulting on dns not working.

mounting these fields seems unnecessarily when not sharing the network

### DIFF
--- a/distrobox-create
+++ b/distrobox-create
@@ -653,7 +653,6 @@ generate_command() {
 			fi
 		done
 	fi
-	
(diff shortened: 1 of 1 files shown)


89luca89:
Thanks @bubuntux this actually makes a lot of sense :) Thanks!

PR #768, opened 2023-05-24
compatibility.md: add deepin support to the documentation

BLumia:
deepin now support `distrobox` since deepin v23 (beige).

Users can also install deepin v20 (apricot) and v23 (beige) from `distrobox`.

Examples:

- To install distrobox on deepin v23: `sudo apt install distrobox`
- To create a deepin v23 box: `distrobox create beige --image docker.io/linuxdeepin/beige`

Sidenote: The deepin branding guideline says the name "deepin" starts with a lower-case `d` is the preferred way to use, similar to openSUSE starts with a lower-case `o`. So, in this patch we use "deepin" instead of "Deepin" here, which is not a typo :)

cc @black-desk


black-desk:
Maybe you should add icon as well. 

https://github.com/89luca89/distrobox/tree/main/docs/assets/png/distros

89luca89:
Thanks a lot @BLumia 

If the CI finishes successfully I'll merge thanks a lot! :)

PR #792, opened 2023-06-12
Fix issue when a symbolic link points to a non-existing file

bubuntux:
reopening  #760  after watering the branch 
please review this comment https://github.com/89luca89/distrobox/pull/760#issuecomment-1588193431 this fix is actually need it

### DIFF
--- a/distrobox-export
+++ b/distrobox-export
@@ -404,7 +404,7 @@ export_application() {
 
 		# we wanto to export the application's icons
 		mkdir -p "${icon_home_directory}"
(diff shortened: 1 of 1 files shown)


89luca89 (approved):

89luca89:
Thanks a lot @bubuntux :)

profetik-777:
Prob not worth it anymore, but just in case someone needs a temp solution for a situation where this was caused by removing a flatpak version of the same app (e.g. I removed zoom flatpak, and installed as a distrobox), you can do the following: 

1. flatpak repair command will "clean filepaths" up and allow for export thereafter.

PR #807, opened 2023-06-20
Fix user initialization for ubuntu 23.04

pettermk:
Ubuntu 23.04 comes preinstalled with a user named ubuntu and uid
1000. This conflicts with current logic and this change detect this situation and ensures that this is detected and remediated


89luca89:
Perfect :) thanks a lot for this!

PR #812, opened 2023-06-21
Fix typo in instructions for setting podman-host path

bitestring:
`podman-exec` should be `podman-host` in this instruction

```
Open VSCode settings (Ctrl+,) and head to `Remote>Containers>Docker Path` and
set it to the path of `podman-exec`, like in the example
```


89luca89:
Thanks a lot!

PR #821, opened 2023-06-24
Remove --name since never used

mirkobrombin:
I might be wrong but the `--name` flag is never used in the source, so removing from the help message.

```
distrobox rm --name apx-alpine --force
ERROR: Invalid flag '--name'

distrobox version: 1.5.0.2

Usage:

	distrobox-rm [-f/--force] container-name [container-name1 container-name2 ...]

Options:

	--name/-n:		name for the distrobox
	--force/-f:		force deletion
	--rm-home:		remove the mounted home if it differs from the host user's one
	--root/-r:		launch podman/docker with root privileges. Note that if you need root this is the preferred
				way over "sudo distrobox" (note: if using a program other than 'sudo' for root privileges is necessary,
				specify it through the DBX_SUDO_PROGRAM env variable, or 'distrobox_sudo_program' config variable)
	--help/-h:		show this message
	--verbose/-v:		show more verbosity
	--version/-V:		show version
```

### DIFF
--- a/distrobox-rm
+++ b/distrobox-rm
@@ -103,7 +103,6 @@ Usage:
 
 Options:
 
-	--name/-n:		name for the distrobox
 	--force/-f:		force deletion
 	--rm-home:		remove the mounted home if it differs from the host user's one
 	--root/-r:		launch podman/docker with root privileges. Note that if you need root this is the preferred
--- a/docs/usage/distrobox-rm.md
+++ b/docs/usage/distrobox-rm.md
@@ -32,6 +32,6 @@ You can also use environment variables to specify container manager and name:
 # ENVIRONMENT VARIABLES
 
 	DBX_CONTAINER_MANAGER
-	DBX_CONTAINER_
+	DBX_CONTAINER_NAME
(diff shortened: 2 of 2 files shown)


bubuntux:
@mirkobrombin  probably also want to remove it from the md file (distrobox-rm.md)

mirkobrombin:
> @mirkobrombin probably also want to remove it from the md file (distrobox-rm.md)

Thanks, I missed it.

luc14n0 (changes requested):
Yeah, that option flags is currently not implemented in the script. So, for me, it makes sense to drop those `--name` mentions.

If you could also drop it from the manpages:

``` sh
$ rg 'distrobox[- ]rm --name'
man/man1/distrobox.1
1818:distrobox-rm --name container-name [--force]

man/man1/distrobox-rm.1
51:distrobox-rm --name container-name [--force]

docs/usage/distrobox-rm.md
27:     distrobox-rm --name container-name [--force]
```

Then we could hope and wait for Luca's final word on this change :+1:

89luca89:
Thanks a lot! :heart:

PR #823, opened 2023-06-26
Add a flag to upgrade only running instances

bubuntux:
i created multiple instances for different projects, and updating all instances means starting all instances that some do heavy operations at startup, so i want to keep them up-to-date but only for the ones i'm currently using so i thought that --up variable could be useful to add for other/similar usages

### DIFF
--- a/distrobox-upgrade
+++ b/distrobox-upgrade
@@ -31,7 +31,7 @@ if [ -n "${SUDO_USER}" ] || [ -n "${DOAS_USER}" ]; then
 fi
 
 all=0
-up=0
+running=0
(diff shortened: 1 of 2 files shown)


89luca89:
Thanks @bubuntux for the PR
I'd say the idea is ok, but I'd like a more "self-explanatory" term than `up`, maybe a `--only-running` or `--only-started`?

89luca89:
Quickly fixed it, I also prefer to be paired with all

so it's going to be `distrobox-upgrade --all --running` for all the running distroboxes

Thanks for the idea and the contribution! :smile:

Issue #826, opened 2023-06-28
[Error] Archlinux container with additional packages reinstalls all basic packages in every restart


Lin-Junhao:
**Describe the bug**
I created an Archlinux container with init system support using this command:

```
$ distrobox create --image archlinux:latest --name distrobox-archlinux --additional-packages "systemd" --init
```

I notice that during every restart, it reinstalls all basic packages along with systemd. Also it fails to launch if there is no internet connection. Since Archlinux image has systemd built-in, I tried creating another container without `--additional-packages "systemd"` and it behaves normal.

**To Reproduce**
On the host machine:
```
$ distrobox stop distrobox-archlinux
$ distrobox enter distrobox-archlinux
```
It spent a long time on "Installing basic packages...".

**Expected behavior**
Distrobox should detect dependencies correctly and that the container has done setting up, launches the container without reinstalling packages.

**Logs**
https://0x0.st/HQEw.log

**Desktop**
- Host OS: SteamOS 3.4.8
- Container OS: Arch Linux
- Podman version: 4.5.1
- Distrobox version: 1.5.0.2

Lin-Junhao:
I've updated my log file to only include a single restart.

https://github.com/89luca89/distrobox/blob/55052318202388ac17886f4e8cc12ed93d13ae0d/distrobox-init#L330-L336

@LukeShortCloud What do think of the conditions here? I believe if any of these conditions are met it will trigger the upgrade? But in a restart the container should have had all these dependencies and `/run/.containersetupdone` file as well. I don't know what went wrong, here are the corresponding logs:

<details open>
<summary>$ podman logs distrobox-archlinux</summary>

```
+ '[' 0 -ne 0 ']'
+ command -v find
/usr/sbin/find
+ command -v mount
/usr/sbin/mount
+ command -v passwd
/usr/sbin/passwd
+ command -v sudo
/usr/sbin/sudo
+ command -v useradd
/usr/sbin/useradd
+ command -v diff
/usr/sbin/diff
+ command -v pinentry
/usr/sbin/pinentry
+ command -v wget
/usr/sbin/wget
+ command -v curl
/usr/sbin/curl
+ command -v less
/usr/sbin/less
+ command -v bc
/usr/sbin/bc
time
+ command -v time
+ command -v lsof
/usr/sbin/lsof
+ command -v bash
/usr/sbin/bash
+ '[' -n ' systemd' ']'
+ '[' '!' -e /run/.containersetupdone ']'
```
</details>

[...]

PR #828, opened 2023-06-28
Arch Linux speed improvements

LukeShortCloud:
This PR makes the following changes:

- Pacman repository cache is only updated once, at the beginning, instead of every command.
- Pacman will no longer re-install packages. Without the `--needed` argument, `pacman -S` will re-install packages.
- Pacman upgrades are no longer run every time. They only run if, from within the container, `distrobox-init --upgrade` is ran. This is the expected behavior.

This addresses performance concerns brought up here: https://github.com/89luca89/distrobox/issues/826

I manually tested this and can confirm that re-entering a stopped Arch Linux container is much faster and that upgrades still work.

### DIFF
--- a/distrobox-init
+++ b/distrobox-init
@@ -617,9 +617,12 @@ if [ "${upgrade}" -ne 0 ] || ! command -v find || ! command -v mount || ! comman
 		fi
 
 	elif command -v pacman; then
+		# Update the package repository cache exactly once before installing packages.
+		pacman -S -y -y
+
 		# If we need to upgrade, do it and exit, no further action required.
 		if [ "${upgrade}" -ne 0 ]; then
-			pacman -Syyu --noconfirm
+			pacman -S -u --noconfirm
 			exit
 		fi
 		# In archlinux official images, pacman is configured to ignore locale and docs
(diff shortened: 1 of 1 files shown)


D3vil0p3r:
It is useful. The usage of `pacman -Sy <pkgname>` is very dangerous and not supported on Arch. It must be used only for global upgrading.
Source: https://wiki.archlinux.org/title/System_maintenance#Partial_upgrades_are_unsupported

xz-dev:
It's merged to [dustrobox-boost](https://github.com/xz-dev/distrobox-boost/blob/d8333f10453300aa0d73c81257014de9e8975ec0/src/distro/package_manager.rs#L16), you can use it to create image for distrobox faster.

LukeShortCloud:
@xz-dev Thanks for sharing! Your project looks very promising: a faster Rust implementation of `distrobox` with additional tweaks added on-top.

@89luca89 Is there any chance we can also get this PR merged into upstream `distrobox`? I've been using this for a month for my Arch Linux container and it has been working great. Apparently `distrobox-boost` takes a similar approach for package management on Arch Linux. I'm open to feedback!

89luca89:
Thanks @LukeShortCloud 
That's very appreciated :)

PR #834, opened 2023-07-04
feat: add user arg for enter command

xz-dev:
Allow users to choose their own login user instead of binding to the user of the current environment.
This is helpful to use root user inside the container to prevent permissions invalidation issues of heterogeneous containers (such as riscv64/debian:sid) or use other usernames to isolate file permissions.

### DIFF
--- a/distrobox-enter
+++ b/distrobox-enter
@@ -21,7 +21,7 @@
 # POSIX
 # Expected env variables:
 #	HOME
-#	USER
+#	USER || ENTER_USER
 # Optional env variables:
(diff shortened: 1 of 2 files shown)


xz-dev:
I find the change of assemble is unnecessary, I will back them latter.

89luca89:
Hi @xz-dev thanks for this PR

But this is not in the scope of distrobox, as it's target is only to integrate with the current user's userland
Thanks anyway for the PR, and the interest in improving the project!

PR #840, opened 2023-07-05
Should fix issues with paths containing spaces

Silenc3IsGold3n:
Found an issue when distrobox-export when distrobox paths contain spaces that would cause the export file to point to the wrong location.  Encapsulating the path with quotations should fix this issue.


89luca89:
Thanks @Silenc3IsGold3n 
But if you want to prevent this, you need to escape the quotes (`\"${...`) or they won't be written in the wrapper file

PR #850, opened 2023-07-11
create: fix pure flake NixOS system mounted files

mcanueste:
Hi, first of all, thanks for the great work. Distrobox is really useful!

This PR is for fixing the mounting issue specific to pure flake NixOS systems (see [flakes](https://nixos.wiki/wiki/Flakes)).

## The issue:

Pure flake NixOS systems use `/etc/profiles/per-user/` and `/ect/static/profiles/per-user` paths for symlinking the installed applications instead of `~/.nix-profile/` and `/nix/var/nix/profiles` (see [nix profiles](https://nixos.org/manual/nix/stable/package-management/profiles.html)).

So the linking goes like this:

```
/etc/profiles/per-user/alice -> /etc/static/profiles/per-user/alice -> /nix/store/...
```

And the `PATH` environment variable is set like this:

```
PATH=...:/home/alice/.nix-profile/bin:/etc/profiles/per-user/alice/bin:/nix/var/nix/profiles/default/bin:/run/current-system/sw/bin:...
```

Since the system is installed via flakes purely, only the `/etc/profiles/per-user/alice` is defining proper path to the installed application symlinks that are pointing to `/nix/store/...`.

This requires mounting these directories on `create` as otherwise the symlinks that exist under these paths are not mounted to the box on `distrobox create`, resulting in not being able to find the installed applications of the host system inside the box.

## The fix:

Adding the following dirs to [this section](https://github.com/89luca89/distrobox/blob/40fdc93ff1b5a450abbccdb8503fdc8bd63bd449/distrobox-create#L577) fixes the issue:

- `/etc/profiles/per-user`
- `/etc/static/profiles/per-user`

I am not aware of any distro using these paths by default, so I think this would not break distrobox usage for other distros.

## Additional info:

The `/env/profiles/per-user` content:

```
$ ls -alF /etc/profiles/per-user/
lrwxrwxrwx 34 root alice -> /etc/static/profiles/per-user/mcst/
```

The `/env/static/profiles/per-user` content:

```
$ ls -alF /etc/static/profiles/per-user/
lrwxrwxrwx 60 root alice -> /nix/store/hd3jlybnnlsshiwbm0cqm84n34j405kd-user-environment
```

The `/nix/store/hd3jlybnnlsshiwbm0cqm84n34j405kd-user-environment` content:

```
$ ls -alF /nix/store/hd3jlybnnlsshiwbm0cqm84n34j405kd-user-environment
dr-xr-xr-x - root bin
dr-xr-xr-x - root etc
dr-xr-xr-x - root lib
dr-xr-xr-x - root libexec
dr-xr-xr-x - root sbin
dr-xr-xr-x - root share
```

The NixOS version: NixOS Unstable (23.05)

### DIFF
--- a/distrobox-create
+++ b/distrobox-create
@@ -578,6 +578,8 @@ generate_command() {
 		/nix
 		/gnu
 		/run/current-system/sw
+		/etc/profiles/per-user
+		/etc/static/profiles/per-user
 	"
 	for nix_dir in ${nix_dirs}; do
 		if [ -d "${nix_dir}" ]; then


89luca89:
Hi @mcanueste 

to be honest I was thinking of completely removing this integration (`/nix`, `/run/current-sw` and so on)
I've been using NixOS for some time, and I feel like distrobox should integrate with host's data (home, runtime) but NOT the system

Doesn't make much sense to me to bring nix stuff inside the container **by default** and instead
I'd say if NixOS's maintainers want to keep the integration, they can ship a custom config (in /etc, or whatever) that has some customization

mcanueste:
Thanks @89luca89 .

I understand. After using distrobox a bit with NixOS, I also realized that loading the system inside the container make things a bit buggy/weird due to how NixOS system works compared to other distros (i.e. had to do hacky workarounds). 

I was also able to find a nix solution to the use case I had, which was the initial reason for looking into distrobox. I guess, there is not many use cases where loading NixOS system into containers would be needed, and maintaining support for those rare use cases might not be worth it.

I am closing this PR.

89luca89:
Thanks @mcanueste 
I felt the same while using it on NixOS so I'll probably implement the removal of these mounts in the next days

I just pushed a config change that permits the customization of additional mounts for all containers, so if NixOS packagers want to reintroduce the mounts, they can

PR #859, opened 2023-07-13
Export to host_home by default

bubuntux:
when you create a container using a diff home path, you export the app/bin into the container home rather than the host home, since the idea of exporting is to make available that application on the host system i think we need to export by default into the host_home instead.


89luca89:
Thanks a lot @bubuntux!

PR #860, opened 2023-07-13
fix: --sudo when not --login

bubuntux:
when you have --sudo and not --login the start_shell is empty when we should have "sudo', this PR fixes that


89luca89:
Thanks a lot again :) @bubuntux

PR #861, opened 2023-07-14
Add KDE neon icon and pointers to container registry

jriddell:
At KDE we've made our own packages and distro called KDE neon for some years and I've always created Docker images for that.
https://neon.kde.org/
https://community.kde.org/Neon/Docker
Ages ago I made a script neondocker which did what Distrobox does, so it makes sense to move to use Distrobox instead.  Here's adding an icon and pointers to our container repository.


89luca89:
Thanks a lot @jriddell very appreciated :)

Issue #797, opened 2023-06-14
[Discussion] Improvements to distrobox-export --bin


Scafir:
Hi,

Following the discussion with @fredizzimo in #651, I think that it is best to discuss how to tackle the exported binaries.

First of all, let me write down the goal (in my opinion) that we should strive for with binaries export:
"Exported binaries should produce the same result than the same command run from inside the distrobox"

To reach this goal, I suggest using a login shell by default, as not using it may only cause hard to debug issues, with no clear advantage.
I would also argue that using a login shell and executing the command within that also solve an issue raised in the original merge request: in what shell should the command be run?
There are three level to this:
1. What shell interprets the arguments given? (globbing, etc) -> the shell from which the exported binary is called
2. What shell actually executes the exported binary? -> ??
3. What is the $SHELL present while executing the exported binary? -> the default shell set inside the distrobox

I would argue that number 2 is not actually important, as no matter the shell, the end result should be the same.
This allows us to use a single shell for all use case inside the exported script. This is a big advantage over using $SHELL, as it allows us to avoid having to determine what syntax should be used (e.g: fish does not use `$@`). 

Being careful avoiding issues with binary files that may not be executed with a simple `sh -l myBinaryFile`, we would get a end result along the lines of:
```
distrobox-enter ${rootful} -n  myDistrobox -- ${is_sudo} sh -lc 'path/to/command "$@"' command ${extra_flags} "$@"
```
I will most likely submit a merge request for this, do not hesitate to let me hear your thoughts.

89luca89:
Thanks @Scafir for the issue!

I'm not completely sure that using the login shell solves all the escaping/globbing problems:

`~$ sh -lc "ls *"`

the glob is still resolved by the main shell before calling sh, you need to escape that glob anyway, and we're back to square one :) 

the shell that calls the exec is known, and it's `/bin/sh` because that's what is in the shebang of the wrapper script in the bin

fredizzimo:
One thing that the login shell helps with, is dealing with environment variables and isolated environments as described in this issue https://github.com/89luca89/distrobox/issues/656, towards the bottom of it.

89luca89:
Hi @Scafir thanks a lot for your effort!

I'm  a bit stunned by the `sh -lc 'path/to/command "$@"' commandName "$@"` syntax to be honest :joy: 

I'd prefer to keep the current escaping algorithm as it's already tested from a few releases

89luca89:
[...]

PR #869, opened 2023-07-18
distrobox-export: improvements, login shell, escaping

Scafir:
This pull request follow the discussion #797. Here are the main changes:
* Do not use sudo if it is not necessary.
* Improve escaping mechanism, always run with sh
* Always use login shell, remove option to do so


89luca89:
Hi @Scafir thanks a lot for your effort!

I'm a bit stunned by the sh -lc 'path/to/command "$@"' commandName "$@" syntax to be honest joy

I'd prefer to keep the current escaping algorithm as it's already tested from a few releases

89luca89:
Hi @Scafir thanks for the PR

After thinking about it, I prefer keeping the current setup, but I'll make `login` option as default as you suggested
Thanks for the work on the PR!

PR #873, opened 2023-07-22
Make distrobox-stop use default container name if none is supplied

ricardoseriani:


89luca89:
Thanks a lot for the PR @ricardoseriani

PR #886, opened 2023-07-31
replace posix-libc-utils with libc-utils

xz-dev:
posix-libc-utils is not exist at alpine, it should been libc-utils


xz-dev:
I had been fixed it at [distrobox-boost](https://github.com/xz-dev/distrobox-boost/blob/d8333f10453300aa0d73c81257014de9e8975ec0/src/distro/distrobox_requirements.rs#L17), if anyone need it.

89luca89:
Thanks a lot for the fix!

PR #889, opened 2023-08-03
export: Add call to update-desktop-database tool

luc14n0:
That's important to ensure exported applications will show up in whatever the user/DE is using desktop files for (taskbars, desktop menus, etc) right after exportation.

Desktop Environments should already install `update-desktop-database` by default - that's how they handle packages installing `.desktop` files. I'd believe that most Window Managers should also have it.

### DIFF
--- a/distrobox-export
+++ b/distrobox-export
@@ -45,7 +45,7 @@ verbose=0
 version="1.5.0.2"
 
 # We depend on some commands, let's be sure we have them
-base_dependencies="basename grep sed find"
+base_dependencies="basename find grep sed update-desktop-database"
 for dep in ${base_dependencies}; do
 	if ! command -v "${dep}" > /dev/null; then
 		printf >&2 "Missing dependency: %s\n" "${dep}"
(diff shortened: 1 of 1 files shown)


89luca89:
Thanks @luc14n0 

If we add a base dependency we need to add it to the `init` also

Instead of this, I was thinking we probably want to update the host's desktop-database
So instead of this maybe we can do this:

`	/usr/bin/distrobox-host-exec --yes update-desktop-database -q 2> /dev/null || :`

So this will either force-update the desktop-db on the host, or fail silently if the host doesn't have the tool

luc14n0:
> Instead of this, I was thinking we probably want to update the host's desktop-database

Duh, my bad.

> So instead of this maybe we can do this:
> 
> `	/usr/bin/distrobox-host-exec --yes update-desktop-database -q 2> /dev/null || :`
> 
> So this will either force-update the desktop-db on the host, or fail silently if the host doesn't have the tool

Sounds good to me.

89luca89:
Added the fix quickly and merging :) 

Thanks a lot!

luc14n0:
Sure thing. Thanks for the amendment.

Issue #895, opened 2023-08-06
[Error] assemble can wait for unprompted user input, with possibly destructive results


AdamIsrael:
**Describe the bug**

Scenario: I've defined a container with a custom home directory but if I run `assemble create` twice, the second run will prompt for user input without notification, and the wrong keystroke can lead to the deletion of the ACME folder.

**To Reproduce**

```
[acme]
image=docker.io/library/ubuntu:22.04
home=/var/home/stone/Clients/ACME
init=false
nvidia=false
pull=true
root=false
replace=true
```

For the first run, everything works as expected:

```
$ distrobox assemble create --file ~/Clients/distrobox.ini
 - Deleting acme...
Error: no such container acme
Cannot find container acme.
 - Creating acme...
Trying to pull docker.io/library/ubuntu:22.04...
Getting image source signatures
Copying blob 3153aa388d02 skipped: already exists
Copying config 5a81c4b850 done
Writing manifest to image destination
5a81c4b8502e4979e75bd8f91343b95b0d695ab67f241dbed0d1530a35bde1eb
Creating 'acme' using image docker.io/library/ubuntu:22.04	 [ OK ]
Distrobox 'acme' successfully created.
To enter, run:

distrobox enter acme
```

The second run appears to hang while deleting the container:

```
$ distrobox assemble create --file ~/Clients/distrobox.ini
 - Deleting acme...
```

Under the covers, we're calling `distrobox rm -f ACME`, which is prompting for user input but the message is never raised to the user calling `distrobox assemble`:

```
$ distrobox rm -f acme
Do you want to remove custom home of container acme (/var/home/stone/Clients/ACME)? [y/N]:
```

**Expected behavior**

I would expect one of two things: either `distrbox rm -f acme` doesn't default to prompting for a destructive operation (see issue #894 for my suggestion) or the prompt for input is captured/displayed to the caller of `distrobox assemble` so they know what's expected of them.

Issue #894, opened 2023-08-06
[Discussion] By default, `distrobox rm` can be destructive when combined with a mounted home directory


AdamIsrael:
> Is your feature request related to a problem? Please describe.

Scenario: `distrobox rm` can delete the directory mounted in the container via `--home`.

```
$ distrobox rm -f acme
Do you want to remove custom home of container acme (/var/home/stone/Clients/ACME)? [y/N]: y
acme
Successfully removed /var/home/stone/Clients/ACME
```

I would make the opinionated argument that this is a dangerous prompt to display by default.

> Describe the solution you'd like

There is a `--rm-home` flag. In my mind, that flag should be required before prompting the user to remove the custom home directory. It's a destructive operation, so all caution should be taken to ensure the user really wants that to happen.

In other words, I think the command should look like this:

```
$ distrobox rm -f --rm-home acme
Do you want to remove custom home of container acme (/var/home/stone/Clients/ACME)? [y/N]: y
acme
Successfully removed /var/home/stone/Clients/ACME
```

Running the command without `--rm-home` should ignore the home directory and only delete the container.

PR #896, opened 2023-08-06
fix: Only prompt to delete home if `--rm-home` is passed

AdamIsrael:
Only prompt the user to delete their mounted home directory (if it differs from $HOME) if they pass the `--rm-home` flag.

```
$ distrobox-rm -f acme
acme
```

```
$ distrobox-rm -f --rm-home acme
Do you want to remove custom home of container acme (/var/home/stone/Clients/ACME)? [y/N]: y
acme
Successfully removed /var/home/stone/Clients/ACME
```

This is a suggested solution to the pending discussion in issue #894 and resolves the issue raised in #895.

### DIFF
--- a/distrobox-rm
+++ b/distrobox-rm
@@ -272,7 +272,7 @@ delete_container() {
 	# Prompt for confirmation
 	if [ "${container_home}" != "${HOME}" ]; then
 		if [ "${non_interactive}" -eq 0 ] &&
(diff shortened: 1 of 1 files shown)


89luca89:
Thanks! This is indeed very important, thanks a lot for the contribution! :smile:

PR #897, opened 2023-08-07
Custom command

rrahl0:
This Pull Request introduces a new flag called --custom-command <bin>. This creates directly a desktop file with the specified binary to enter. 

I see this use case for using tmux directly, so there is no need to actually export it manually. Additionally, when using it with ephemeral, you get directly "thrown" into tmux, no manually executing tmux necessary (and creates also the desktop file with tmux as the command).

Hope you could critique a bit more, as this is my first steps into programming, and my first PR.

### DIFF
--- a/distrobox-assemble
+++ b/distrobox-assemble
@@ -35,6 +35,7 @@ version="1.5.0.2"
 additional_flags=""
 additional_packages=""
 entry=""
+custom_command=""
 home=""
 image=""
 init=""
@@ -244,6 +245,9 @@ run_distrobox() (
 	fi
 	if [ -n "${entry}" ] && [ "${entry}" -eq 0 ]; then
 		result_command="${result_command} --no-entry"
+	fi
+	if [ -n "${custom_command}" ]; then
+		result_command="${result_command} --custom-entry $(sanitie_variable "${custom_command}")"
 	fi
 	if [ -n "${nvidia}" ] && [ "${nvidia}" -eq 1 ]; then
 		result_command="${result_command} --nvidia"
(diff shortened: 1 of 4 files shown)


89luca89:
Hi @rrahl0 !

I'm not sure about this I'll explain why:

1 - for `ephemeral` you can simply specify the command like:

`distrobox-ephemeral --exec tmux` or `distrobox-ephemeral -- tmux`

2 - for `generate-entry` one thing you could do is to set tmux as login shell for the user

inside the container use:

`sudo chsh -S /usr/bin/tmux $USER`

then your default shell will be tmux, and the entry icon will automatically launch it

rrahl0:
Ahh, didn't know there was an --exec for ephemeral, as it doesn't autocomplete (and I completely overlooked when getting this written down.)

with generate-entry idk, I think it's quite annoying if you create a few containers, that you would need to do with every single one of them. (I have mostly ini files, which I try to keep as ephemeral as possible).

89luca89:
> Ahh, didn't know there was an --exec for ephemeral, as it doesn't autocomplete (and I completely overlooked when getting this written down.)

yea ephemeral autocompletion should be better indeed :/

> with generate-entry idk, I think it's quite annoying if you create a few containers, that you would need to do with every single one of them. (I have mostly ini files, which I try to keep as ephemeral as possible).

you can use an `init_hook` with `passwd -s /usr/bin/tmux YOUR_USERNAME` and that's it :+1:

aleshmannie:
Thanks okay 👍 can I talk to you I got some world 🌎 informations I got to
practice are with I will let you know all this things are real let's fuck
tha beast together

PR #898, opened 2023-08-08
debian: succeed distrobox-init if custom image has undergone releaseinfo change

aquaherd:
If you use an old development image that has changed from stable to oldstable or from oldstable to oldoldstable between image creation and distrobox creation, plain apt-update will require you to interactively or explicitly allow it. This patch fixes both the first run and upgrade scenarios.

Additionally, if you use the vanilla PS1 prompt of a debian install on the host, debian defaults to prepend ($debian_chroot) to it. Setting this to either CONTAINER_ID or distrobox makes it easier.


89luca89:
Hi @aquaherd 

I'm not sure about this

In the 1st scenario (line 433) this should never happen, especially if you pull during create
In the 2nd scenario it would be better to add `Acquire::AllowReleaseInfoChange::Suite "true"` in /etc/apt.conf.d in line 1400?

aquaherd:
I just checked, the Acquire... hook does not work:
```
+ export debian_chroot=distrobox
+ [ 0 -ne 0 ]
+ rm -f /etc/dpkg/dpkg.cfg.d/excludes
+ export DEBIAN_FRONTEND=noninteractive
+ apt-get update
Get:1 http://security.debian.org/debian-security buster/updates InRelease [34.8 kB]
Get:2 http://deb.debian.org/debian buster InRelease [122 kB]
Get:3 http://deb.debian.org/debian buster-updates InRelease [56.6 kB]
Reading package lists...
E: Repository 'http://security.debian.org/debian-security buster/updates InRelease' changed its 'Suite' value from 'oldstable' to 'oldoldstable'
E: Repository 'http://deb.debian.org/debian buster InRelease' changed its 'Suite' value from 'oldstable' to 'oldoldstable'
E: Repository 'http://deb.debian.org/debian buster-updates InRelease' changed its 'Suite' value from 'oldstable-updates' to 'oldoldstable-updates'
+ [ 100 -ne 0 ]
+ printf Error: An error occurred\n
Error: An error occurred
```
Apparently the hook is established after 'Installing basic packages'
However it holds the second scenario when distrobox update hits suite change.

89luca89:
What type of image are you using? It seems to me strange that it wants to move from stable->oldstable during the init (where an image should be fresh)

aquaherd:
The image is frozen in around 2019 from then stable debian buster (10) and used as-is for ephemeral build-only containers daily.
The project to be built is roughly 10 years old and will be maintained for another 10 or 20, if possible with no changes to the tool-chain, compiler, make etc. It barely builds under bullseye (11) and worse under bookworm (12) and needs lots of patches to the host so it is feasible to have its pet container today.

In another project from 2014 I even have a debian jessie (8), but I was smart to point the container images' sources list to snapshot.debian.org so that this problem does not manifest. If you can't accept my patch I would modify the above image, too.

I don't know if my use case is too esoteric but imagine the benefit distrobox could provide to the maintainability of critical infrastructure projects - they also have maintenance cycles in the decades and may want to substitute some pet VMs with the much smaller distroboxes and maybe they will run into the same problem as me.

89luca89:
@aquaherd 

I get it now, but as you can see from the CI, this flag is not compatible with older versions of debian: 
https://github.com/89luca89/distrobox/actions/runs/5847366923/job/15853648563?pr=898

image

So I'm unsure this is a PR I can merge

Can you see if in your use-case, using a `pre-init-hook` works? Like:

`distrobox-create --image your:image --pre-init-hooks "apt-get update --allow-releaseinfo-change" [....]`

This would do the update first thing before the package setup, and you would be able to fix it before the package management.
This would easily be added to your `distrobox.ini` if you manage the box declaratively, or even in your ~/.distroboxrc if you
want to add it to every distrobox you create

89luca89:
@aquaherd seems like the "-y" works for every release
Does it resolve the releaseinfo thing also?

aquaherd:
the -y does not work but adding the option as -o did.

89luca89:
Shouldn't the -o flag be the same as putting the config in the apt.conf file?

aquaherd:
Yes. I will update it accordingly and test. Unfortunately I am on vacation for the next 2-3 weeks. If its OK by you, keep this open until I am back.

89luca89:
> Yes. I will update it accordingly and test. Unfortunately I am on vacation for the next 2-3 weeks. If its OK by you, keep this open until I am back.

No problem at all I'm on vacation too :smile:

89luca89:
Thanks @aquaherd reviewed this and seems all good to me :)

Thanks a lot, merging!

aquaherd:
Thank you, too. :)

PR #900, opened 2023-08-08
opensuse: init: handle untrusted repositories when installing basic packages

g7:
If the image ships some untrusted repositories initial package installation might fail for these two separate reasons

1) The package list obtained with `zypper se` is polluted with warnings coming
   from zypper itself about those untrusted repositories
2) Even with a correct package list, zypper would still return a non-zero exit
   code to acknowledge those untrusted repositories

This commit tries to tackle both points. The reason of checking for the exit code rather than using `--no-gpg-checks` is that this way untrusted packages wouldn't get in without user intervention.

### DIFF
--- a/distrobox-init
+++ b/distrobox-init
@@ -1066,8 +1066,9 @@ EOF
 			libvulkan_intel
 			libvulkan_radeon
 		"
+		# Mark gpg errors (exit code 106) as non-fatal, but don't pull anything from unverified repos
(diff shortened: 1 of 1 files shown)


89luca89:
Hi @g7 there are some linting comments to fix :+1:

g7:
Hi, thanks! It should be good now :)

89luca89:
Thanks a lot @g7 for the contribution :smile:

Issue #482, opened 2022-10-17
Make exported binaries runnable within other Distrobox containers


jgierer12:
**Is your feature request related to a problem? Please describe.**
When exporting a binary from `distrobox-foo` by running
```bash
distrobox-export --bin /usr/bin/mybin --export-path ~/.local/bin
```
 it will only be runnable from within `distrobox-foo` and the host. If I try to run `mybin` within a different Distrobox container `distrobox-bar`, it will throw an error:

```
/usr/bin/mybin: No such file or directory
```

I would like exported binaries to be runnable from both the host as well as any other Distrobox container.

**Describe the solution you'd like**
Changing the generated `~/.local/bin/mybin` to something like this would work:

```diff
  #!/bin/sh
  # distrobox_binary
  # name: distrobox-foo
  if [ ! -f /run/.containerenv ] && [ ! -f /.dockerenv ]; then
    /usr/bin/distrobox-enter  -n distrobox-foo --  /usr/bin/mybin  "$@"
+ elif command -v distrobox-host-exec > /dev/null; then
+   distrobox-host-exec --yes /usr/bin/distrobox-enter  -n distrobox-foo --  /usr/bin/mybin  "$@"
  else
    /usr/bin/mybin "$@"
  fi
```

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

89luca89:
oh I see what you mean, makes sense :+1: 
I'll check a way to implement this properly thanks for the suggestion :+1:

mcepl:
I had to patch exported scripts as well. I have it more simple:

```sh
#!/bin/sh
# distrobox_binary
# name: vis
if [ ! -f /run/.containerenv ] && [ ! -f /.dockerenv ]; then
        command="/usr/bin/distrobox-enter  -n vis --  /usr/bin/vis  "

        for arg in "$@"; do
                if echo "${arg}" | grep -Eq "'|\""; then
                        command="${command} \
                                $(echo "${arg}" | sed 's|\\|\\\\|g' |
                                sed 's| |\\ |g' |
                                sed 's|\$|\\\$|g' |
                                sed "s|'|\\\'|g" |
                                sed 's|"|\\\"|g')"
                elif echo "${arg}" | grep -q "'"; then
                        command="${command} \"${arg}\""
                else
                        command="${command} '${arg}'"
                fi
        done
        eval ${command}
elif [ -x "/usr/bin/vis" ] ; then
        /usr/bin/vis "$@"
else
[...]

PR #906, opened 2023-08-12
Make exported binaries runnable within other Distrobox containers

FrauH0lle:
Solves #482

- test if we are not in a container and run distrobox-enter if not

- test if we are in a container and if the container name is not the same as the one we are exporting from

- else run the binary


89luca89:
Thanks a lot @FrauH0lle for the contribution :smile:

Issue #688, opened 2023-04-04
[Discussion] Consider removing `--mount type=devpts,destination=/dev/pts`


alexpdp7:
When using distrobox with Podman nested inside an LXC container, `distrobox enter` does not work, although in general `podman` works.

If I patch distrobox not to use `--mount type=devpts,destination=/dev/pts`, then it seems to work OK.

I traced the introduction of this option to toolbx, which introduced this in https://github.com/containers/toolbox/issues/568#issuecomment-708595672 . However, it seems that this change does not address the underlying problem- the issue is still open.

Given that distrobox supports Docker, which does not even support this option... perhaps we could remove it? It doesn't seem to serve its original purpose, and it causes issues in some (admittedly "unsupported"- and I think it's fine not to support nesting in LXC) environments.

(I suspect there might be an underlying "bug" in podman/crun that's triggered by using this flag, but it's just a hunch.)

alexpdp7:
This seems to be related to https://github.com/89luca89/distrobox/pull/671 ? The actual error I get is this https://github.com/alexpdp7/alexpdp7/issues/10#issuecomment-1493430815 , which looks somewhat similar.

There seems to be something interesting lurking around, but I don't know enough to pin it down :(

ericcurtin:
Hit the same issue, trying to run CentOS Stream 9 QT-based UI apps using distrobox + podman + crun on debian bookworm lxc environment on a aarch64 Chromebook running Chrome OS (hope that makes sense), but I ran into many issues to be honest, this is just one.

89luca89:
As of commit f1913c2 the init is now doing the devpts internally
can anyone test if this still creates problems for LXC?

ericcurtin:
Different bug now, on latest version of git main branch:

```
$ distrobox enter -r fedora-38
Container fedora-38 is not running.
Starting container fedora-38
run this command to follow along:

 sudo podman logs -f fedora-38

 Error: could not start entrypoint.
You must run  entrypoint inside a container!
distrobox-init should only be used as an entrypoint for a distrobox!

This is not intended to be used manually, but instead used by distrobox-enter
to set up the container's entrypoint.
Error: An error occurred
```

ericcurtin:
This would be a significant feature for ChromeOS if we could fix this, it would mean you could run almost any Linux distro's UI app on ChromeOS

89luca89:
How is your setup?

Lxc with inside rootless podman?

ericcurtin:
[...]

PR #916, opened 2023-08-15
Add extra check to detect if we are in a container

ericcurtin:
In a podman container, if the container environment variable is set, it
means you are in a container.

--mount type=tmpfs,destination=/run only in initful docker

It's required for systemd on docker, on podman it's taken care of
--systemd=always flag.

### DIFF
--- a/distrobox-create
+++ b/distrobox-create
@@ -547,7 +547,7 @@ generate_command() {
 		--label \"manager=distrobox\"
 		--env \"SHELL=$(basename "${SHELL:-"/bin/bash"}")\"
 		--env \"HOME=${container_user_home}\"
-		--mount type=tmpfs,destination=/run
+		--env \"container=${container_manager}\"
 		--volume /:/run/host:rslave
 		--volume /dev:/dev:rslave
 		--volume /sys:/sys:rslave
@@ -556,6 +556,15 @@ generate_command() {
 		--volume \"${distrobox_export_path}\":/usr/bin/distrobox-export:ro
 		--volume \"${distrobox_hostexec_path}\":/usr/bin/distrobox-host-exec:ro
 		--volume \"${container_user_home}\":\"${container_user_home}\":rslave"
+
+	if [ "${init}" -eq 1 ] && echo "${container_manager}" | grep -q "docker"; then
+		result_command="${result_command}
+			--cgroupns host
+			--stop-signal SIGRTMIN+3
+			--mount type=tmpfs,destination=/run
+			--mount type=tmpfs,destination=/run/lock
+			--mount type=tmpfs,destination=/var/lib/journal"
+	fi
 
 	# This fix is needed so that the container can have a separate devpts instance
 	# inside
--- a/distrobox-init
+++ b/distrobox-init
@@ -169,7 +169,7 @@ while :; do
 done
 
 # Check we're running inside a container and not on the host
-if [ ! -f /run/.containerenv ] && [ ! -f /.dockerenv ]; then
+if [ ! -f /run/.containerenv ] && [ ! -f /.dockerenv ] && [ -z "${container}" ]; then
 	printf >&2 "You must run %s inside a container!\n" " $(basename "$0")"
 	printf >&2 "distrobox-init should only be used as an entrypoint for a distrobox!\n\n"
 	printf >&2 "This is not intended to be used manually, but instead used by distrobox-enter\n"
--- a/distrobox-export
+++ b/distrobox-export
@@ -167,7 +167,7 @@ if [ "${verbose}" -ne 0 ]; then
 fi
 
 # Check we're running inside a container and not on the host.
-if [ ! -f /run/.containerenv ] && [ ! -f /.dockerenv ]; then
+if [ ! -f /run/.containerenv ] && [ ! -f /.dockerenv ] && [ -z "${container}" ]; then
 	printf >&2 "You must run %s inside a container!\n" " $(basename "$0")"
 	exit 126
 fi


ericcurtin:
I saw @rhatdan in this issue so I trust it:

https://github.com/containers/podman/issues/3586

this is tested and resolves issue:

https://github.com/89luca89/distrobox/issues/688

I still get this crun error, but it enters the container:

```
$ distrobox enter -r fedora-38
Container fedora-38 is not running.
Starting container fedora-38

 sudo podman logs -f fedora-38

 Starting container...                           [ OK ]
 Installing basic packages...                    [ OK ]
 Setting up read-only mounts...                  [ OK ]
 Setting up read-write mounts...                 [ OK ]
 Setting up host's sockets integration...        [ OK ]
 Integrating host's themes, icons, fonts...      [ OK ]
 Setting up package manager exceptions...        [ OK ]
 Setting up rpm exceptions...                    [ OK ]
 Setting up sudo...                              [ OK ]
 Setting up groups...                            [ OK ]
 Setting up package manager exceptions...        [ OK ]
 Setting up rpm exceptions...                    [ OK ]
 Setting up sudo...                              [ OK ]
 Setting up groups...                            [ OK ]
 Setting up users...                             [ OK ]
 Executing init hooks...                         [ OK ]

Container Setup Complete!
Error: crun: open /dev/pts/4: No such file or directory: OCI runtime attempted to invoke a command that was not found
```

distrobox-host-exec
```diff
@@ -117,7 +117,7 @@ if [ "${verbose}" -ne 0 ]; then
 fi
 
 # Check we're running inside a container and not on the host
-if [ -z "${CONTAINER_ID:-}" ]; then
```
ericcurtin, line 120: CONTAINER_ID is docker-specific it seems so wouldn't work in podman, although we could add a `&& [ -z "${CONTAINER_ID:-}" ]` to the overall check if we wanted

alexpdp7:
I tested 9fcd76782e21f7af46c0d02f019ae41fd98d2bbb and it worked in LXC/Podman. Happy to retest as needed.

ericcurtin:
Linter errors seem to be from 24ff0c0c

alexpdp7:
Yep, latest commit also working for me. Thanks @ericcurtin !

distrobox-host-exec
89luca89: I think we can just check the 2 files and CONTAINER_ID

CONTAINER_ID is not a docker-specific thing, but a distrobox specific thing that we add in enter here:
https://github.com/89luca89/distrobox/blob/065a0965076b1bfcea7b952c255d05f041c9a7a3/distrobox-enter#L318

We could move this `--env` from enter to create, and we should be able to detect this :+1:

ericcurtin: Neither of these three things can be guaranteed in podman, that's why I do these three and the $container env var

89luca89:
@ericcurtin thanks for the PR :smile: 

Yes this problem has been introduced by 1894e4f3c7f58c9fe52e649d9c792c72b612f909

One option would be to either add the `CONTAINER_ID` env at creation time, and check for /.dockerenv && /run/.containerenv && CONTAINER_ID

Or (and probably the best thing) is to restrict the tmpfs mount to initful docker:

```sh
if [ "${init}" -eq 1 ] && echo "${container_manager}" | grep -q "docker"; then 
    --mount type=tmpfs,destination=/run"
fi
```

This should keep the current tests working (/.dockerenv and /run/.containerenv) as the tmpfs mount is performed exclusively on docker (and docker uses /.dockerenv so no override)

ericcurtin:
Will we still work with initful podman (as well as initful docker) after this change?

initful podman is getting popular, we are using it in my team at work.

ericcurtin:
Oh wait you explained never mind

ericcurtin:
Added extra commit with docker specific tmpfs thing

89luca89:
> Will we still work with initful podman (as well as initful docker) after this change?
> 
> initful podman is getting popular, we are using it in my team at work.

Yea using a lot myself too

I think we might reproduce what podman does without the flag (so that we do that also on docker)

From the podman's docs, it seems like:

```
  * Podman mounts tmpfs file systems on the following directories                 
    * _/run_                                                                      
    * _/run/lock_                                                                 
    * _/tmp_                                                                      
    * _/sys/fs/cgroup/systemd_                                                    
    * _/var/lib/journal_                                                          
  * Podman sets the default stop signal to **SIGRTMIN+3**.                        
  * Podman sets **container_uuid** environment variable in the container to the   
  first 32 characters of the container ID.                                        
  * Podman does not mount virtual consoles (_/dev/tty\d+_) when running with **--privileged**.
```

The /tmp, /var/lib/journal and /sys/fs/cgroup/systemd can be handled by using `--cgroupns host`

distrobox-host-exec
89luca89: Both here (host-exec) and in export, the CONTAINER_ID is guaranteed by the `distrobox-enter` that injects this variable in the env

We must pass trough an enter so it's safe to assume that CONTAINER_ID is present

distrobox-create
```diff
@@ -547,7 +547,6 @@ generate_command() {
 		--label \"manager=distrobox\"
 		--env \"SHELL=$(basename "${SHELL:-"/bin/bash"}")\"
 		--env \"HOME=${container_user_home}\"
-		--mount type=tmpfs,destination=/run
```
89luca89, line 550: Here I'd add `--env \"container=${container_manager}\"`
So that also on docker we will have the `container=docker` env variable like on podman we have `container=podman`

distrobox-init
```diff
@@ -169,7 +169,7 @@ while :; do
 done
 
 # Check we're running inside a container and not on the host
-if [ ! -f /run/.containerenv ] && [ ! -f /.dockerenv ]; then
+if [ ! -f /run/.containerenv ] && [ ! -f /.dockerenv ] && [ -z "${container}" ] && [ -z "${CONTAINER_ID:-}" ]; then
```
89luca89, line 172: Here we can just check $container  (and the files for backward compatibility)
As we inject `container` at creation (see above) 

CONTAINER_ID will not be present here in the init as it's injected at `enter` time

89luca89 (changes requested):

ericcurtin:
Made changes, squashed as it made the review comments easier to address. Tested using:

```
sudo podman rm -f centos-stream9 || true
distrobox create -r -i centos:stream9
distrobox enter -r centos-stream9
sudo dnf install -y epel-release
sudo dnf install -y kate # assuming mesa-dri-drivers got installed by distrobox
QT_QPA_PLATFORM=wayland kate
```

on ChromeOS host

89luca89:
Great stuff thanks a lot for your work @ericcurtin !! :smiley:

89luca89 (approved):

ericcurtin:
And we are green

PR #917, opened 2023-08-15
Add ChromeOS as supported host OS

ericcurtin:
This is paricularly useful in the case of aarch64 Chromebooks, as there
is a lack of well upstreamed reasonably priced aarch64 machines.
aarch64 Chromebooks fill this gap well. This is using the built in
Linux on ChromeOS mode which is debian-based.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -174,7 +174,7 @@ Fedora Silverblue for the [uBlue](https://github.com/ublue-os) project
 ## Why
 
 - Provide a mutable environment on an immutable OS, like [Endless OS,
-  Fedora Silverblue, OpenSUSE MicroOS](compatibility.md#host-distros)  or [SteamOS3](posts/install_rootless.md)
+  Fedora Silverblue, OpenSUSE MicroOS, ChromeOS](compatibility.md#host-distros)  or [SteamOS3](posts/install_rootless.md)
 - Provide a locally privileged environment for sudoless setups
   (eg. company-provided laptops, security reasons, etc...)
 - To mix and match a stable base system (eg. Debian Stable, Ubuntu LTS, RedHat)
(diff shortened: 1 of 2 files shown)


ericcurtin:
On merge of #916 ChromeOS is working and supported.

89luca89:
Hi @ericcurtin thanks!
I'm reviewing the other PR and then we can merge this :smile: 

Can we specify that this is inside the "linux mode" of Chromeos and not native?

ericcurtin:
Fedora Asahi Remix as a host OS also works since:

https://github.com/89luca89/distrobox/pull/702

was merged. But maybe that just falls under Fedora umbrella in general.

ericcurtin:
Added @89luca89

89luca89:
Thanks!

paulmenzel:
Do you have detailed install instructions?

ericcurtin:
At present it's just:

```
sudo apt install podman
curl -s https://raw.githubusercontent.com/89luca89/distrobox/main/install | sudo sh -s -- --next
```

in order to install from the HEAD of this git repo. It'll take a while to get to the various distro's packages.

In the ChromeOS Linux env of course:

https://support.google.com/chromebook/answer/9145439?hl=en

Feel free to reach out if you encounter any issues.

ericcurtin:
@paulmenzel you have to upgrade to Debian 12 also I forgot to say.

You have to do a text replace of "bullseye" to "bookworm" in various /etc/apt files and run "apt dist-upgrade" (maybe a reboot after to be extra safe) if memory serves me right. Google even maintain "bookworm" repos for their tools on this that work great, but they don't seem to have pushed that update to Chromebooks yet.

ericcurtin:
It would actually be really cool if Google changed this whole this ChromeOS thing from:

ChromeOS -> crosvm/kvm -> lxc/lxd -> Debian -> distrobox -> podman

to something like:

ChromeOS -> crosvm/kvm -> Debian -> distrobox -> podman

or it could even possibly be something like:

ChromeOS -> crosvm/kvm -> podman -> Debian -> distrobox -> podman

with the recent disagreement between Canonical and the ex-Canonical lead maintainers of lxc/lxd.

When lxc/lxd was chosen for ChromeOS, it had an advantage that lxc/lxd could run initful but other containers could not, this is no longer true in podman at least. Would also be one less tool, one less layer of indirection. And podman/distrobox are simpler to use.

PR #912, opened 2023-08-15
enter: fix `XDG_DATA_DIRS` and `XDG_CONFIG_DIRS` handling

ttimasdf:
Resolves #911 

when `XDG_DATA_DIRS` is empty

```bash
++ standard_paths='/usr/local/share /usr/share'
++ container_paths=
++ for standard_path in '$standard_paths'
++ pattern='(:|^)/usr/local/share(:|$)'
++ '[' -z '' ']'
++ container_paths=/usr/local/share
++ for standard_path in '$standard_paths'
++ pattern='(:|^)/usr/share(:|$)'
++ '[' -z /usr/local/share ']'
++ echo /usr/local/share
++ grep -Eq '(:|^)/usr/share(:|$)'
++ container_paths=/usr/local/share:/usr/share
++ set +x
```

when `XDG_DATA_DIRS` is not empty at the beginning

```bash
++ standard_paths='/usr/local/share /usr/share'
++ container_paths=/home/user/.local/share/flatpak/exports/share:/var/lib/flatpak/exports/share:/usr/local/share
++ for standard_path in '$standard_paths'
++ pattern='(:|^)/usr/local/share(:|$)'
++ '[' -z /home/user/.local/share/flatpak/exports/share:/var/lib/flatpak/exports/share:/usr/local/share ']'
++ echo /home/user/.local/share/flatpak/exports/share:/var/lib/flatpak/exports/share:/usr/local/share
++ grep -Eq '(:|^)/usr/local/share(:|$)'
++ for standard_path in '$standard_paths'
++ pattern='(:|^)/usr/share(:|$)'
++ '[' -z /home/user/.local/share/flatpak/exports/share:/var/lib/flatpak/exports/share:/usr/local/share ']'
++ echo /home/user/.local/share/flatpak/exports/share:/var/lib/flatpak/exports/share:/usr/local/share
++ grep -Eq '(:|^)/usr/share(:|$)'
++ container_paths=/home/user/.local/share/flatpak/exports/share:/var/lib/flatpak/exports/share:/usr/local/share:/usr/share
++ set +x
```

Issue #911, opened 2023-08-15
host provided XDG_DATA_DIRS not handled correctly


ttimasdf:
https://github.com/89luca89/distrobox/blob/cdaa8600569397907ebd5000988262b5d81d895c/distrobox-enter#L357-L367

When launching distrobox-enter with empty `XDG_DATA_DIRS`, standard paths are not added to environment variable.

```
++ standard_paths='/usr/local/share /usr/share'
++ container_paths=
++ for standard_path in '${standard_paths}'
++ '[' -n '' ']'
++ for standard_path in '${standard_paths}'
++ '[' -n '' ']'
++ result_command='docker exec
# skipped lines..
		--env "PATH=/usr/local/sbin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
		--env "XDG_DATA_DIRS="'
++ set +x
```

This happens when launching integrated VSCode app from GNOME 2 menu launcher.

environment: bash 4.2 on CentOS 7

```
sh-4.2$ sh --version
GNU bash, version 4.2.46(2)-release (x86_64-redhat-linux-gnu)
```

PR #918, opened 2023-08-16
enter: use natural order from host PATH

ttimasdf:
Similar problem to #911, that the way distrobox used now to append standard paths to container path is not working. This PR fix the problem with similar ways to #912 but changed the logic a little bit more. To preserve the order of both the host path and additional standard path instead of reversing them.

the `standard_paths="/usr/local/sbin /usr/local/bin /usr/sbin /usr/bin /sbin /bin"` is left untouched.

Examples:

host PATH: `/opt/mambaforge/bin:/opt/mambaforge/condabin:/opt/devkitpro/tools/bin:/home/vm/.cargo/bin:/usr/local/sbin:/usr/bin:/sbin:/bin:/opt/go/bin:/snap/bin:/home/vm/.local/bin`

output `container_paths=/opt/mambaforge/bin:/opt/mambaforge/condabin:/opt/devkitpro/tools/bin:/home/vm/.cargo/bin:/usr/local/sbin:/usr/bin:/sbin:/bin:/opt/go/bin:/snap/bin:/home/vm/.local/bin:/usr/local/bin:/usr/sbin`

### DIFF
--- a/distrobox-enter
+++ b/distrobox-enter
@@ -349,7 +349,7 @@ generate_command() {
 	done
 	# append additional standard paths to host PATH to get final container_paths
 	if [ -n "${container_paths}" ]; then
-		container_paths="${PATH}:${container_paths}"
+		container_paths="${container_paths}:${PATH}"
 	else
 		container_paths="${PATH}"
 	fi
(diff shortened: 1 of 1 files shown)


ttimasdf:
Or maybe we should prepend standard to the start instead of appending to the end?

89luca89:
Hi @ttimasdf 

I think prepending would be better

89luca89:
Thanks a lot @ttimasdf for the work! :smile:

PR #919, opened 2023-08-16
docs: ChromeOS compatibility should say Debian 12 bookworm

ericcurtin:
It works better on Debian 12 bookworm, this requires an upgrade, all
repos in the environment have 12 versions including the Google
maintained one for cros-packages. But there is an option to use docker
and Debian 11 bullseye. Original testing was done on podman with Debian
12, both rootful and rootless.

### DIFF
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -52,7 +52,7 @@ Distrobox has been successfully tested on:
 | Alpine Linux | | To setup rootless podman, look [HERE](https://wiki.alpinelinux.org/wiki/Podman) |
 | Arch Linux | | `distrobox` and `distrobox-git` are available in AUR (thanks [M0Rf30](https://github.com/M0Rf30)!). <br> To setup rootless podman, look [HERE](https://wiki.archlinux.org/title/Podman) |
 | CentOS | 8 <br> 8 Stream <br> 9 Stream | `distrobox` is available in epel repos. (thanks [alcir](https://github.com/alcir)!) |
-| ChromeOS | Debian 11 | This is using the built in Linux on ChromeOS mode which is debian-based. |
+| ChromeOS | Debian 11 (docker with make-shared workaround #non-shared-mounts) <br> Debian 12 (podman) | using built-in Linux on ChromeOS mode which is debian-based, which can be [upgraded](https://wiki.debian.org/DebianUpgrade) from 11 bullseye to 12 bookworm (in fact 12 is recommended) |
 | Debian | 11 <br> Testing <br> Unstable | `distrobox` is available in default repos in `testing` and `unstable` (thanks [michel-slm!](https://github.com/michel-slm!)!) |
 | deepin | 23 <br> Testing <br> Unstable | `distrobox` is available in default repos in `testing` and `unstable` |
 | EndlessOS | 4.0.0 | |


ericcurtin:
I actually always used Debian 12 for testing, I just said Debian 11 as I copied it from elsewhere in the matrix

docs/compatibility.md
```diff
@@ -52,7 +52,7 @@ Distrobox has been successfully tested on:
 | Alpine Linux | | To setup rootless podman, look [HERE](https://wiki.alpinelinux.org/wiki/Podman) |
 | Arch Linux | | `distrobox` and `distrobox-git` are available in AUR (thanks [M0Rf30](https://github.com/M0Rf30)!). <br> To setup rootless podman, look [HERE](https://wiki.archlinux.org/title/Podman) |
 | CentOS | 8 <br> 8 Stream <br> 9 Stream | `distrobox` is available in epel repos. (thanks [alcir](https://github.com/alcir)!) |
-| ChromeOS | Debian 11 | This is using the built in Linux on ChromeOS mode which is debian-based. |
+| ChromeOS | Debian 12 | This is using the built in Linux on ChromeOS mode which is debian-based (may require an upgrade of all repos to 12/bookworm tfirst). |
```
89luca89, line 55: We could add 2 entries here, one for deb12 (this one) and the other with deb11 stating that it works only with docker and the make-shared workaround, linking to The section below? 

https://github.com/89luca89/distrobox/blob/main/docs/compatibility.md#non-shared-mounts

89luca89 (changes requested):

docs/compatibility.md
ericcurtin: Tagging @maxamillion for review also. Maybe as a follow on commit you could open a PR documenting the upgrade process, you've done it more frequently than me? Since it's basically a recursive sed (of only the apt sources files) and a `apt dist-upgrade` I think?

maxamillion: Ah! Sorry, I just noticed this PR. I created this other one. https://github.com/89luca89/distrobox/pull/919 

Debian 11 does work if you install docker and do the [non shared mounts](https://github.com/89luca89/distrobox/blob/main/docs/compatibility.md#non-shared-mounts) steps or install podman v4+ from the OBS repos.

Thoughts @ericcurtin ?

ericcurtin: I would have liked to poke people to upgrade to Debian 12, that way you have to worry about less legacy issues, older bugs in Debian 11, to worry about, but that works also I'm open minded... So your PR works for me too :smile:

ericcurtin: In #922 , we'd at least have to say Debian 12 upgrade is required first in the podman case, because the podman documentation just says do "sudo apt-get -y install podman" which isn't sufficient as you end up with old podman.

maxamillion: We could combine them!

Recommend upgrading to Debian 12 following the docs here https://wiki.debian.org/DebianUpgrade but offer the option of staying on Debian 11. 🤔

ericcurtin: That Debian doc is perfect! Thanks for finding that, want to merge my changes into yours? Or vice-versa, I don't mind either way.

maxamillion: Honestly, the more I think about it I think you're right and we should just say "Debian 12+ only" but I think it would be good to also provide the docker and podman install links. 😄

ericcurtin: I pushed a merging of sorts here anyway, with 11 and 12 options.

maxamillion: I'll close my PR, let's just go with Debian 12+

maxamillion (approved):
LGTM TY!

89luca89:
Thanks @ericcurtin and @maxamillion for the fix and review!

PR #925, opened 2023-08-18
fix code example

bronson:
As written previously, Markdown would join the two lines into a one-line paragraph and show an invalid command.


89luca89:
Thanks a lot!!

PR #934, opened 2023-08-25
assemble/enter: enter or create based on entry in manifest file

bobslept:
With this feature it would be possible to create a box based on an entry in the distrobox.ini manifest file. Instead of choosing the default `container_image_default` in noninteractive mode.

Problem:
In [Bluefin](https://github.com/ublue-os/bluefin) we ship a predefined `distrobox.ini` which contains an `ubuntu` and `fedora` entry. We map shortcurts to let the user easily start the distrobox with `distrobox enter ubuntu`. If the box does not exist, it will create a box with the `container_image_default` image. This results in a box with a fedora image where the user expects a ubuntu image. Reference to our [issue](https://github.com/ublue-os/bluefin/issues/383)

We could let the user assemble the manifest boxes before hitting those shortcuts, but this option let us do it both ways.

Changes:
An argument to `distrobox-assemble` in the form of `--name/-n` to specify an entry name to act on a single entry of the `distrobox.ini` manifest.

Configuration options to use that  while using `distrobox enter` in noninteractive mode.
```
non_interactive_assemble="1"
non_interactive_assemble_manifest="/etc/distrobox/distrobox.ini"
```

-- Bob

### DIFF
--- a/distrobox-enter
+++ b/distrobox-enter
@@ -50,8 +50,6 @@ container_manager_additional_flags=""
 container_name=""
 container_name_default="my-distrobox"
 non_interactive=0
-non_interactive_assemble=0
-non_interactive_assemble_manifest="/etc/distrobox/distrobox.ini"
 
 # Use cd + dirname + pwd so that we do not have relative paths in mount points
 # We're not using "realpath" here so that symlinks are not resolved this way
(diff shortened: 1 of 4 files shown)


distrobox-enter
```diff
@@ -50,6 +50,8 @@ container_manager_additional_flags=""
 container_name=""
 container_name_default="my-distrobox"
 non_interactive=0
+non_interactive_assemble=0
```
89luca89, line 53: I'm not sure I like this
Enter should do enter things, and at most the "if not exist create a default"

I'd prefer the enter to stick to how it is now and not be entwined with assemble

89luca89 (changes requested):

89luca89:
@bobslept thanks for the PR!

Just a comment about the `enter` thing, but I really like the assemble addition :+1:

distrobox-enter
bobslept: Okay that's understandable. Too bad. I will remove those changes.

89luca89 (approved):

89luca89:
Thanks a lot for the PR @bobslept :smile:

PR #961, opened 2023-09-10
Trim comments from ini file lines only when `#` is outside of quotes

### LINKED ISSUE #956: [Discussion] Is there a way to escape `#` in distrobox.ini?
Is there a way that I missed that allows for escaping the comment character (`#`) when used in the middle of a quoted string in a `distrobox.ini` file?

 I have this `distrobox.ini` file:
```.ini
[my_debian]
image=debian:12
additional_packages="chromium git python3-pip python3-setuptools python3-wheel xsltproc"
init_hooks="sudo python3 -m pip install --break-system-packages git+https://github.com/medic/pyxform.git@medic-conf-1.17#egg=pyxform-medic"
```

Running the `distrobox-assemble create` command in the directory with that file results in the following error:

```
$ distrobox-assemble create
/tmp/tmp.m2ufOShp04: line 3: unexpected EOF while looking for matching `"'
 - Creating my_debian... 
Creating 'my_debian' using image debian:12       [ OK ]
Distrobox 'my_debian' successfully created.
To enter, run:

distrobox enter my_debian
```

Obviously, removing the `#` from my `init_hooks` makes the error go away (but breaks the command).  I tried escaping it using `\#`, but that did not resolve the issue.

jkuester:
Updates the `distrobox-assemble` logic for reading the INI file so `#` characters within quotes are not considered the start of a comment.  Given the following lines in the INI file:

```ini
init_hooks="echo 'Hello #World'"
init_hooks="echo 'Hello'" #World
```

With the current implementation, the first line will cause the following error when running `distrobox-assemble`:

```
line 1: unexpected EOF while looking for matching `"'
```

The _desired_ behavior is that these lines should cause the following to be printed in the container (and no errors when assembling):
```
Hello #World
Hello
```

This behavior is achieved by the code change in this PR!

Closes #956

### DIFF
--- a/distrobox-assemble
+++ b/distrobox-assemble
@@ -360,6 +360,35 @@ run_distrobox() (
 	fi
 )
 
+# Remove comments from ini file line. Respects quoted instances of `#`.
+# Arguments:
+#   a value string
+# Outputs:
+#   a value string with the comments removed
+remove_comments() {
+  line="${1}"
+  filtered_line=$(echo "${line}" | awk '
+    # Configure awk to treat each character as a separate field
+    BEGIN { FS="" }
+    {
+      # Loop through each character of the line
+      for(i = 1; i <= NF; i++){
+        # If character is a quote, toggle the quoting flag
+        if ($i == "\"") {
+          q = !q
+        }
+        # If character is a hash and we are outside of quotes, stop processing
+        if (!q && $i == "#") {
+          exit
+        }
+        # Print the current character
+        printf("%s", $i)
+      }
+    }
+  ')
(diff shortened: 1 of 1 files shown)


89luca89:
Hi @jkuester thanks for the PR!

I'm not convinced about the use of an external dependency `awk`
I'd prefer to stick to the basic stuff offered in the most minimal coreutils (eg: busybox)

89luca89:
Thanks @jkuester for the effort in this
I've pushed a solution I like more (using sed) and solved the issue

Thanks anyway for the interest and the effort!

PR #1001, opened 2023-10-01
distrobox-assemble fix missing ¤ in volume handling

samuelba:
With this '¤' volumes can be on multiple lines

```
volume="/abc:/abc "
volume="/def:/def"
```

basically the same as for `additional_packages` and `additional_flags`.


89luca89:
Thanks :)

PR #1005, opened 2023-10-03
init: debian: fix installation of deprecated libgl1-mesa-glx and libegl1-mesa

### LINKED ISSUE #995: [Error] Kali Distrobox does not allow to enter due to missing apt package in their repositories
**Describe the bug**

When entering a Kali container, this rolling release based on Debian testing¹ may not bring all packages that are going to be installed during `distrobox-init` and fails to start.

Here it is a Desktop dependency on `libgl1-mesa-glx`, but could as well be any other.

**To Reproduce**

```
distrobox create --image docker.io/kalilinux/kali-rolling:latest --name kali
distrobox enter kali
```

    Installing basic packages...           	Error: An error occurred

**Expected behavior**

A shell is opened in a Kali environment.

**Logs**

```
Reading package lists...
Building dependency tree...
Reading state information...
Package libgl1-mesa-glx is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

E: Package 'libgl1-mesa-glx' has no installation candidate
+ [ 100 -ne 0 ]
Error: An error occurred
+ printf Error: An error occurred\n
```

 - Which version of podman? 4.6.2
 - Which version of distrobox? 1.5.0.2
 - Which host distribution? Fedora 38
 - How did you install distrobox? `dnf install distrobox`

**Additional context**

The issue is also described in:

¹ https://www.reddit.com/r/linux/comments/16u73fx/dependency_issue_libgl1mesaglx/

---

Asides from the fact that this could have happened to any package in the dependencies list:

Maybe this occurrence is an invitation to consider CLI-only distroboxes, which explicitly do not bring the graphic/desktop related dependencies?

`distrobox create --console`, for example, could use a more minimal `distrobox-init` with less package dependencies on graphics support (Mesa, Vulkan, NVidia), which also decreases first-run boot up time.

g7:
Fixes #995

Since at least Debian Buster (2019), libgl1-mesa-glx and libegl1-mesa are transitional packages for libgl1+libglx-mesa0 and libegl1 respectively.

In Debian Trixie, the transitional packages have been removed. Install the actual packages directly.

If you think that it would be better to add a check for even older releases I could add that as well (so that libgl1-mesa-glx and libegl1-mesa are installed on ancient releases), but I'm not sure if anyone has such an usecase :thinking:

### DIFF
--- a/distrobox-init
+++ b/distrobox-init
@@ -570,8 +570,9 @@ if [ "${upgrade}" -ne 0 ] ||
 			xauth
 			xz-utils
 			zip
-			libegl1-mesa
-			libgl1-mesa-glx
+			libgl1
+			libegl1
+			libglx-mesa0
 			libvulkan1
 			mesa-vulkan-drivers
 		"


agherzan:
This is a bit obvious but worth noting - it also affects Ubuntu (e.g. 23.10).

89luca89:
Thanks a lot!
I'll add also back the version for older releases, as now thanks to #1005 it will be more robust in detecting them

PR #1011, opened 2023-10-09
Update compatibility.md for openSUSE

vyskocilm:
There's a special distrobox-compatible flavor of openSUSE, see https://en.opensuse.org/Distrobox. This is configured as a default image when using distribution provided distrobox packages.


89luca89:
Thanks :)

PR #1036, opened 2023-11-05
Use "policy" command as a more reliable way to detect missing packages

guss77:
The output from `apt-cache policy` is much more strict and limited in its variability and content and meant for machine parsing, so it should be preferred over `apt-cache show` when trying to detect the availability of a package.

This change fixes issue #995 . Though that issue is already fixed in `main`, I believe the current fix - that greps the output of `apt-cache show` is not reliable enough and could get us in trouble in the future.


almereyda:
This will be good to consider in the next release, which has already grown suitably, with activity on `main` declining in the last month, inviting for cutting a new version soon:

- https://github.com/89luca89/distrobox/compare/1.5.0.2...main
- https://github.com/89luca89/distrobox/compare/1.5.0.2...bf65e627b7b5c9b505932a671618c378aa84c48f as of today

89luca89:
Thanks! :)

PR #1040, opened 2023-11-06
export: fix hard path icon export

lribic:
icon_name would always have a trailing '@', resulting in the hard path check to always fail.

Also moving the hard path check into the for loop for multi icon support.


89luca89:
Thanks! :)

PR #1041, opened 2023-11-07
init: fix mount flags for files (e.g., `/etc/localtime`)

### LINKED ISSUE #1029: [Error] `/etc/localtime` failed to bind mount during distrobox-init
**Describe the bug**
The command `distrobx enter ...` failed due to bind mount failure of `/etc/localtime`. The backing filesystem is ZFS (2.2.0) with native overlay enabled.

**To Reproduce**
- Create a new distrobox: `distrobox create --name ubuntu-22.04 --image ubuntu:22.04 --home /path/to/ubuntu-home`
- Enter: `distrobox enter ubuntu-22.04`

**Expected behavior**
We should expect the commands finish without error, initialize all the mount points successfully, and enter the distrobox/container.

**Logs**
- Log of `distrobox enter ubuntu-22.04`: [distrobox-enter.log](https://github.com/89luca89/distrobox/files/13175765/distrobox-enter.log)
- Log from `podman`: [podman.log](https://github.com/89luca89/distrobox/files/13175825/podman.log)

   - **podman**
 - Which version or podman, docker or lilipod?
   - **podman 4.7.1**
 - Which version of distrobox?
   - **distrobox 1.5.0.2**
 - Which host distribution?
   - **Arch Linux**
   - **From Arch's repo (extra)**

**Additional context**
This issue seems to happen only when the backing filesystem is ZFS. I did not encounter such error when using xfs or btrfs.

As a workaround, I moved `/etc/localtime` from `HOST_MOUNTS_RO` to `HOST_MOUNTS`, which I think avoids being (bind) mounted as read-only. I'm not sure if this is a valid fix, but it shouldn't break any permission work assuming rootless podman/docker. https://github.com/89luca89/distrobox/blob/9309e1ebe5c70fd1f9b5286fd3a76f3e7bd995bd/distrobox-init#L1394-L1441
And it seemed working (at least the commands finished without errors/warnings and I can continue working inside container).

Let me know if you need any further testing.

weiren2:
Fixes #1029.

Mount points may have additional flags (e.g., `nodev`, `nosuid`) to check for. However, the function [`get_locked_mount_flags`](https://github.com/89luca89/distrobox/blob/bf65e627b7b5c9b505932a671618c378aa84c48f/distrobox-init#L211) exits immediately and returns an empty flag if the mount point is a file (e.g., `/etc/localtime`). If the original mount point is a file and it has any ["locked" flags](https://github.com/89luca89/distrobox/blob/bf65e627b7b5c9b505932a671618c378aa84c48f/distrobox-init#L239), then the file will fail to bind mount as read-only without those "locked" flags (in rootless mode).

I don't think it is necessary to exclude files when checking the mount flags. This PR removed that checking and updated the variable names to reflect the change.

### DIFF
--- a/distrobox-init
+++ b/distrobox-init
@@ -205,35 +205,31 @@ fi
 
 # Print mount flags considered "locked".
 # Arguments:
-# 	source_dir
+# 	source (path to the file/directory)
 # Outputs:
 # 	Comma-separated list of locked mount flags
 get_locked_mount_flags() (
-	source_dir="$1"
-	prev_dir=""
+	source="$1"
+	prev=""
 	locked_flags=""
 
-	if [ ! -d "${source_dir}" ]; then
+	# If we can't read the file/directory, exit
+	if ! ls "${source}" 2> /dev/null > /dev/null; then
 		return 0
 	fi
 
-	# If we can't read the directory, exit
-	if ! ls "${source_dir}" 2> /dev/null > /dev/null; then
-		return 0
-	fi
-
-	# Get mount flags of given directory, using nearest mountpoint.
+	# Get mount flags of given file/directory, using nearest mountpoint.
 	# Earlier versions of findmnt did not check parents until it found a mountpoint,
 	# so we use a workaround with dirname.
 	while true; do
-		flags="$(findmnt --noheadings --output OPTIONS --target "${source_dir}" || :)"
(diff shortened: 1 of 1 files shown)


89luca89:
Thanks! :)

PR #1043, opened 2023-11-07
Sidesteping sssd non-standard USER env var with a id command.

boormat:
This is a possible way of bypassing **sssd** generating non standard USER environment variables by adding @domain.  

This is hopefully a more robust approach than trying to replace the illegal characters, which the Kerberos solution is doing.  Not tested on a host setup with Kerberos as yet.

Amazon Workspaces Ubuntu and Amazon Linux exhibit this behavior, where sssd is adding the Active Directory domain in the $USER name, e.g. user@domain.name.

### DIFF
--- a/distrobox-enter
+++ b/distrobox-enter
@@ -305,7 +305,7 @@ generate_command() {
 	result_command="${result_command}
 		--detach-keys=\"\""
 	result_command="${result_command}
-		--user=\"${USER}\""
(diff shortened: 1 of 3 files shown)


boormat:
Found some issues when this when testing when based on main branch, seemed OK when applied to v1.5.0.2

boormat:
Turns out already fixed in main branch.

Issue #1060, opened 2023-11-19
[Error] Firefox segfaults when using systemd/init


pavinjosdev:
**Describe the bug**
Firefox and Firefox based browsers like Mullvad Browser segfaults when creating container with systemd/init flag using Debian 12 toolbox container image (on LMDE 6/Debian 12 host). Issue persists even when using `--unshare-all` flag. No issues when creating distrobox without systemd/init.

**To Reproduce**
```
# Create distrobox with init
distrobox create --image quay.io/toolbx-images/debian-toolbox:12 --init --additional-packages 'systemd libpam-systemd firefox-esr' --name debs

# Enter distrobox
distrobox enter debs

# Open firefox in normal mode
firefox

# Or safe mode
firefox --safe-mode

# Visit youtube and play a video in a new tab
# If no issue, stop distrobox and enter it again.
# Then open youtube with multiple tabs.
# If still no issue, restart host and open youtube in distrobox
```

**Expected behavior**
Firefox works without segfaults and other errors.

**Distrobox Logs including firefox issue at the end**
[distrobox_debug.txt](https://github.com/89luca89/distrobox/files/13404838/distrobox_debug.txt)

**Podman Logs**
[podman_debug.txt](https://github.com/89luca89/distrobox/files/13404839/podman_debug.txt)

**Desktop (please complete the following information):**
 - Are you using podman, docker or lilipod? podman
 - Which version or podman, docker or lilipod? podman version 4.3.1
 - Which version of distrobox? distrobox: 1.6.0 stable
 - Which host distribution? LMDE 6 (Debian 12.1)
 - How did you install distrobox? from github curl/bash script using latest stable version

**Additional context**
I have been having this issue for a few weeks but not been able to figure out the cause. Issue reproduced when using Fedora 38/39 Silverblue host & container, Debian trixie/sid host & Debian 12 container using distrobox v1.5.0.2 dev and stable as well.

Issue happens when visiting complex websites like YouTube, other large/complicated programs such as Mullvad VPN GUI are also affected. Issue persists when running Firefox in both normal mode and safe mode (addons/extensions disabled)

Opening this issue here as I was not able to reproduce the issue on a basic podman container running on the same host with systemd init, X11, and Pulse audio passthrough.

Thanks for any help in troubleshooting this.

Updates:
Issue reproduced on distrobox stable v1.6.0 and with/without deleting firefox profile (/home/pavin/.mozilla)

89luca89:
[...]

PR #1069, opened 2023-11-22
Fix systemd/init container startup

pavinjosdev:
This PR fixes issue #1060, with Firefox and other complex programs segfaulting due to the following issues:

1. Entering distrobox prior to systemd completing bootup and user integrations.
2.  [pam_systemd](https://www.freedesktop.org/software/systemd/man/latest/pam_systemd.html) not creating new user session when opening default shell or running program as argument to distrobox-enter.

**Test results for hosts**
1. LMDE 6:
- Debian 12 container with init: ✅
- Debian 12 container without init: ✅
2. Fedora 39 Silverblue:
- Debian 12 container with init: ✅
- Debian 12 container without init: ✅
- Fedora 39 container with init: ✅
- Fedora 39 container without init: ✅
- OpenSuse tumbleweed container with init: ❌ 
- Arch Linux container with init: ❌

**Remarks**
1. OpenSuse tumbleweed: init support remains partially broken as `systemd-logind.service` initializes `seat0` but does not initialize a new session when using `su <user>` to login. Using `sudo --user <user> --login` does create a new session and fixes the issue. Cannot use `su` or `sudo` login shell by default as it erases existing environment variables making them not suitable for distrobox purposes.
2. Arch Linux: init support remains permanently broken as `systemd-logind.service` initializes `seat0` but does not initialize a new session when using `su` or `sudo` to login.

Overall, no side effects to existing or new container workflows.

### DIFF
--- a/distrobox-enter
+++ b/distrobox-enter
@@ -564,7 +564,8 @@ if [ "${container_status}" != "running" ]; then
 	printf >&2 "\nContainer Setup Complete!\n"
 	# check if container was created with init/systemd, so we can
 	# wait until the system has been booted up before entering
-	if ${container_manager} inspect "${container_name}" --format "{{.Args}}" | grep -q 'init 1' && ${container_manager} exec "${container_name}" sh -c 'command -v systemctl' > /dev/null 2>&1; then
+	if ${container_manager} inspect "${container_name}" --format "{{.Args}}" | grep -q 'init 1' && \
+	${container_manager} exec "${container_name}" sh -c 'command -v systemctl' > /dev/null 2>&1; then
 		printf >&2 "Waiting for systemd startup...\n"
 		WAIT_COUNTER=0
 		while true; do
--- a/distrobox-enter
+++ b/distrobox-enter
@@ -343,6 +343,7 @@ generate_command() {
 	# and export them to the container.
 	set +o xtrace
 	# disable logging for this snippet, or it will be too talkative.
+	ENV_LIST=""
 	for i in $(printenv | grep '=' | grep -Ev ' |"|`|\$' |
 		grep -Ev '^(HOST|HOSTNAME|HOME|PATH|PROFILEREAD|SHELL|XDG_SEAT|XDG_VTNR|XDG_.*_DIRS|^_)'); do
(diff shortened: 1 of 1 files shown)


pavinjosdev:
Please hold off on this PR.
Arch and Suse creates a new session when using `login` command. Just need to figure out how to execute a command after calling it.

uncomfyhalomacro:
i wish this will be merged since my dev container is broken

ZariTen:
Yeah me too, my distrobox arch container with Podman is broken since the latest version.

89luca89:
From my testing this also fixes #1078

89luca89:
Thanks a lot @pavinjosdev for the awesome work!

I've adjusted a couple of thing, but seems like now all is working as intended, thanks!

Merging! :rocket:

pavinjosdev:
@89luca89 Unfortunately the transfer of code to wait for systemd startup from `distrobox-enter` to `distrobox-init` causes systemd to not assign a session on the first login. This makes sense because we've already used `su` to enter the container prior to systemd startup! 💥

Subsequent logins create a session but I'm getting lots of errors with pipewire and thus losing the audio. I think this is broken in its current implementation.

I just added the waiting part back to `distrobox-enter` and everything works fine.
The implications go beyond just this though. Because as the PAM config was also moved to `distrobox-init`, distros without a pam config for `su` will not be assigned a systemd session on first login.

It looks like [this](https://github.com/89luca89/distrobox/pull/1069/commits/d1a50e6f961b5d752a62e031761dec6a1ddd90f1) needs to be reverted.

89luca89:
@pavinjosdev the latest commits I've pushed waits for login with `systemctl is-system-running` and only after that is done, it prints `container_setup_done`, so that the `enter` can then do its thing

pavinjosdev:
> @pavinjosdev the latest commits I've pushed waits for login with `systemctl is-system-running` and only after that is done, it prints `container_setup_done`, so that the `enter` can then do its thing

Ah okay, my bad. I missed distrobox-enter waiting for the `container_setup_done` string from distrobox-init. 🤦

PR #1074, opened 2023-11-24
fix: fixed assemble start_now option with root

diogo464:
when assembling	the following example file:
```
[fedora]
root=true
start_now=true
```
distrobox fails	to enter the container because it does not pass the --root flag. this commit fixes that.


89luca89:
Thanks a lot @diogo464 !

Issue #978, opened 2023-09-18
[Error] On NixOS, the /nix directory is still mounted into containers


A-Lovett:
Despite [this](https://github.com/89luca89/distrobox/commit/5dd7b00e3f095e1976c0e2aa4f972efdcd6f4f3a) change, multiple people have confirmed they still see the /nix directory mounted into distrobox-created containers. This is after installing distrobox from the current main branch. Thanks.

chewblacka:
I can confirm. `/nix` is still mounted into all containers. Running the latest main branch.

89luca89:
If that's really needed, then the Nix packager for distrobox can add this to the default config

```conf
container_additional_volumes="/nix:/nix"
```

In one of the supported configuration system paths:

```
	/usr/share/distrobox/distrobox.conf
	/usr/share/defaults/distrobox/distrobox.conf
	/usr/etc/distrobox/distrobox.conf
	/usr/local/share/distrobox/distrobox.conf
	/etc/distrobox/distrobox.conf
```

But this shouldn't be the case
The stuff inside the container should not be inherited by the container
If it is, it's because of ~/.profile or the shell rc file, which should be made then container-aware to add the Nix path only outside of the container

89luca89:
Also just for headup

If the /nix is still mounted after an update
It's because you'll need to recreate the container for the change to be effective

nashamri:
I can confirm that after the https://github.com/89luca89/distrobox/commit/5dd7b00e3f095e1976c0e2aa4f972efdcd6f4f3a commit I'm getting the same error @chewblacka's getting:
```
 Error: could not start entrypoint.
exec /usr/bin/entrypoint: no such file or directory
```
when using `distrobox` on NixOS.

But is this a distrobox issue or a NixOs packaging issue :thinking: ?

89luca89:
Thats a duplicate of #315 and a whole other thing

But i'm still not able to resolve that, and it is not a high prio for me now

PR #1080, opened 2023-11-27
NixOS: Allow package maintainers to ship a static config file

pschmitt:
On NixOS, in order for the package maintainers to ship a default config file (as suggested [here](https://github.com/89luca89/distrobox/issues/978#issuecomment-1724077978)), the list of config files needs to be adjusted. 

None of the current config file paths are available from the derivation.
This PR essentially adds a config file path relative to the currrent script.

Ref: https://github.com/NixOS/nixpkgs/pull/268800

### DIFF
--- a/distrobox-create
+++ b/distrobox-create
@@ -74,6 +74,7 @@ unshare_ipc=0
 unshare_netns=0
 unshare_process=0
 unshare_devsys=0
+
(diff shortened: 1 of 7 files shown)


89luca89:
Thanks @pschmitt for the PR!

PR #1082, opened 2023-11-27
Use $XDG_CACHE_HOME, keep cache files in $HOME/.cache/distrobox by default

bjornfor:
Two commits:
- Read $XDG_CACHE_HOME, fall back to $HOME/.cache
- Change default cache dir from ~/.cache to ~/.cache/distrobox

### DIFF
--- a/distrobox-create
+++ b/distrobox-create
@@ -94,6 +94,8 @@ distrobox_hostexec_path="$(cd "$(dirname "${0}")" && pwd)/distrobox-host-exec"
 verbose=0
 version="1.6.0.1"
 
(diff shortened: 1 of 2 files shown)


bjornfor:
Should I add `XDG_CACHE_HOME` to the `Optional env variables:` comment?

bjornfor:
> Should I add `XDG_CACHE_HOME` to the `Optional env variables:` comment?

It seems to only list `DBX_*` env vars. E.g. SUDO_USER and DOAS_USER aren't listed.

89luca89:
Thanks a lot @bjornfor that's very useful thanks!

Issue #1083, opened 2023-11-30
[Error] Splitting individual bins from `exported_bins`


felixhoffmnn:
**Describe the bug**

The docs suggest defining multiple binaries within one statement inside the assembly file (see [here](https://github.com/89luca89/distrobox/blob/main/docs/usage/distrobox-assemble.md))

When doing so, it appears that the string of binaries is not separated correctly. When only defining one binary per line, it works.

```ini
exported_bins="/usr/bin/nvim" # works
exported_bins="/usr/bin/tmux /usr/bin/pipx /usr/bin/poetry" # dosent work
```

**To Reproduce**

```ini
[test-distrobox]
image=registry.fedoraproject.org/fedora-toolbox:39
replace=true
pull=true
root=false
start_now=false
# Set up utility packages
additional_packages="python3.11 python3-pip pipx neovim tmux"
# Export apps and binaries to the host
exported_bins="/usr/bin/nvim"
exported_bins="/usr/bin/tmux /usr/bin/pipx /usr/bin/pip3"
```

**Expected behavior**

Based on the docs, I expect that the string gets split by spaces, resulting in distrobox exporting each binary individually.

**Logs**

[logs.txt](https://github.com/89luca89/distrobox/files/13518431/logs.txt)

**Desktop (please complete the following information):**
 - Podman version 4.7.2
 - Distrobox: 1.6.0.1
 - Fedora Linux 39.20231130.0 (Silverblue)
 - `dnf`

PR #1084, opened 2023-12-01
FIX: Split multiple apps and bins by space

felixhoffmnn:
## Description

> See #1083

Based on the docs, it should be possible to provide a list to `exported_bins`. This PR addresses this issue and extends the functionality also to the `exported_apps`.

**Note:** This is my first PR to this project. Please don't hesitate to test and modify it to your standards.

## Testing

By this fix, I was able to run the example shown in the referred issue.

### DIFF
--- a/distrobox-assemble
+++ b/distrobox-assemble
@@ -370,15 +370,27 @@ run_distrobox() (
 		"${distrobox_path}"/distrobox enter "${name}" -- touch /dev/null
 
 		IFS="¤"
-		for app in ${exported_apps}; do
-			app="$(echo "${app}" | tr -d '[:space:]')"
-			"${distrobox_path}"/distrobox enter "${name}" -- distrobox-export --app "${app}"
+		for apps in ${exported_apps}; do
+			# Split the string by spaces
+			IFS=" "
+			for app in ${apps}; do
+				# Remove spaces
+				app="$(echo "${app}" | tr -d '[:space:]')"
+				# Export the app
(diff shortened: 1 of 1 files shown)


felixhoffmnn:
I think it would also be good to add an example for exporting multiple apps.

distrobox-assemble
```diff
@@ -370,15 +370,27 @@ run_distrobox() (
...
 		IFS="¤"
-		for app in ${exported_apps}; do
-			app="$(echo "${app}" | tr -d '[:space:]')"
-			"${distrobox_path}"/distrobox enter "${name}" -- distrobox-export --app "${app}"
+		for apps in ${exported_apps}; do
+			# Split the string by spaces
+			IFS=" "
+			for app in ${apps}; do
+				# Remove spaces
+				app="$(echo "${app}" | tr -d '[:space:]')"
```
dkrulev-sdc, line 378: the script involves splitting a string by spaces and then trimming spaces from each resulting element. However, I'm curious if the trimming step is actually necessary.

I tried the following code in shell, results are identical
```sh
apps="1   2    3  4 5 6"
IFS=" "
echo with trimming
for app in ${apps}; do
        app="$(echo "${app}" | tr -d '[:space:]')"
        echo "|$app|"
done
echo without trimming
for app in ${apps}; do
        echo "|$app|"
done
```

felixhoffmnn: You're right. I removed the `tr -d '[:space:]')`.

89luca89:
Thanks a lot @felixhoffmnn this will Fix #1083

Issue #1097, opened 2023-12-11
[Error] xargs parsing error at exporting app 


lrdass:
**Describe the bug**

`distrobox-export --app emacs` isn´t working as expected. 
it throws the output:

`xargs: unmatched single quote; by default quotes are special to xargs unless you use the -0 option
Error: cannot find any desktop files.
Error: trying to export a non-installed application.`

Trying to debugging it by setting only the folder `/usr/share/applications` it works properly. 
So probably the bug is in one of the queries to find an `emacs` installation. 
```
 xargs -t -I{} grep -le Exec=.*emacs.* -le Name=.*emacs.* {}

 xargs -t -I{} grep -Le Exec=.*/home/deck/.local/bin/distrobox-enter.* {}

 xargs -t -I{} printf %s@ {}
```

**To Reproduce**
Install emacs `(ubuntu:latest)`:
`sudo apt-get install emacs`

and export it:
`distrobox-export --app emacs`

**Expected behavior**
Emacs exported to my local user.

**Logs**
(I've added the `-t` verbose tag to `xargs` to extend the grep command) 

```
+ [ ! -f /run/.containerenv ]
+ grep -q rootless=0 /run/.containerenv
+ id -u
+ [ 1000 -eq 0 ]
+ [ -z emacs ]
+ [ -n emacs ]
+ [ -n  ]
+ [ -n  ]
+ [ -n ubuntu ]
+ container_name=ubuntu
+ [ -z ubuntu ]
+ [ 0 -ne 0 ]
+ container_command_prefix=/home/deck/.local/bin/distrobox-enter  -n ubuntu -- /bin/sh -l -c 
+ [ -z  ]
+ exported_app_label= (on ubuntu)
+ export_application
+ canon_dirs=
+ IFS=:
+ [ -n /home/deck/.local/share/flatpak/exports/share:/var/lib/flatpak/exports/share:/usr/local/share:/usr/share ]
+ [ -d /home/deck/.local/share/applications ]
+ canon_dirs= /home/deck/.local/share/applications
+ [ -n /home/deck/.local/share ]
+ [ -d /home/deck/.local/share/flatpak/exports/share/applications ]
+ canon_dirs= /home/deck/.local/share/applications /home/deck/.local/share/flatpak/exports/share/applications
+ [ -d /var/lib/flatpak/exports/share/applications ]
+ [ -d /usr/local/share/applications ]
+ [ -d /usr/share/applications ]
[...]

PR #1098, opened 2023-12-13
Fix: xargs was breaking mid find search due to files with quotes on filename

lrdass:
**Issue**: https://github.com/89luca89/distrobox/issues/1097

**What it does:**
Now any file name searched by `find`, with scape characters `'` or `"` will print it's full name and passed down to `xargs`.

### DIFF
--- a/distrobox-export
+++ b/distrobox-export
@@ -348,10 +348,10 @@ export_application() {
 	# by its launcher name.
 	desktop_files=$(
 		# shellcheck disable=SC2086,SC2038
-		find ${canon_dirs} -type f -printf "%p\n" -o -type l -printf '"%p"\n' |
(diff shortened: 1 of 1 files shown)


yump:
Is there a reason for not using `find -print0 | xargs -0`?  That is the standard solution to this class of problem.

mrnerdhair:
I submitted #1102 with the `-print0` approach. For reference, I am a dummy and did not even realize this PR existed at the time, or I would have suggested changes here instead. Still, though, I think the `-print0` mechanism is the better solution to this problem.

lrdass:
Hey guys! I appreciate you guys input! if #1102 closes this, I can close this pr then. Thank you guys for the reply!

89luca89:
Thanks @lrdass 

Will continue with the #1102 PR then

PR #1102, opened 2023-12-18
export & rm: correctly handle paths containing whitespace and quotes

mrnerdhair:
~~In order to reliably handle paths with whitespace, quotes, or nonprintable characters, `find`, `xargs`, and `grep` should be configured to separate them using the NUL character rather than using newlines or ad-hoc escaping. (This is documented in the shellcheck linting rule SC2038, which documents a real issue and should not be disabled.)~~

Update:

In order to reliably process paths with whitespace or quotes to xargs without using non-POSIX features, each character must be escaped. (Quoting will handle whitespace, but not filenames with quotes in them.)

### DIFF
--- a/distrobox-export
+++ b/distrobox-export
@@ -389,11 +389,11 @@ export_application() {
 	# it is possible to export an app not only by its executable name but also
 	# by its launcher name.
 	desktop_files=$(
-		# shellcheck disable=SC2086,SC2038
-		find ${canon_dirs} -type f -printf "%p\n" -o -type l -printf '"%p"\n' |
-			xargs -I{} grep -le "Exec=.*${exported_app}.*" -le "Name=.*${exported_app}.*" "{}" |
-			xargs -I{} grep -Le "Exec=.*${DISTROBOX_ENTER_PATH:-"distrobox-enter"}.*" "{}" |
-			xargs -I{} printf "%s@" {}
+		# shellcheck disable=SC2086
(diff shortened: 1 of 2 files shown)


lrdass (approved):
closes #1097 
nice solution!

89luca89:
Hi @mrnerdhair the problem I see here is that `grep -Z` is not POSIX supported
So this would fail for example in an alpine container:

image

mrnerdhair:
Interesting; I'll see if I can't find a workaround.

mrnerdhair:
I made it work on Alpine by using `xargs -d "\n"` before realizing that in terms of strict POSIX compliance that's verboten as well. In fact, there's no way at all to use a different delimiter! That leaves quoting and escaping, and quoting is hard because you have to have special cases for quoting quotes. So escaping it is, and while `sed` is very happy to throw a backslash in front of literally every character for us it does hurt me a little inside.

Also this fancy new solution now can't handle filenames containing newline characters, so I guess I'll just label people who use those kind of filenames as social deviants who inherently deserve to be shunned. After all, the alternative is to question why the fundamental standards of my ~~society~~ operating system are so broken.

mrnerdhair:
Also found some more xargs commands with the same issues in distrobox-rm; same solution, so I threw in another commit to address those as well.

89luca89:
Thanks a lot @mrnerdhair ! :heart:

PR #1107, opened 2023-12-20
Add workaround for large groups

arjo129:
This PR adds a work around for large groups. Currently if groups exceed a certain size in go, the system gives up https://github.com/golang/go/blob/a2a2c5b947263ee9328674d229892841197a0a94/src/os/user/cgo_lookup_unix.go#L181 searching. When podman look up uids and gids it means that exceeding a threshold will trigger an error.

>> Error: unable to start container "980124...": creating temporary passwd file for container 980124...: failed to get current group: user: lookup groupid group_id: internal buffer exceeds 1048576 bytes

This PR works around the issue by providing uid and gid of the user that should be retained.

### DIFF
--- a/distrobox-create
+++ b/distrobox-create
@@ -783,7 +783,7 @@ generate_command() {
 		# Use keep-id only if going rootless.
 		if [ "${rootful}" -eq 0 ]; then
 			result_command="${result_command}
-				--userns keep-id"
+				--userns=keep-id:uid=$(id -u),gid=$(id -g)"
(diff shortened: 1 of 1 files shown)


89luca89:
Thanks @arjo129 

I think this would be better if we check for the size, and, if exceeded, we use this workaround, what do you think?

89luca89:
Closing due to inactivity, let me know if you want to reopen

PR #1114, opened 2023-12-25
fix: binary exports should handle arguments containing spaces

mrnerdhair:
Without these quotes, arguments (particularly paths) containing spaces are "splatted" by the shell, turning them into multiple arguments (and causing path-not-found issues in particular).


89luca89:
Thanks @mrnerdhair this was collateraly fixed in another preious PR

Thanks anyway for your work!

PR #1121, opened 2023-12-28
export: Allow overriding start shell

EyeCantCU:
Adds the --start-shell argument to distrobox-export, allowing users to override the start shell for applications/binaries


89luca89:
Thanks @EyeCantCU 
But as of now the export does not use anymore a start_shell and instead just uses the regular login

Thanks anyway for your work!

EyeCantCU:
> Thanks @EyeCantCU 
> But as of now the export does not use anymore a start_shell and instead just uses the regular login
> 
> Thanks anyway for your work!

No problem :). Happy to help out

Issue #1119, opened 2023-12-26
[Error] "Distrobox create" uses Fedora38 when on Rawhide


trytomakeyouprivate:
I am not sure if this is because of rawhide, or because the automatically chosen version is outdated.

```
distrobox create
Image registry.fedoraproject.org/fedora-toolbox:38 not found.
Do you want to pull the image now? [Y/n]: 
```

```
Operating System: Fedora Linux 40 (Kinoite-beta)
KDE Plasma Version: 5.90.0
KDE Frameworks Version: 5.246.0
Qt Version: 6.6.1
Kernel Version: 6.7.0-0.rc6.20231222git24e0d2e527a3.52.fc40.x86_64 (64-bit)
```

osalbahr:
FYI, this is because the default image is [hardcoded](https://github.com/89luca89/distrobox/blob/3435f4d27070a99668bfa29a3e508db4ecc09009/distrobox-create#L56C10-L56C10) as 38.

trytomakeyouprivate:
oh lol okay. I would like to discuss using the hosts Distro version is nothing is added?

https://github.com/89luca89/distrobox/pull/1156

That fixes the version for Fedora for now

PR #1134, opened 2024-01-02
Bump default container image to Fedora 39

osalbahr:
Best-effort, search-and-replace. Checked manually using:

    % rg -iC1 'fedora.*38'

Fedora 39 was released 2023-11-07 (about 2 months ago)

imho, 2 months is enough of a grace period.

Thoughts?

- Fixes 89luca89/distrobox#1119


89luca89:
Thanks a lot @osalbahr !

PR #1140, opened 2024-01-06
Updated URLS for Fedora Images, Removed F36

sfalken:
Updated the list of Fedora Images to properly reflect the location of Rawhide images, and Removed F36 as it's completely out of support


89luca89:
Thanks a lot @sfalken !

PR #1147, opened 2024-01-08
fix(assemble): Use read instead of cat

p-fruck:
Hey there,

not sure if this was intentional, but I would prefer to use `read` instead of `cat` since `cat` tends to break stuff from time to time (escaping)


89luca89:
Thanks @p-fruck !

PR #1169, opened 2024-01-24
Add --enter-flags to distrobox-export

FrauH0lle:
This PR allows the user to specify additional flags for `distrobox-enter` as described [here](https://github.com/89luca89/distrobox/blob/main/docs/usage/distrobox-enter.md) when using `distrobox-export`.

For example, I had the use case that I wanted the container not to run in the current working directory as this was a symlink from my root partition and thus not available inside the container. This would result in an error when starting the container.
Appending `--no-workdir` to the exported command was the solution.


89luca89:
Thanks @FrauH0lle ! Nice idea!

PR #1170, opened 2024-01-24
Capture full container command when exporting app

FrauH0lle:
When exporting RStudio from a Fedora container, the resulting .desktop file contains the line
```
Exec=/usr/bin/distrobox-enter  -n rstudio -- /bin/sh -l -c  env RSTUDIO_DISABLE_CHECK_FOR_UPDATES=1 rstudio  %F 
```

Executing the file results in only `env` being executed. The command should be wrapped in quotes like
```
Exec=/usr/bin/distrobox-enter  -n rstudio -- /bin/sh -l -c  'env RSTUDIO_DISABLE_CHECK_FOR_UPDATES=1 rstudio  %F' 
```

This PR fixes that.

### DIFF
--- a/distrobox-export
+++ b/distrobox-export
@@ -451,7 +451,7 @@ export_application() {
 		# Add closing quote
 		# If a TryExec is present, we have to fake it as it will not work
 		# through the container separation
-		sed "s|^Exec=|Exec=${container_command_prefix} |g" "${desktop_file}" |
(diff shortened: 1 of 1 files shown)


ebits21:
I've noticed this issue with Syncthing and Insync as well (been fixing manually)

I've also noticed extra spaces here and there.

89luca89:
Thanks a lot @FrauH0lle !

Issue #1164, opened 2024-01-19
[Error] Not able to change hostname in init_hooks


briandipalma:
Having a fake hostname causes weird failures when running Java/Hibernate code. I guess Hibernate tries to connect to a DB file using the hostname. The errors that are raised are ClassNotFound ones so it was painful to debug what was going on. Setting the hostname to my local machine one, the correct one, fixes the issue. I've been trying to add it to my init_hooks with no success so far.

```
init-hooks=hostname "$(uname -n)";
```

Causes this

```
brian@brian-pop-os-pc-specialist:~/dev/iac/images$ distrobox assemble create 
/usr/bin/distrobox-assemble: 4: /tmp/tmp.tzoNM46AWF: init-hooks=hostname brian-pop-os-pc-specialist;: not found
```

While

```
"$(uname -n)" > /etc/hostname
```

Causes

```
brian@brian-pop-os-pc-specialist:~/dev/iac/images$ distrobox assemble create 
/usr/bin/distrobox-assemble: 4: /tmp/tmp.7ILlQSqxEf: cannot create /etc/hostname: Permission denied
```

While

```
sudo "$(uname -n)" > /etc/hostname
```

Causes

```
Setting up groups...                    	 [ OK ]
Setting up users...                     	 [ OK ]
Setting up skel...                      	 [ OK ]
Executing init hooks...                 	 Error: An error occurred
```

It would be nice if we could create containers that used their host's hostname as the way I imagine most people use distrobox containers is as local host like environments. I doubt people are deploying them as server host environments.

Something like

`change_hostname=false`

in a distrobox.ini file would be fantastic.

I see the code in `generate_command`, would a PR be accepted?

michaelhaaf:
In the time since making that comment I've come to thinking/realizing a few things:

(1) It's probably that the result of this `init-hook` method creates a new FQDN for the intents and purposes of both resolving to the host name and being unique/individually identifiable on a local network (i.e. there's nothing wrong with `boxname.hostname`). 

It turns out there's already been quite a lot of discussion about this [in this repo](https://github.com/89luca89/distrobox/issues/62) and in [toolbox](https://github.com/containers/toolbox/issues/969) which I only skimmed, but the important parts seem to work (e.g. ssh -x forwarding) in this scheme. 
 
(2) The documentation has a typo, which is a separate issue that I also had run into when trying this out (which is why I knew the fix). I've made a pull request to address this [here](https://github.com/89luca89/distrobox/pull/1182).

[...]

PR #1182, opened 2024-02-01
docs: fix init-hook script for hostname matching

michaelhaaf:
The init-hook provided in the [useful tips documentations](https://github.com/89luca89/distrobox/blob/main/docs/useful_tips.md) for creating a distrobox with the the same hostname as its host has two typos:

- Command substitution wrapping `uname -n` suppressing stdout redirect
- trailing "`"

Using this init-hook as suggested causes build errors (see #1164). This fix corrects the suggested script by removing the command substitution and the trailing "`".

To elaborate: I think command substitution here is misplaced since `uname -n` already prints to stdout and can be redirected directly. Or you can use `echo` on the provided command substitution. Or you can use the `hostname` command. I.e.

`uname -n > /etc/hostname`
`echo "$(uname -n)" > /etc/hostname`
`hostname "$(uname -n)"`

Tested on Fedora 39 below:

```
# Also tested the 'echo "$(uname -n)" > /etc/hostname' with the same result

user@hostname $ distrobox create --image ghcr.io/username/sample-image:latest -n box-name \ 
    --init-hooks 'uname -n > /etc/hostname' && distrobox enter box-name

...

Setting up groups...                    	 [ OK ]
Setting up users...                     	 [ OK ]
Setting up skel...                      	 [ OK ]
Executing init hooks...                 	 [ OK ]

Container Setup Complete!

user@box-name.hostname $ cat /etc/hostname 
box-name.hostname

user@box-name.hostname $ hostname
box-name.hostname  
```

### DIFF
--- a/docs/useful_tips.md
+++ b/docs/useful_tips.md
@@ -222,7 +222,7 @@ creating it with the following init-hook:
 
 ```sh
 distrobox create --name test --image your-chosen-image:tag \
-                  --init-hooks '"$(uname -n)" > /etc/hostname'`
+                  --init-hooks 'echo "$(uname -n)" > /etc/hostname'
 ```
 
 This will ensure SSH X-Forwarding will work when SSH-ing inside the distrobox:


89luca89:
Hi @michaelhaaf 
You're right about the wrong command

But I think the correct one would be 

```
user@hostname $ distrobox create --image ghcr.io/username/sample-image:latest -n box-name \ 
    --init-hooks "echo $(uname -n) > /etc/hostname; hostname $(uname -n)" && distrobox enter box-name
```

Tested and it works

michaelhaaf:
The following are equivalent and all work:

`uname -n > /etc/hostname`
`echo "$(uname -n)" > /etc/hostname`
`uname -n > /etc/hostname; hostname "$(uname -n)"`
`echo "$(uname -n)" > /etc/hostname; hostname "$(uname -n)"`

Maybe there's room for preference; personally the first choice makes the most sense to me, the rest have superfluous commands. 

(to elaborate: The call to `hostname` seems redundant, beyond that it is not system-independent and does not make the change permanent, unlike overwriting `/etc/hostname`. I think the first choice is the most concise: why evaluate `uname -n` then `echo` that evaluation if you can just redirect `uname -n` stdout to begin with.)

89luca89:
The `uname -n > /etc/hostname` will result in the container injecting **its own** hostname in the file

The `echo $(uname -n) > /etc/hostname` works because the `$(uname -n)` is evaulated at `distrobox-create` time, so in the end the result is just one `echo foo > /etc/hostname` in the init_hook

michaelhaaf:
> The `echo $(uname -n) > /etc/hostname` works because the `$(uname -n)` is evaulated at `distrobox-create` time, so in the end the result is just one `echo foo > /etc/hostname` in the init_hook

It's unclear to me how that works (in my tests I got the same result using both).

But, in the interest of getting a working version in the documentation/removing the typo/moving on, I've updated my PR to use `echo "$(uname -n)" > /etc/hostname` instead.

89luca89:
Thanks a lot @michaelhaaf

robinhellmers:
Adding a little bit to this conversation as there was a typo in this change. Having the single quotation marks outside of the double quotation marks will prevent `$(uname -n)` from being evaluated during the `distrobox-create` command. That does, for me, result in `<container-name>.<hostname>`, meaning no change at all.

Changing
`--init-hooks 'echo "$(uname -n)" > /etc/hostname'`
to
`--init-hooks "echo '$(uname -n)' > /etc/hostname"`
solves that issue.

Other than that, the command
`echo '$(uname -n)' > /etc/hostname`
is not equivalent to
`echo '$(uname -n)' > /etc/hostname; hostname '$(uname -n)'`

as the first example requires a reboot for the `hostname` command to actually use the changed `/etc/hostname`. Meanwhile if you also execute `hostname "$(uname -n)"`, a reboot isn't required.

Issue #1115, opened 2023-12-26
[Error] Provided config.fish cause error


SiriosDev:
**Describe the bug**
I premise that I am Linux newbie so forgive if the terms are not correct

the configuration of fish (`/etc/fish/conf.d/config.fish`) provided by `distrobox`

to be precise fish (latest) doesn't accept the syntax `$(...)`, instead it uses `(...)` without `$`, in fact removing the dollar sign, make fish works correctly again
```fish
/etc/fish/conf.d/config.fish (line 1): $(...) is not supported. In fish, please use '(id)'.
test -z "$USER" && set -Ux USER "$(id -un 2> /dev/null)"
                                 ^
from sourcing file /etc/fish/conf.d/config.fish
	called on line 253 of file /usr/share/fish/config.fish
from sourcing file /usr/share/fish/config.fish
	called during startup
source: Error while reading file '/etc/fish/conf.d/config.fish'
Welcome to fish, the friendly interactive shell
Type help for instructions on how to use fish
```
Browsing the repo I found the possible "responsible" code: https://github.com/89luca89/distrobox/blob/3435f4d27070a99668bfa29a3e508db4ecc09009/distrobox-init#L1749-L1766

**To Reproduce**
 - Having `fish` installed on host
 - Create a container with `distrobox` (in my case Ubuntu 22.04)
 - Enter the container

**Expected behavior**
Fish running normally.

**Desktop (please complete the following information):**
 - Are you using podman, docker or lilipod?
     - Podman
 - Which version or podman, docker or lilipod?
     - 4.8.2
 - Which version of distrobox?
     - 1.6.0.1
 - Which host distribution?
     - Arch
 - How did you install distrobox?
     - Pacman

profetik-777:
@SiriosDev , thanks for submitting. Sometimes noobs just throw their hands up and move on - so appreciate the extra effort.

I'm not sure if non-bash shells are supported. So this technically, might not be a "bug" but an "Out of Scope" use case. 

@89luca89 does "distrobox" as  product care if it is ran through fish command shell? Or it has to be bash? 

My understanding - they diverge enough to discourage bash / fish swap outs - and it isn't assumed that users should expect them to behave the same.

SiriosDev:
Don't know if it is out of scope, however being in the distrobox init mentioned above I thought so

Yey007:
[...]

PR #1186, opened 2024-02-03
Fish fixes

yump:
Doing a git-diff of my `~/.config/fish/` (on the host) , I found some mysterious additions to fish_variables (`SHELL`, `UID`, `MARONNE`). Thanks to the unique string "MARONNE", I was able to track down the culprit: an `/etc/fish/conf.d/config.fish` written (seven times) by distrobox-init.

"MARONNE" seems most likely to be the result of a find/replace with clipboard contents? The command that sets it is otherwise the same as the previous line.

And then I searched for "fish" in the bug tracker to see if this had been reported already, and found #1115, so I fixed that too.

### DIFF
--- a/distrobox-init
+++ b/distrobox-init
@@ -1748,7 +1748,7 @@ EOF
 
 # It's also importanto to keep this working on fish shells
 if [ -e "/etc/fish/config.fish" ]; then
-	cat << EOF >> /etc/fish/conf.d/config.fish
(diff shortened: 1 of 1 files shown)


89luca89:
Thanks a lot @yump yes that was a debug string leftover :sweat_smile:  thanks for the PR!

PR #1192, opened 2024-02-05
feat(init): Implement support for Wolfi

EyeCantCU:
Install the Wolfi base package and libc utilities


EyeCantCU:
Depends on: https://github.com/89luca89/distrobox/pull/1191

89luca89:
Can we rework this so that it doesn't depend on #1191 ?

From what I understand, we just need that if/else on the base package right?

distrobox-init
```diff
@@ -415,64 +415,111 @@ if [ "${upgrade}" -ne 0 ] ||
...
-			wget
-			which
-			xauth
-			xz
-			zip
-			$(apk search -q mesa-dri)
-			$(apk search -q mesa-vulkan)
-			vulkan-loader
-		"
+		if apk add wolfi-base; then
```
89luca89, line 418: I would declare the common packages in `deps` outside of the "if" and then just add the packages of difference augmenting the common ones like:

```sh
deps="$deps
foo
bar
baz
"
```

89luca89 (changes requested):

distrobox-init
EyeCantCU: Cleaned up the deps lists :)

distrobox-init
```diff
@@ -415,63 +415,74 @@ if [ "${upgrade}" -ne 0 ] ||
...
 		fi
-		deps="
+		if apk add wolfi-base; then
+			deps="
+				busybox
+				gnutar
+				man-db
+				mesa
+				posix-libc-utils
+				sudo-rs
```
89luca89, line 425: This package conflicts with `util-linux-login` that is needed for a full `su` package

Either we remove sudo-rs or we solve this upstream 

image

89luca89 (changes requested):

distrobox-init
```diff
@@ -415,63 +415,74 @@ if [ "${upgrade}" -ne 0 ] ||
...
+				pinentry
+				sudo
+				tar
+				vte3
+				which
+				$(apk search -q mesa-dri)
+				$(apk search -q mesa-vulkan)
+			"
+		fi
+		deps="${deps}
```
89luca89, line 446: it woud be good to change `${deps}` to `${deps:-}` so that we have a default value in case the if/else before fails

89luca89 (changes requested):

89luca89 (approved):

89luca89:
Great stuff @EyeCantCU  thanks! :heart:

EyeCantCU:
No problem Luca! Happy to help out

PR #1194, opened 2024-02-07
Add host terminfo as a fallback

sandorex:
This should allow the user to use a modern terminal without installing anything additional inside the container itself but it will still prioritize terminfo from the container


89luca89:
Hi @sandorex 

I think we can avoid the mount, and just do

`--env \"TERMINFO_DIRS=/usr/share/terminfo:/run/host/usr/share/terminfo\"`

Host's rootfs is mounted inside /run/host so we could use that

89luca89:
Thanks a lot @sandorex !

PR #1214, opened 2024-02-18
Add --hostname argument and DBX_CONTAINER_HOSTNAME env var

sandorex:
I thought it would require more changes but its already passed to podman so why not have it as an option when it requires so little work

### DIFF
--- a/distrobox-create
+++ b/distrobox-create
@@ -29,6 +29,7 @@
 #	DBX_CONTAINER_IMAGE
 #	DBX_CONTAINER_MANAGER
 #	DBX_CONTAINER_NAME
+#	DBX_CONTAINER_HOSTNAME
 #	DBX_CONTAINER_GENERATE_ENTRY
 #	DBX_NON_INTERACTIVE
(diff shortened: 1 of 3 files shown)


sandorex:
This would also remove the need for `--init-hooks 'echo "$(uname -n)" > /etc/hostname'` hacks which while they do kinda work they dont set the env var `HOSTNAME` and require more hacks on top of it

89luca89:
That's a good idea @sandorex thanks!

Can you also adjust the documentation in `docs/usage` and the entry in `useful_tips`? 
Thanks!

sandorex:
> Can you also adjust the documentation in docs/usage and the entry in useful_tips?

Sure ill look into it now

sandorex:
I have not updated the shell autocompletions cause frankly it looks like magic

89luca89:
Autocompletion works with the `--help` so that should already be sorted :)

89luca89:
Thanks a lot @sandorex ! :heart:

PR #1215, opened 2024-02-18
docs: fix hostname matching with --init-hooks

robinhellmers:
The change #1182 had the intention to evaluate 'uname -n' while calling 'distrobox-create'. But had the quotation marks wrong. Double quotation marks must be the outer ones for the evaluation to occur directly.

When changing /etc/hostname, hostname does not change without a reboot. To avoid forcing the user to reboot the container after creating and entering the container, add 'hostname $(uname -n)' to be called with the --init-hooks flag.

See related issue #1164

### DIFF
--- a/docs/useful_tips.md
+++ b/docs/useful_tips.md
@@ -222,7 +222,7 @@ creating it with the following init-hook:
 
 ```sh
 distrobox create --name test --image your-chosen-image:tag \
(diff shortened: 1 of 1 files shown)


89luca89:
Hi @robinhellmers thanks for the PR but this has been obsoleted by #1214 and now we have a dedicated flag to set hostnames

Thanks anyway for the work!

PR #1219, opened 2024-02-20
Check files sums before copy for fix cp the same file error

Boria138:


distrobox-init
```diff
@@ -1605,7 +1605,9 @@ if [ "${nvidia}" -eq 1 ]; then
 
 		if [ "${type}" = "link" ]; then
 			mkdir -p "$(dirname "${dest_file}")"
-			cp -d "${nvidia_lib}" "${dest_file}"
+			if [ "$(md5sum "${nvidia_lib}" | awk '{print $1}')" != "$(md5sum "${dest_file}" | awk '{print $1}')" ]; then
```
89luca89, line 1608: Can we use `cut` instead of `awk`? just to be sure this work on minimal images :+1:

89luca89 (changes requested):

89luca89:
Thanks a lot @Boria138 just a comment to sort out :)

89luca89:
Thanks a lot @Boria138 !

Issue #1209, opened 2024-02-15
[Error] Fedora's shadow-utils passwd implementation needs '--stdin'


nathanchance:
**Describe the bug**

After Fedora switched their `passwd` implementation away from its own package and into `shadow-utils` (https://src.fedoraproject.org/rpms/shadow-utils/c/91360f25a8c8b810d59bec2803a2477a2647c775?branch=rawhide), I am unable to start a container that includes it.

**To Reproduce**

```
$ git show -s --format=%H
5201a2d2e547bf1fe6a89cfdbe29274096b95be8

$ ./distrobox create -i ghcr.io/nathanchance/dev/fedora:latest -n dev-fedora
Creating 'dev-fedora' using image ghcr.io/nathanchance/dev/fedora        [ OK ]
Distrobox 'dev-fedora' successfully created.
To enter, run:

distrobox enter dev-fedora

$ ./distrobox enter dev-fedora
Starting container...                            [ OK ]
Installing basic packages...                     [ OK ]
Setting up devpts mounts...                      [ OK ]
Setting up read-only mounts...                   [ OK ]
Setting up read-write mounts...                  [ OK ]
Setting up host's sockets integration...         [ OK ]
Integrating host's themes, icons, fonts...       [ OK ]
Setting up package manager exceptions...         [ OK ]
Setting up rpm exceptions...                     [ OK ]
Setting up dpkg exceptions...                    [ OK ]
Setting up distrobox profile...                  [ OK ]
Setting up sudo...                               [ OK ]
Setting up groups...                             [ OK ]
Setting up users...                              Error: An error occurred
```

**Expected behavior**

I expect to be able to start my container :)

**Logs**

```
$ podman logs dev-fedora
...
+ '[' /home/nathan '!=' /home/nathan ']'
+ '[' '!' -e /etc/passwd.done ']'
++ cat /proc/sys/kernel/random/uuid
+ temporary_password=7867745a-41ea-4493-8b7b-83d8fd18924a
+ printf %s:%s nathan 7867745a-41ea-4493-8b7b-83d8fd18924a
+ chpasswd -e
+ printf %s: nathan
+ chpasswd -e
+ '[' 0 -eq 0 ']'
+ printf '%s\n%s\n' 7867745a-41ea-4493-8b7b-83d8fd18924a 7867745a-41ea-4493-8b7b-83d8fd18924a
+ passwd root
The password for root is unchanged.
Changing password for root
Enter the new password (minimum of 8 characters)
[...]

PR #1221, opened 2024-02-20
init: Add '--stdin' flag to 'passwd root' call if supported

nathanchance:
Fedora recently changed their `passwd` implementation from its own package to the implementation in `shadow-utils`, adding `--stdin` in the process (which has been subsequently accepted upstream). Without that flag, entering up a Fedora container for the first time fails with

```
+ printf '%s\n%s\n' 7867745a-41ea-4493-8b7b-83d8fd18924a 7867745a-41ea-4493-8b7b-83d8fd18924a
+ passwd root
The password for root is unchanged.
Changing password for root
Enter the new password (minimum of 8 characters)
Please use a combination of upper and lower case letters and numbers.
+ '[' 1 -ne 0 ']'
+ printf 'Error: An error occurred\n'
Error: An error occurred
```

Look for `--stdin` in the help output of `passwd` and use it if it is supported so the root password is properly changed.

Closes: https://github.com/89luca89/distrobox/issues/1209
Link: https://github.com/shadow-maint/shadow/commit/49001ca846d1e2894b4719d66889ab7e1b4b698a

### DIFF
--- a/distrobox-init
+++ b/distrobox-init
@@ -1987,7 +1987,11 @@ if [ ! -e /etc/passwd.done ]; then
 
 	if [ "${rootful}" -eq 0 ]; then
 		# We're rootless so we don't care about account password, so we remove it
-		printf "%s\n%s\n" "${temporary_password}" "${temporary_password}" | passwd root
+		passwd_cmd=passwd
+		if passwd --help 2>&1 | grep -q -- --stdin; then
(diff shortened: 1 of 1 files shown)


89luca89:
Thanks @nathanchance for the patch!

the linting errors can be ignored are my fault on main branch :sweat_smile:

PR #1223, opened 2024-02-20
Prevent podman from logging stderr verbose output as priority 3 journald errors

pavinjosdev:
Podman by default logs stderr as priority 3 journald errors. Don't know if other container managers do the same 😖
Github issue: https://github.com/containers/podman/issues/20728

`distrobox-create` by default enables verbose output for `distrobox-init` which causes the latter to output all xtrace output as priority 3 journald errors. 🤪

This PR redirects stderr in distrobox-init to stdout to prevent this *crazy* behavior. 😌

### DIFF
--- a/.github/workflows/compatibility.yml
+++ b/.github/workflows/compatibility.yml
@@ -72,9 +72,9 @@ jobs:
 
     runs-on: ubuntu-latest
     needs: setup
+    timeout-minutes: 10
     strategy:
       fail-fast: false
-      max-parallel: 8
       matrix:
         distribution: ${{fromJSON(needs.setup.outputs.targets)}}
         container_manager: ["podman"] #, "docker"]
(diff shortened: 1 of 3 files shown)


89luca89:
Hey @pavinjosdev 

I'm unsure about this, will break how `enter` keeps track of the init status:

image

This also has the side effect of entering the container too early on initful ones, because it doesn't wait for `container_setup_done` to be printed on stdout, but catches it on stderr!

A fix can be done by changing the switch-case in `distrobox-enter`, line 584, and remove the leasing `*` so that it will capture only strings starting with the wanted prefix, instead of matching whole lines, would you like to include this changes in this PR? :smiley:

pavinjosdev:
@89luca89 Done, good catch to prevent future issues 😅

89luca89:
Thanks a lot @pavinjosdev !

89luca89 (approved):

PR #1225, opened 2024-02-21
Update useful_tips.md to add note on copying to host clipboard

briandipalma:


docs/useful_tips.md
```diff
@@ -835,3 +836,7 @@ Then reload systemd daemon to apply the changes:
 ```bash
 systemctl --user daemon-reload
 ```
+
+## Copy text to host clipboard
+
+To copy/yank text from the container to the host clipboard you need to install `xsel` in the container for Xorg hosts or `wlroots` for wayland hosts.
```
89luca89, line 842: we need to split this line or the linter will complain

89luca89 (changes requested):

89luca89 (approved):

PR #1239, opened 2024-02-28
Changed echo for printf in distrobox-init

sfalken:
Probably not an issue, but the rpmlint bashism checks caught this one, with the following error: possible bashism in distrobox-init line 2294 (unsafe echo with backslash):

Changed it for printf, with a properly escaped '/n'

### DIFF
--- a/distrobox-init
+++ b/distrobox-init
@@ -2286,7 +2286,7 @@ if [ -e /usr/lib/systemd/systemd ] || [ -e /lib/systemd/systemd ]; then
 	# Start user Systemd unit, this will attempt until Systemd is ready
 	sh -c "	sleep 1 && while true; do \
 	    systemctl is-system-running | grep -E 'running|degraded' && break; \
-	    echo 'waiting for systemd to come up...\n' && sleep 1; \
+	    printf "waiting for systemd to come up...'\n'" && sleep 1; \
 	done && \
 	systemctl start user@${container_user_name}.service && \
 	systemctl start user-integration@${container_user_name}.service && \


distrobox-init
```diff
@@ -2286,7 +2286,7 @@ if [ -e /usr/lib/systemd/systemd ] || [ -e /lib/systemd/systemd ]; then
 	# Start user Systemd unit, this will attempt until Systemd is ready
 	sh -c "	sleep 1 && while true; do \
 	    systemctl is-system-running | grep -E 'running|degraded' && break; \
-	    echo 'waiting for systemd to come up...\n' && sleep 1; \
+	    printf "waiting for systemd to come up...'\n'" && sleep 1; \
```
89luca89, line 2289: Problem here is that the printf quotes actually are closing che `sh -c` ones

I think it's a bit easier with echo, maype we can add an rpmlint exception? Because I already have shellcheck quite scritly set and it didn't give me any problems

89luca89 (changes requested):

sfalken:
oh,  it's completely fine if this is problematic, to just pitch it, the rpmlint check was a warning, not an error, it doesn't affect it building or it's packaging, other than throwing the warning.

I obviously didn't test as much as I should have.

sfalken:
Rescinding PR, It's merely a warning and not an Error, that doesn't cause any actual issues with the downstream packaging.    And I didn't test as much as I should have.

89luca89:
Thanks anyway!

PR #1245, opened 2024-03-01
Update install

briandipalma:


89luca89:
Hi @briandipalma thanks for the PR!
I already had this pushed yesterday but it was on another branch, and I've merges this already

Thanks anyway for the work!

Issue #1249, opened 2024-03-02
With distrobox 1.7.0 gpg-agent stopped working


mcepl:
**Describe the bug**
I am running MicroOS-based distro (which means read-only host and all work done in podman containers using distrobox). Because I am afraid gpg-agent got confused when it was started from inside a container, I am running it on a host with systemd --user services (configuration according to https://wiki.archlinux.org/title/GnuPG#gpg-agent). When trying to decrypt a GPG-encrypted file on host, everything works fine, but when I try to decrypt a file in a container (or when using pass(1)) suddenly I get an error (it worked for months before, and I am really not certain what did change now):

```sh
tumbleweed-pkg~$ LANG=en_GB.utf8 gpg --decrypt ~/.local/share/password-store/mozilla/identita.csob.cz.gpg
gpg: encrypted with rsa4096 key, ID 77D15A36BD4211B2, created 2016-04-27
      "Matěj Cepl <mcepl@cepl.eu>"
gpg: Warning: not using 'D96484AC' as default key: No secret key
gpg: Warning: not using '880BC9D8' as default key: No secret key
gpg: all values passed to '--default-key' ignored
gpg: keydb_search failed: IPC syntax error
gpg: public key decryption failed: No secret key
gpg: decryption failed: No secret key
tumbleweed-pkg~$ 
```

When looking at log-file I see this:

```
2024-03-02 10:53:20 gpg-agent[2434] DBG: chan_10 <- OPTION xauthority=
2024-03-02 10:53:20 gpg-agent[2434] DBG: chan_10 -> ERR 67109140 Chyba syntaxe IPC <GPG agent> - option argument expected
2024-03-02 10:53:20 gpg-agent[2434] DBG: chan_10 <- BYE
```

(“Chyba syntaxe” means obviously “A syntax error” in Czech). I have to admit I am a bit lost in that protocol log. Who is sending that “OPTION xauhtoríty=” line and what should be the right value of it, when running on Wayland (no Xorg around)?

**To Reproduce**
see above

**Expected behavior**
`gpg --decrypt` inside the container just works, perhaps `pinentry` is activated to collect password.

**Logs**
[gpg-agent-log.txt](https://github.com/89luca89/distrobox/files/14470686/gpg-agent-log.txt)

**Desktop (please complete the following information):**
 - Are you using podman, docker or lilipod? podman
 - Which version or podman, docker or lilipod? podman 4.9.3
 - Which version of distrobox? distrobox 1.7.0
[...]

PR #1252, opened 2024-03-03
Don't create empty XAUTH\* variables.

mcepl:
Creating these variables as empty, could have negative impact on other programs. For example, it apparently breaks `gpg-agent` (https://lists.gnupg.org/pipermail/gnupg-users/2024-March/066956.html).

Fixes: https://github.com/89luca89/distrobox/issues/1249

### DIFF
--- a/distrobox-init
+++ b/distrobox-init
@@ -1794,8 +1794,15 @@ export SHELL="\$(getent passwd "\${USER}" | cut -f 7 -d :)"
 # Ensure we have these two variables from the host, so that graphical apps
 # also work in case we use a login session
 test -z \$DISPLAY && export DISPLAY="\$(host-spawn sh -c "echo \\\$DISPLAY")"
-test -z \$XAUTHORITY && export XAUTHORITY="\$(host-spawn sh -c "echo \\\$XAUTHORITY")"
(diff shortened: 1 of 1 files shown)


89luca89:
THanks a lot @mcepl !

distrobox-init
```diff
@@ -1826,8 +1833,15 @@ set -gx SHELL (getent passwd "\$USER" | cut -f 7 -d :)
...
-test -z \$XAUTHORITY && set -gx XAUTHORITY (host-spawn sh -c "echo \\\$XAUTHORITY")
-test -z \$XAUTHLOCALHOSTNAME && set -gx XAUTHLOCALHOSTNAME (host-spawn sh -c "echo \\\$XAUTHLOCALHOSTNAME")
+if test -z \$XAUTHORITY
+    set -gx XAUTHORITY (host-spawn sh -c "printf "%s" \\\$XAUTHORITY")
+    # if the variable is still empty, unset it, because empty it could be harmful
+    test -z \$XAUTHORITY ; and set -e XAUTHORITY
+end
+if test -z \$XAUTHLOCALHOSTNAME
+    set -gx XAUTHLOCALHOSTNAME (host-spawn sh -c "printf "%s" \\\$XAUTHLOCALHOSTNAME")
+    test -z \$XAUTHLOCALHOSTNAME && set -e XAUTHLOCALHOSTNAME
```
mcepl, line 1843: Actually, this is pretty ugly. It should be `; and` instead of `&&`.

PR #1256, opened 2024-03-04
Switch to doas for AlpineLinux as it is the default

misery:
Use doas-sudo-shim to allow to use "sudo" as an alias.

### DIFF
--- a/distrobox-init
+++ b/distrobox-init
@@ -434,7 +434,7 @@ if [ "${upgrade}" -ne 0 ] ||
 				mandoc
 				musl-utils
 				pinentry
-				sudo
+				doas-sudo-shim
 				tar
 				vte3
 				which


misery:
Hi!

```sudo``` should work of course. But since Alpine 3.16.0 they replaced sudo by doas as main package. I thought it would be much more intuitive to use the distro "defaults" as much as possible.
But I can live with ```sudo``` as well. :-)

```
Significant changes
   sudo has been moved to community repository, which means that only latest stable release branch will get security updates in the future.
   Suggested replacement is [doas](https://pkgs.alpinelinux.org/packages?name=doas) and [doas-sudo-shim](https://pkgs.alpinelinux.org/packages?name=doas-sudo-shim).
```
https://alpinelinux.org/posts/Alpine-3.16.0-released.html

89luca89:
Got it, tbh as long as it works, I prefer to keep the various distros as similar as possible (to reduce my maintenance load :P )

PR #1266, opened 2024-03-08
distrobox-init: use "zypper se --match-exact" (instead of "zypper se -x")

phoppermann:
Old zypper versions do not support "se -x". Use the equivalent "--match-exact" instead such that even distributions with old zypper versions can run in distrobox.


89luca89:
Thanks @phoppermann :+1:

PR #1271, opened 2024-03-09
Add info on upgrading package on steamdeck

### LINKED ISSUE #1270: [Suggestion] Improve documentation on usage with steamdeck
The [documentation](https://github.com/89luca89/distrobox/blob/813026c5fb6b114760f320b430037a6e1b14005a/docs/posts/steamdeck_guide.md) (correctly) mentions that `SteamOS` 3.5 and later already comes pre-installed with `distrobox` but it does not mention that `SteamOS` 3.5 is using an old unsupported version of `distrobox` (1.4.2.1-3).

The documentation should make it clear that `distrobox` must be updated to a more current version before it is used (to avoid any time wasted wondering why `distrobox` is not working as expected).

**Additional context**
I am not sure if the updated `distrobox` will be replaced with the older version when `SteamOS` is updated.

nathanpovo:
Fixes #1270 

**References**:

https://steamcommunity.com/app/1675200/discussions/0/4208119548515682878/

https://www.reddit.com/r/SteamDeck/comments/1b4wkig/updating_the_builtin_distrobox/

https://www.reddit.com/r/SteamDeck/comments/t8al0i/install_arch_packages_on_your_steam_deck/

https://help.steampowered.com/en/faqs/view/671A-4453-E8D2-323C

### DIFF
--- a/docs/posts/steamdeck_guide.md
+++ b/docs/posts/steamdeck_guide.md
@@ -1,11 +1,37 @@
 Latest SteamOS (version 3.5 and later) already pre-installed `distrobox` and `podman`.
 
-You might need to upgrade `distrobox` to the latest version before using it since SteamOS provides an older version of
-`distrobox`. You can confirm this by running the command `distrobox version` (on SteamOS 3.5 version 1.4.2.1-3 of
-`distrobox` was installed).
+Before using `distrobox` on SteamOS, it may be necessary to upgrade to the latest version since the version provided by
+SteamOS may be outdated. You can verify the currently installed version by running the command `distrobox version`. For
+instance, on SteamOS 3.5, version 1.4.2.1-3 of `distrobox` is installed.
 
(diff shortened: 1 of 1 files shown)


nathanpovo:
> I don't think it's necessary to remove the read-only thing
> 
> Just install in $HOME and the path should prioritize it right?

In my case, I wanted to replace the existing installation of `distrobox` so that I didn't need to modify anything in the path. When the version of `distrobox` shipped with SteamOS is upgraded, this would allow me to just continue using it as if I never needed to upgrade anything.

If you prefer, I could maybe update the documentation to document both approaches but it might make the documentation more complicated.

nathanpovo:
Done.

I have added 2 options on how to upgrade the version of `distrobox` on the steamdeck.

89luca89:
That's great thanks! :smile:

PR #1274, opened 2024-03-10
Fix Nvidia integration on Fedora containers

Boria138:
When I added file comparison I tested it on Ubuntu and Debian, it worked there, but on Fedora it didn't work as it turned out


Boria138:
If I add quotation marks like shellcheck asks, the code breaks.

89luca89:
Closing due to inactivity, let me know if you want to reopen

PR #1277, opened 2024-03-11
Fix a nit.

mcepl:
Just making the style of the fish script more fish-like and consistent within itself.

Fixes: #1252


89luca89:
Thanks a lot :)

PR #1278, opened 2024-03-11
Update install

briandipalma:


89luca89:
heh always forget this! thanks! :sweat_smile:

PR #1282, opened 2024-03-13
Fixes installation script

ludanin:
Hi, I was trying to update to the latest git version on Bazzite and noticed the installation script was failing when trying to install icons, these changes fixed it for me when installing to `$HOME/.local`


89luca89:
Thanks!

PR #1298, opened 2024-03-20
init: massively speed up apk deps install

JamiKettunen:
Instead of running `apk info` on each package individually instead just compare the list of requested packages to an `apk search` of them and pick the ones that could be found; perhaps something similar could be done for the other package managers as well to avoid such wasted time :p

On alpine due to the way searching for providers of `openssh-client` and exact matching of the found packages (`procps` is `procps-ng`) adjust these both accordingly to keep the end result consistent for alpine and wolfi.

On my end this halves the setup time for the non-toolbox alpine and wolfi images from ~30 and 40 seconds to ~15 and 20 seconds respectively.

### DIFF
--- a/distrobox-init
+++ b/distrobox-init
@@ -421,6 +421,7 @@ if [ "${upgrade}" -ne 0 ] ||
 				gnutar
 				man-db
 				mesa
+				openssh-client
 				posix-libc-utils
 			"
 		elif apk add alpine-base; then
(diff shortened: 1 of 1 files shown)


JamiKettunen:
Hm actually just remembered, should I do something about `openssh-client-default` not matching anything on `alpine:3.13` and below (since there it is `openssh-client`)? https://pkgs.alpinelinux.org/packages?name=openssh-client*&branch=v3.13&arch=x86_64

JamiKettunen:
Since I split some problematic pkgs away from the generic path already it may be better to actually just drop the loop entirely, I'll check on that tomorrow

PR #1299, opened 2024-03-21
Add option to run with clean PATH

sandorex:
The flag name is debatable but ill explain why i would like this feature to exist

_PATH should not be passed by default IMO_, but at least i would like there to be a way to remove it without writing more logic in .profile or .bashrc/.zshrc

This fixes the problem of having cargo, go packages on host get priority over the ones inside the container as my path looks like this
```
...
/home/sandorex/.cargo/bin
/home/sandorex/.golang/bin
/home/sandorex/.npm-packages/bin
...
/home/sandorex/.dbx/dev/.cargo/bin
/home/sandorex/.dbx/dev/.golang/bin
/home/sandorex/.dbx/dev/.npm-packages/bin
...
```
_container `dev` with `~/.dbx` prefix_

### DIFF
--- a/distrobox-enter
+++ b/distrobox-enter
@@ -25,6 +25,7 @@
 # Optional env variables:
 #	DBX_CONTAINER_MANAGER
 #	DBX_CONTAINER_NAME
+#	DBX_CONTAINER_CLEAN_PATH
 #	DBX_SKIP_WORKDIR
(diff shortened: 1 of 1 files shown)


89luca89:
Thanks a lot @sandorex :smile:

Issue #1303, opened 2024-03-22
[Error] Distrobox enter hangs with podman


baduhai:
Please, before opening a bug:

- Ensure there isn't already a closed/resolved issue about this.
    - [There is](https://github.com/89luca89/distrobox/issues/610).

**Describe the bug**
When running `distrobox enter` using podman on the backend, it just hangs at "Starting container...". This was previously fixed in [fea7a84](https://github.com/89luca89/distrobox/commit/fea7a84065ce4493c4ea3e7d2100575f2b1ed664), but has been reverted in [a7df145](https://github.com/89luca89/distrobox/commit/a7df1458f4901dd2ba7dd81f976990c7817932f2).

**To Reproduce**
Use distrobox 1.6.0.1 with podman 4.9.3. Likely affecting other versions of podman.

**Expected behavior**
`distrobox enter` shouldn't hang.

**Logs**
<details><summary><code>distrobox enter debian-latest --verbose</code></summary>

```console
 > distrobox enter debian-latest --verbose     
+ '[' -z debian-latest ']'
+ '[' '!' -t 0 ']'
+ '[' '!' -t 1 ']'
+ case "${container_manager}" in
+ command -v podman
+ container_manager=podman
+ command -v podman
+ '[' 1 -ne 0 ']'
+ container_manager='podman --log-level debug'
+ '[' 0 -ne 0 ']'
+ container_home=/home/user
+ container_path=/run/wrappers/bin:/home/user/.local/share/flatpak/exports/bin:/var/lib/flatpak/exports/bin:/home/user/.nix-profile/bin:/nix/profile/bin:/home/user/.local/state/nix/profile/bin:/etc/profiles/per-user/user/bin:/nix/var/nix/profiles/default/bin:/run/current-system/sw/bin
+ '[' 0 -ne 0 ']'
+ container_status=unknown
++ podman --log-level debug inspect --type container --format 'container_status={{.State.Status}};
        {{range .Config.Env}}{{if slice . 0 5 | eq "HOME="}}container_home={{slice . 5 | printf "%q"}};{{end}}{{end}}
        {{range .Config.Env}}{{if slice . 0 5 | eq "PATH="}}container_path={{slice . 5 | printf "%q"}}{{end}}{{end}}' debian-latest
INFO[0000] /run/current-system/sw/bin/podman filtering at log level debug
[...]

Issue #610, opened 2023-02-04
Dropbox enter hangs via podman


s3rgeym:
Sorry if it is bug of piece of shit, that name is Podman. I don't use this before.

I only see `Starting container...`:

```zsh
~ 
❯ distrobox create -i archlinux -n arch-distrobox
Image archlinux not found.
Do you want to pull the image now? [Y/n]: 
Resolved "archlinux" as an alias (/etc/containers/registries.conf.d/00-shortnames.conf)
Trying to pull docker.io/archlinux/archlinux:latest...
Getting image source signatures
Copying blob 607448f29175 done  
Copying blob 607448f29175 done  
Copying blob 25449fe340a2 done  
Copying config 0e3dc80833 done  
Writing manifest to image destination
Storing signatures
0e3dc80833824e6190b05f591be3eda7dae0fc6348637ccdecb2c79b0e60835c
Creating 'arch-distrobox' using image archlinux  [ OK ]
Distrobox 'arch-distrobox' successfully created.
To enter, run:

distrobox enter arch-distrobox

arch-distrobox

~ took 36s 
❯ distrobox enter arch-distrobox
Container arch-distrobox is not running.
Starting container arch-distrobox
run this command to follow along:

 podman logs -f arch-distrobox

 Starting container...
```

I press `Ctrl + C`:

```zsh
~ took 31m30s 
```

Same with Fedora:

```zsh
❯ distrobox-enter
Error: inspecting object: no such container my-distrobox
Cannot find container my-distrobox
Create it now, out of image registry.fedoraproject.org/fedora-toolbox:36? [Y/n]: 
Creating the container with command:
  /usr/bin/distrobox-create -i registry.fedoraproject.org/fedora-toolbox:36 -n my-distrobox
Image registry.fedoraproject.org/fedora-toolbox:36 not found.
Do you want to pull the image now? [Y/n]: 
Trying to pull registry.fedoraproject.org/fedora-toolbox:36...
Getting image source signatures
Copying blob cc95ab9e6dd0 done  
Copying blob a78a55e65f90 done  
Copying config 445df7cd91 done  
Writing manifest to image destination
Storing signatures
445df7cd91fecfa3e900345e9fc7b92428c31112f60bb730c6fb720b7767704e
Creating 'my-distrobox' using image registry.fedoraproject.org/fedora-toolbox:36         [ OK ]
Distrobox 'my-distrobox' successfully created.
To enter, run:

distrobox enter my-distrobox

my-distrobox
[...]

PR #1307, opened 2024-03-25
fix: missing .fifo logfile in distrobox-enter

michaelhaaf:
## Overview

For the past week I've noticed issues where distroboxes will spontaneously crash in CLI, especially during a long process like system package updates. I find it difficult to reproduce consistently, but something like a large `paru -Syu` or `apt update && apt upgrade -y` for example, would almost always trigger the crash.

Upon trying to re-run the distrobox, the process would hang:

```
 Starting container...                          
```

After interrupting with CTRL-C:

```
/usr/bin/distrobox-enter: line 514: /home/user/.cache/.ubuntu-distrobox.fifo: No such file or directory
```

[This comment](https://github.com/89luca89/distrobox/issues/610#issuecomment-1971360249) in #610 suggests that the original fix for this problem has been reverted. 

I've tried that out on my own system with this code in this pull request and so far 
I am no longer experiencing the missing .fifo nor crashing container problem. 

## Details

- issue #610: bug and fix description
- original fix: fea7a84
- commit reverting the fix: a7df145

## Related

- issue #1303

### DIFF
--- a/distrobox-enter
+++ b/distrobox-enter
@@ -575,13 +575,16 @@ if [ "${container_status}" != "running" ]; then
 		fi
 		# save starting loop timestamp in temp variable, we'll use it
 		# after to let logs command minimize possible holes
-		${container_manager} logs --since "${log_timestamp}" -f "${container_name}" 2> /dev/null \
-			> "${app_cache_dir}/.${container_name}.fifo" &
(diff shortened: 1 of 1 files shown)


89luca89:
mmh Seems like the revert was for CI problems
That now are not present anymore, so we can re-merge the fix

Thanks a lot @michaelhaaf !

PR #1309, opened 2024-03-26
Update install

briandipalma:
I think this may be incorrect, I can't update to the latest release:

image


89luca89:
Thanks but it was just me missing the tag sorry 

Thanks anyway for the PR tho :)

PR #1316, opened 2024-03-30
docs: Specify that distrobox is in the extra repo

nicknamenamenick:


89luca89:
Thanks for the clarification!

PR #1318, opened 2024-03-30
export: consider only absolute candidate icon paths that exist

JamiKettunen:
Previously (for nearly two years since 4522f29?) attempting e.g. `distrobox-export -a htop` with a file/dir named `htop` in `$PWD` it would mistakenly get detected as the icon for the program, copied over to `~/.local/share/icons/` and attempted to be used by the `.desktop` file.

Not sure if this is following the spec exactly but it does fix this very odd issue I only noticed because I happened to have a local git tree checkout of a program in my `$HOME` and while there tried to export a launcher for it to my host and dug deeper when the icon was showing up as entirely blank in GNOME.

### DIFF
--- a/distrobox-export
+++ b/distrobox-export
@@ -477,7 +477,8 @@ export_application() {
 		icon_name="$(grep Icon= "${desktop_file}" | cut -d'=' -f2- | paste -sd "@" -)"
 
 		for icon in ${icon_name}; do
-			if [ -e "${icon_name}" ]; then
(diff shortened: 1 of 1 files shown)


JamiKettunen:
@89luca89 Would you mind taking a look at some of the open pull requests when you have time? Thanks :)

89luca89:
Hi @JamiKettunen sorry was pretty busy :)

LGTM thanks!

PR #1324, opened 2024-04-02
init: massively speed up apt/pacman/xbps deps install

JamiKettunen:
For more details check the commit descriptions. Originally this just had `apt` changes, let me know if I should split the others to different PRs.

## apt
Similar to the previous `apk` optimization (#1298) instead of checking each package one by one instead grab the matching ones from `apt list` and pass them directly over to `apt-get install`.

## pacman
On Arch we can `grep` the `pacman -Ssq` output (all packages) for the ones we're interested in.

`--needed` was already used previously but was dropped with bdf78c39f2bd71296d85da2f0b4a0b7a9624e20c, commented there as well.

## xbps
Similar to Arch on Void we can instead `grep` the `xbps-query -Rs '*'` output (all packages) for the ones we're interested in after filtering out install status, description and versions.

### DIFF
--- a/distrobox-init
+++ b/distrobox-init
@@ -600,14 +600,8 @@ if [ "${upgrade}" -ne 0 ] ||
 			libvulkan1
 			mesa-vulkan-drivers
 		"
-		install_pkg=""
-		for dep in ${deps}; do
-			if apt-cache policy "${dep}" | grep Candidate | grep -qv none; then
-				install_pkg="${install_pkg} ${dep}"
-			fi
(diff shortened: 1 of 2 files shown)


89luca89:
Hi @JamiKettunen this won't work on older Debian/Ubuntu, as `apt` has an unstable CLI interface
Took the liberty to reimplement your idea but with `apt-cache show` so that we keep backward compatibility :+1: 

Anyway that's one hell of a job thank you!

PR #1327, opened 2024-04-04
Support musl variant of Void Linux

JamiKettunen:
The "C library diversity" is one of the  main points listed on https://voidlinux.org so we might as well support it here too :)


89luca89:
Great job, thanks!

PR #1328, opened 2024-04-05
all: properly handle env vars and configs in all commands

xarvex:
While I was adding an option to quiet the output (which I will make a PR for after this), I noticed that not all commands handled environment variable options and/or configs, even though they use options that would be affected such as `non_interactive` and `verbose`.
There are as well certain tweaks for consistency and readability, please see commits for specifics on that. I have highlighted the important details here:

## Environment Variables

`DBX_NON_INTERACTIVE` was added in 7203563, but has not been added to new files since then.
`DBX_VERBOSE` was added in 78b3319, but was not handled in all files that use verbose as an option.

## Configs

Were not being processed in all commands, despite being relevant to them.

## Affected

Not counting minor tweaks.

* `distrobox-assemble`
* `distrobox-ephemeral`
* `distrobox-export`
* `distrobox-host-exec`
* `distrobox-init`


89luca89:
Thanks a lot @xarvex! Not in all places this is needed (see export,host-exec,init)

Reverted that for those, and keeping for all the others :) thanks!

PR #1340, opened 2024-04-11
export: cancel export when trying to delete

### LINKED ISSUE #1332: [Error] Deleting the export of unexported binary will export it
**Describe the bug**
Deleting an export of a binary that has not been exported yet is actually exporting it.

**To Reproduce**
1. Install fish
2. Run `distrobox-export --delete --bin /usr/bin/fish`

**Expected behavior**
I would expect it to throw an error. 

**Additional context**
This is most likely caused by a missing return statement in this line:
https://github.com/89luca89/distrobox/blob/867f4f85c762bc74bf398c8654984b801138e154/distrobox-export#L382-L383

taukakao:
When deleting an export that doesn't exist, the script would instead export the application.

Fixes #1332

### DIFF
--- a/distrobox-export
+++ b/distrobox-export
@@ -371,15 +371,19 @@ export_binary() {
 	mkdir -p "${dest_path}"
 
 	# If we're deleting it, just do it and exit
-	if [ "${exported_delete}" -ne 0 ] &&
+	if [ "${exported_delete}" -ne 0 ]; then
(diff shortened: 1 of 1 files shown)


89luca89:
This will also Fix https://github.com/Vanilla-OS/apx/issues/344

Thanks a lot @taukakao

PR #615, opened 2023-02-07
Add support for macOS

kevindurb:
This gets the installer working on macos systems, you still need to `brew install coreutils` so we might want to add that to the README.md

This also changes mounting of volumes for macos only to skip `/, /dev, /tmp, /sys` and changes the user home mount to be a "regular" shared mount rather than rslave

PR #1364, opened 2024-05-01
install: use more portable install(1) arguments

JamiKettunen:
GNU arguments `-D` and `-t` are respectively either not implemented or different in [FreeBSD `install(1)`](https://man.freebsd.org/cgi/man.cgi?query=install&sektion=1&format=html) which is used on [Chimera Linux](https://chimera-linux.org); use the options that are known to be compatible between the implementations.

Should also help with #615.


89luca89:
great! thanks a lot! :smile:

PR #1390, opened 2024-05-12
use fedora-toolbox 40 by default

alexsch01:


konstantinblaesi:
What happened to this PR, why was it not merged?

PR #1397, opened 2024-05-16
Fix SHELL using absolute path and disable duplicate profile inclusion

eugenesan:
This v2 of the PR.

In previous version I assumed that the issue was absolute path
in SHELL variabe. Leaving only the basename eliminated the hanging on start
but subshell still didn't work.
After doing additional testing I figured that that wasn't the issue.

The issue of broekn shells seems to be related to duplicate inclusion of distrobox hooks
when profile.d available.
While provided patch works a more comprehensive solution might be required to cover all cases.

In general we should inject hooks only once (first start or creation).
Since injecting on every start prevents customization of the container by the user
and can make it very difficult to identify shell/profile related issues (ask me how I know ;-))

### DIFF
--- a/distrobox-init
+++ b/distrobox-init
@@ -1792,19 +1792,21 @@ printf "distrobox: Setting up distrobox profile...\n"
 # This ensures compatibility with prompts and tools between toolbx and distrobox
 touch /run/.toolboxenv
 
-# Ensure we have some basic env variables and prompt as base.
-rcfiles="
-	/etc/profile
-	/etc/bash.bashrc
-	/etc/bashrc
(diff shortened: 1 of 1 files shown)


eugenesan:
P.S.
We should avoid injecting stuff into container on every start and limit to creation or initial start.
Doing injections repeatedly prevent user from "tuning" the system can make it difficult to identify issues.
It took a while to figure why it didn't work and why I was unable to fix it by simply modifying files in the container.

89luca89 (approved):

89luca89:
Thanks a lot :)

Issue #1399, opened 2024-05-19
[Error] All locales are not generated in Debian 11 container


Boria138:
**Describe the bug**
When I enter the locale -a command, I get this output

```
locale: Cannot set LC_CTYPE to default locale: No such file or directory
locale: Cannot set LC_MESSAGES to default locale: No such file or directory
locale: Cannot set LC_COLLATE to default locale: No such file or directory
C
C.UTF-8
POSIX
en_US.utf8
```

In Fedora 40 I get an output with all locales that are available because of this, when trying to enter any other characters than English in Debian 11 I get “????”, but when I roll back to 1.6.0.1 the error disappears.

**To Reproduce**
distrobox create debian --image debian:11

**Expected behavior**
All locales are generated and everything works, or at least the user's locale is generated in addition to the English locale

**Desktop (please complete the following information):**
 - Are you using podman, docker or lilipod? : podman
 - Which version or podman, docker or lilipod?: 5.0.3-1
 - Which version of distrobox? :1.7.2.1-1
 - Which host distribution?: Arch Linux
 - How did you install distrobox?: From repo

**Additional context**

Boria138:
After generating the desired locale via 
```
sudo dpkg-reconfigure locales 
```
All bugs related to the lack of locale were fixed

Boria138:
Just checked on arch linux rolling to an older version really helped, but in Fedora it doesn't, strange

Boria138:
Fixed this in https://github.com/89luca89/distrobox/pull/1404

89luca89:
Fixed in #1404 :)

dimztimz:
This issue should be reopened as it is not fully fixed in v1.8.2.4 for any distros based on Debian, and very likely for any distros that do not generate all glibc locales. With the fix #1404, only the locale mentioned in the variable LANG in the file `/etc/locale.conf` on the host is installed in the container. `/etc/locale.conf` can contain mixed locales, for example:

```
LANG="en_US.UTF-8"
LC_NUMERIC="it_IT.UTF-8"
LC_TIME="it_IT.UTF-8"
LC_MONETARY="it_IT.UTF-8"
LC_PAPER="it_IT.UTF-8"
LC_NAME="it_IT.UTF-8"
LC_ADDRESS="it_IT.UTF-8"
LC_TELEPHONE="it_IT.UTF-8"
LC_MEASUREMENT="it_IT.UTF-8"
[...]

PR #1404, opened 2024-05-22
Generate host locales in container (fixed issue #1399)

Boria138:


89luca89:
Thanks a lot @Boria138 just added a fallback in case the variables were not set :+1: 

great stuff!

PR #1413, opened 2024-05-25
Allow basic sandboxing of containers

xdom:
Lets the users decide to unshare host home and root directories, and to create unprivileged containers without dropping security measures. Default behaviour of commands stays as-is. Partially related to issue #28.

### DIFF
--- a/completions/zsh/_distrobox-create
+++ b/completions/zsh/_distrobox-create
@@ -16,6 +16,7 @@ _arguments -s \
   '(--init -I)'{-I,--init}'[use init system inside the container]' \
   '--nvidia[try to integrate host nVidia drivers in the guest]' \
   '--unshare-devsys[do not share host devices and sysfs dirs from host]' \
+  '--unshare-groups[do not forward user''s additional groups into the container]' \
   '--unshare-ipc[do not share ipc namespace with host]' \
   '--unshare-netns[do not share the net namespace with host]' \
   '--unshare-process[do not share process namespace with host]' \
--- a/completions/zsh/_distrobox-create
+++ b/completions/zsh/_distrobox-create
@@ -17,9 +17,11 @@ _arguments -s \
   '--nvidia[try to integrate host nVidia drivers in the guest]' \
   '--unshare-devsys[do not share host devices and sysfs dirs from host]' \
   '--unshare-groups[do not forward user''s additional groups into the container]' \
+  '--unshare-home[do not mount host home directory into the container]' \
   '--unshare-ipc[do not share ipc namespace with host]' \
   '--unshare-netns[do not share the net namespace with host]' \
   '--unshare-process[do not share process namespace with host]' \
+  '--unshare-root[do not mount host root directory into the container]' \
   '--unshare-all[activate all the unshare flags]' \
   '(--compatibility -C)'{-C,--compatibility}'[show list of compatible images]' \
   '(--help -h)'{-h,--help}'[show this message]' \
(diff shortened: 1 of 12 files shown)


89luca89:
Hi @xdom I'll take a deeper look at this ASAP this is a big change and I'd like to do further testing

89luca89:
@xdom @janvhs 

I'd like to reflect on the implication of this (and #28 )

What are the use cases for this?

The main use case of using `distrobox` over `podman` (or `docker`) is the integration it has with the system.
And also the "ease of use", as you **could** do with podman what distrobox does, but distrobox abstracts lots of flags (so you can just `distrobox enter` and it works)

This invalidates all the usecases of distrobox
- app/bin exporting
- host integration
- ease of use (we need to add all the various flags to use this)

So my question is:

Isn't using podman directly a better alternative?

I feel like all of this PR (and #28 in general) are going in the opposite direction of what the projects aims to do.

Sandboxing is *not* a target for distrobox, it may add some unsharing bits for useful purposes (say, initful containers) but it's not aiming to replace podman/docker/flatpak

89luca89:
> Thanks for the response @89luca89. For me the use case is primarily having a separate home for my development distrobox, because the tools I use: (Elixir's mix, Rust's cargo, OCaml's opam etc.) litter my home directory. Therefore, I keep my main home clean and remove the complete distrobox when I decide it's time for some clean-up.
> 
> That said, I agree, an unshared root and unprivileged mode might go to far for distrobox's scope and you'd be better of using Podman or bubblewrap directly. This way it's not implied that distrobox is a sandboxing tool.

That's why `--home` is present
it does indeed allow to separate the dotfiles, but keeping the integration (original home is accessible anyway)

then, if a program uses `/home/$USER` instead of `$HOME`, it's a problem of that program. 

But all the unprivileged thing, unsharing host, and completely unsharing home, feels backwards with what distrobox aims to do, and is useful for.

If one doesn't want an integration with the host system, why are they using distrobox in the first place?

WayneSherman:
> Sandboxing is _not_ a target for distrobox...

There was a time when this wasn't much of a concern, but because of widespread cybersecurity threats sandboxing needs to be a target for ALL desktop environments.  Moving forward, running apps and environments in a sandbox is hard requirement for me.  For example, I don't want most apps (e.g. LibreOffice, Gimp, Firefox, vlc) to have general access to all my mounted network shares, or all my dotfiles (especially ~/.ssh folder where my private keys are stored).  Same thing for building software, the build environment should not have access to my whole user profile or all my network mounts.  If distrobox doesn't have the ability to run apps and environments in a safe way (i.e. sandboxed), then unfortunately I may not be able to use it.

WayneSherman:
> If one doesn't want an integration with the host system, why are they using distrobox in the first place?

Even on the host system I want to run apps in a sandbox.  Whether it is a native host app or a distrobox app I want to define the amount of integration and what is visible to the app and hidden from the app.

89luca89:
> > Sandboxing is _not_ a target for distrobox...
> 
> There was a time when this wasn't much of a concern, but because of widespread cybersecurity threats sandboxing needs to be a target for ALL desktop environments. Moving forward, running apps and environments in a sandbox is hard requirement for me. For example, I don't want most apps (e.g. LibreOffice, Gimp, Firefox, vlc) to have general access to all my mounted network shares, or all my dotfiles (especially ~/.ssh folder where my private keys are stored). 

and that's where flatpak exist

> Same thing for building software, the build environment should not have access to my whole user profile or all my network mounts. If distrobox doesn't have the ability to run apps and environments in a safe way (i.e. sandboxed), then unfortunately I may not be able to use it.

and that's where regular docker/podman exist

> Even on the host system I want to run apps in a sandbox. Whether it is a native host app or a distrobox app I want to define the amount of integration and what is visible to the app and hidden from the app.

Distrobox target is to have multiple mutable environments, for compatibility or development purposes.
If security is a concern (for example, you don't trust the build you're performing) you should use proper docker/podman. It doesn't make sense to have a distrobox that works at 50% (or less) because of unsharing all the stuff.

The point of distrobox is integration, compatibility and ease of use, if you take away this, what is the advantage of running bare podman/docker?

89luca89:
Closing as not planned

PR #1434, opened 2024-06-07
Seperate curl and wget commands

axtloss:
This separates the `curl` and `wget` install commands in the readme to make it easier to use the github copy button on the code view to copy the command.


89luca89 (approved):

89luca89:
Thanks a lot @axtloss :)

PR #1439, opened 2024-06-16
compatibility: add missing url for ubuntu:23.04

sheevy:
23.04 was mentioned as supported, but the url wasn't provided. urls are source of truth for `--compatibility`, hence the addition.


89luca89:
Thanks @sheevy

PR #1451, opened 2024-06-22
Add kali icon

andyspectre:
Hopefully I got the distrobox icon proportions (almost) right.


89luca89:
@andyspectre the icon is a bit blurry, is it upscaled? maybe we can find something less blurry?

thanks!

andyspectre:
Yes it was upscaled a bit, trying to get the size and proportions right, the asymmetry of the kali logo doesn't help. I agree with you is a bit blurry, I will find something better.
Thanks

andyspectre:
Hi @89luca89 , I have fixed the kali icon, now it is less blurry. It is also centered and the distrobox icon has the right proportions. What do you think?
Thanks

89luca89:
Thanks @andyspectre

HShazen:
for my case it doesn't appear it is invisible so i edit one for my pc. 
but now i find this conversation saying that you added the icon so is the issue in my pc as when I went to /.local/share/icons/distrobox/ I only find the icons of other distro but not kali linux one

so just want to know if the issue is in my system or not as i am using the latest version of distrobox!

HShazen:
So am using the latest version of distrobox on fedora 40 but the icon of kali is invisible when i create a kali linux container using distrobox-create getting this result, (for other distro like Ubunto and Fedora it is working normal)

I don't know if this issue is in my system or not but I have always to change the icon manually for now, you can see that when i see the .desktop file the path is not as other distro and isn't going to (~/.local/share/icons/distrobox/distro_name.png) I have no idea could any one tell me why.

Untitled


Screenshot from 2024-07-06 00-29-51

PR #1296, opened 2024-03-19
init: ignore findmnt error in routine remounting, Fix #1289

chitao1234:
#1289 is because of apt hook `/etc/distrobox-pre-hook.sh` umounted `/etc/localtime`, so findmnt would error out.
Ignore findmnt error in this case.

Issue #1289, opened 2024-03-16
[Error] apt-get sometime crashes the container


chitao1234:
**Describe the bug**
If I use `apt-get` to install a lot of packages in one go, the container might crash with exit code `137`, `podman inspect` shows `"ExitCode": 1`

**To Reproduce**
```shell
$ distrobox create -n test -i ubuntu:jammy
$ distrobox enter test
$ sudo apt-get install qemu-system
```
qemu-system has a lot of dependencies, so it's quite likely the container would crash.

**Expected behavior**
The packages is installed normally, as it is in a normal podman container.

**Logs**
[create.log](https://github.com/89luca89/distrobox/files/14622456/create.log)
[enter.log](https://github.com/89luca89/distrobox/files/14622454/enter.log)
[podman-logs.log](https://github.com/89luca89/distrobox/files/14622455/podman-logs.log)

**Desktop (please complete the following information):**
 - Are you using podman, docker or lilipod?
 podman
 - Which version or podman, docker or lilipod?
 podman version 4.9.3
 - Which version of distrobox?
distrobox: 1.7.0.1
 - Which host distribution?
EndeavourOS (Archlinux based)
 - How did you install distrobox?
With pacman

**Additional context**
The problem does NOT occur if I changed the `sleep 15` to `sleep 150` in [distrobox-init](https://github.com/89luca89/distrobox/blob/7a56b6e899574c428cdfd3c228198d639149c1a8/distrobox-init#L2175)

jakubiszon26:
I have really simmilar issue but it's pacman that causes distrobox to crash

Lesabotsy:
Tried both Ubuntu and Arch, both crash on package install/update.

chitao1234:
Changing the sleep duration to 150 lowers the probability of the crash happening, and the patch #1296 should fix this reliably.

espentrydal:
> Changing the sleep duration to 150 lowers the probability of the crash happening, and the patch #1296 should fix this reliably.

I tried to apply the changes in the patch manually, it did not make any difference for me. `sleep 150` seems to fix it though.

6cdh:
That patch fixed it for me. I applied it manually and recreate my dev container because the crash corrupted pacman database.

RodoMa92:
Thanks for the temporary patch. I'll test it shortly and report back. Sadly my pacman database was killed in the process and I also had to rebuild the container.

RodoMa92:
[...]

PR #1454, opened 2024-06-26
init: do not fail in "Setting up read-only mounts" if findmnt does not exist

phoppermann:
If `findmnt` does not exist (or returns an error), "Setting up read-only mounts" in `distrobox-init` fails with "Error: An error occurred".

Treat missing `findmnt` the same way as if the file/directory cannot be read.

It seems like some error handling for failing `findmnt` was already forseen:
```
flags="$(findmnt --noheadings --output OPTIONS --target "${src}" || :)"
```
However, the code apparently cannot handle `flags=":"`.

I'm completely unsure if my approach is reasonable or which implications it might have but it allows to at least enter the distrobox.

### DIFF
--- a/distrobox-init
+++ b/distrobox-init
@@ -246,6 +246,11 @@ get_locked_mount_flags()
 	prev=""
 	locked_flags=""
 
+	# If findmnt does not exist, exit
+	if ! command -v findmnt 2> /dev/null > /dev/null; then
+		return 0
+	fi
+
 	# If we can't read the file/directory, exit
 	if ! ls "${src}" 2> /dev/null > /dev/null; then
 		return 0


89luca89:
Hi @phoppermann 

why does this happen?

`findmnt` is added on the pkg setup step, it should definitely be there?

phoppermann:
I don't know. Where is the code which adds `findmnt` in pkg setup? Then I can take a look what's going on.

phoppermann:
https://github.com/89luca89/distrobox/blob/3365ff9ca73f065a3fe36f256d52102226e6b759/distrobox-init#L1375

Apparently, the package does not contain findmnt in the very old version I'm trying to use yet.

However, it doesn't probably doesn't really matter. https://github.com/89luca89/distrobox/pull/1296 should handle this situation as well. Like described in the PR description, this doesn't work (anymore?). So potentially there's also a regression wrt #1289. Solving this would likely fix my issue with missing `findmnt` as well.

89luca89:
@phoppermann if you give more info about the setup/distro you're going for, I can try to reproduce

phoppermann:
You can reproduce my problem by changing
https://github.com/89luca89/distrobox/blob/24716c1b1802222b6475f3387c89b48948e79f98/distrobox-init#L258
to e.g.
```
flags="$(:)"
```
which is what would also happen in any error case.

phoppermann:
It works correctly with
```
flags=":"
```
which is probably the intended behavior.

Note that https://github.com/89luca89/distrobox/commit/b79a7059b88f2018d9e07c6265e4aaba767d6ec0 places the `|| :` behind (and not inside of) `$()`.

89luca89:
@phoppermann 
ok got it

can you share with me which guest distro/image you're using? to understand why findmnt is failing

phoppermann:
It's not failing, it just doesn't exist at all. I'm using a custom image based on an ancient SLES so nothing you could use for testing.

89luca89:
ah got it, welp if it doesn't exists we can indeed skip this that's ok

thanks for helping!

phoppermann:
Ok. There's probably still the problem if `findmnt` really fails which should be tracked in a separate issue.

PR #1456, opened 2024-06-26
Fix issue with user not having permissions to write containers.conf

nhermosilla14:
This is a simple fix, I just replaced a redirection with a pipe to `sudo tee`, because the command as-is does not work unless you run it as root (which doesn't seem to be the intended usage for that step-by-step).


89luca89:
Thanks @nhermosilla14 for the fix :)

PR #1458, opened 2024-06-27
enter: Fix support for Active Directory usernames including backslashes pt. 2

phoppermann:
When the `container_command_user` contains a backslash, it must be quoted for `getent passwd` (otherwise `getent passwd` returns nothing).

see https://github.com/89luca89/distrobox/commit/7473298ee692e5ac2e331fab4fac856c99d6e768


89luca89:
Thanks @phoppermann for the fix!

PR #1468, opened 2024-07-06
 generate-entry: use XDG_DATA_HOME

Samueru-sama:
`$XDG_CONFIG_HOME` was already being used in a similar way, but anyway this makes it follow `$XDG_DATA_HOME` before defaulting to `~/.local/share` for the icons and .desktop files. 

Yes I don't have `~/.local/share` 😆


89luca89:
Thanks a lot! that's a nice fix :)

PR #1489, opened 2024-07-24
Update openSUSE and SLES compatibility guide

alexandrevicenzi:
The issue with Podman was fixed on recent releases.


89luca89:
Thanks!

PR #1491, opened 2024-07-25
distrobox-generate-entry: Ignore the case of os-release.ID

xzl01:
Some Linux distributions have non-standardized release.IDs, resulting in Distrobox failing to correctly recognize and match icons.


xzl01:
Here is deepin os release：

PRETTY_NAME="Deepin 23"
NAME="Deepin"
VERSION_ID="23"
VERSION="23"
ID=Deepin
HOME_URL="https://www.deepin.org/"
BUG_REPORT_URL="https://bbs.deepin.org"
VERSION_CODENAME=beige

PR #1496, opened 2024-07-27
export: Fix XDG_DATA_DIRS and XDG_DATA_HOME handling

e-luks:
Swaps if condition resulting in the matching environment variable being iterated


softmoth (approved):
This looks good to me, simple bug fix.

89luca89:
Thanks a lot for the PR @e-luks and @softmoth  for checking!

PR #1503, opened 2024-08-01
Exclude repos from bind-mounting

lunacd:
When used on a Universal Blue OS like bluefin, the repo configurations are available at `/run/host/etc/yum.repos.d/`. Repos that contain the name nvidia are incorrectly bind-mounted into the container and causes weird errors on boxes using yum and dnf.

See the issue at https://github.com/ublue-os/bluefin/issues/1552


89luca89:
Thanks for the fix @lunacd !

Issue #1482, opened 2024-07-18
[Error] Unable to enter container from a NixOS host ("Empty hostname")


jeaye:
**Describe the bug**
I'm unable to enter new containers, due to an error with the container hostname.

```bash
diff: /run/host/etc/static/hostname: No such file or directory
hostname: Empty hostname
Error: An error occurred
```

This error happens on multiple distros. I tested `archlinux:latest` and `ubuntu:latest` and they both have the same error.

**To Reproduce**
On NixOS 24.05, run the following:

```bash
distrobox create arch --image archlinux:latest
distrobox enter arch
```

**Expected behavior**
I expect to be able to enter the containers without error.

**Logs**
* distrobox logs (Arch): [distrobox-arch-enter.txt](https://github.com/user-attachments/files/16276505/distrobox-arch-enter.txt)
* docker logs (Arch): [docker-arch.txt](https://github.com/user-attachments/files/16276508/docker-arch.txt)
* distrobox logs (Ubuntu): [distrobox-ubuntu-enter.txt](https://github.com/user-attachments/files/16276509/distrobox-ubuntu-enter.txt)
* docker logs (Ubuntu): [docker-logs-ubuntu.txt](https://github.com/user-attachments/files/16276510/docker-logs-ubuntu.txt)

**Desktop (please complete the following information):**
 - Are you using podman, docker or lilipod? `docker`
 - Which version or podman, docker or lilipod? `Docker version 24.0.9, build v24.0.`
 - Which version of distrobox? `distrobox: 1.7.2.1`
 - Which host distribution? `NixOS 24.05`
 - How did you install distrobox? `Nix`

**Additional context**
I do very much suspect that this is related to NixOS just doing things different, BUT this used to work. I have old containers that I was previously using (before a NixOS upgrade -- which also included a distrobox upgrade). I can still enter those old containers.

For the record, my hostname file:

```bash
[nix-shell:~]$ ls -lh /etc/hostname
lrwxrwxrwx 1 root root 20 Jul 17 17:59 /etc/hostname -> /etc/static/hostname

[nix-shell:~]$ cat /etc/hostname
thelio
```

lesserfish:
Hello!

This is indeed a bug caused by NixOS's weird file hierarchy. I've submitted a pull request with a bug fix (#1506).

In case you are in a hurry and want to get this working right away, replace line 2409 of the file `distrobox-init` from 

    file_watch_src="/run/host${file_watch}"

to

[...]

PR #1506, opened 2024-08-03
Fix: Correct file path resolution for symlinks in NixOS

lesserfish:
The issue happens in NixOS, where the file /etc/hostname is a symlink to /etc/static/hostname which is itself another symlink to a file in /nix/store.

This causes the variable file_watch_src to be set to an incorrect path.

By using 'readlink -f', we can ensure that file_watch_src points to the correct file location describing the hostname.

This change ensures that the path resolution works correctly for both regular files and symlinks.

This issues was described in #1482

### DIFF
--- a/distrobox-init
+++ b/distrobox-init
@@ -2406,7 +2406,7 @@ if [ "${init}" -eq 0 ]; then
 			# host in some cases like /etc/localtime, so ignore findmnt errors
 			mount_source="$(findmnt -no SOURCE "${file_watch}")" || :
 			if [ -n "${mount_source}" ] && ! echo "${mount_source}" | grep -q "${id}"; then
-				file_watch_src="/run/host${file_watch}"
+				file_watch_src="readlink -f /run/host${file_watch}"
 				# check if the target file exists
 				if ls -l "${file_watch_src}" 2> /dev/null > /dev/null; then
 					# if it's a symlink and take the source


89luca89:
No fix needed

lesserfish:
Hello! It's been a while, so I had to revisit the issue to see what exactly was failing. I've cloned the last version and it still does not work on NixOS.

You're right—the previous solution worked by chance.

The real cause of the NIX issue is in line 2433 https://github.com/89luca89/distrobox/blob/922175d456b8a5f1af7648943eaa524c006b5744/distrobox-init#L2433 

The output of `cat /etc/hostname` is an empty string, which causes the `hostname` method to fail with the error `hostname: Empty hostname`.

It seems that this can be solved by removing the following lines (https://github.com/89luca89/distrobox/blob/922175d456b8a5f1af7648943eaa524c006b5744/distrobox-init#L2428-L2429

Why are you unmounting this again?

I need a little more time to figure out how everything works, but right now, the current version of distrobox is NOT working on NixOS.

89luca89:
As replied in the issue
it IS working if you use the repo version

you NEED to mount /nix:/nix for this to work, and the repo version does inject this configuration
Either use that or use the config to add this mount by default

it is NOT on distrobox to support an exotic file structure layout

lesserfish:
Jesus, Chill the fuck out. I'm trying to be as respectful as I can.

PR #1545, opened 2024-09-04
Fix distrobox-enter and distrobox-rm to handle containers that have environment variables shoter than 5 symbols

senioroman4uk:
When I was trying to use distrobox with rocm/pytorch:latest I've noticed the following error 
```
Error: template: inspect:3:27: executing "inspect" at <slice . 0 5>: error calling slice: index out of range: 5
```

It turns out this image has an environmental variable CI=1, that causes an exception when distrobox-enter / distrobox-rm is used with this docker image

Added some additional ifs to the corresponding commands to handle this case

### DIFF
--- a/distrobox-enter
+++ b/distrobox-enter
@@ -541,8 +541,8 @@ container_status="unknown"
 eval "$(${container_manager} inspect --type container --format \
 	'container_status={{.State.Status}};
 	unshare_groups={{ index .Config.Labels "distrobox.unshare_groups" }};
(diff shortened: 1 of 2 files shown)


senioroman4uk:
Fixes #1546

softmoth (approved):
Looks good to me!

softmoth (changes requested):
Actually, it's probably better to use the [`startswith` function](https://jqlang.github.io/jq/manual/#startswith).

mirkobrombin (approved):

89luca89:
Thanks @senioroman4uk for the PR and thanks @mirkobrombin for reviewing!

Issue #1542, opened 2024-09-01
[Error] Cannot remove distrobox created with root privileges


cmonty14:
**Describe the bug**
After creating successfully some distrobox with root privileges I now want to remove obsolete ones.
However this fails with error.

**To Reproduce**

1. Identify distrobox with root privileges: `distrobox ls -r`
```
❯ distrobox ls -r
ID           | NAME                 | STATUS             | IMAGE                         
c27aa1d3022a | tumbleweed-pkg       | Exited (143) 3 months ago | registry.opensuse.org/opensuse/distrobox-packaging:latest
d20ff279a71c | arch-root            | Up 20 hours        | docker.io/library/archlinux:latest
d56f9032f5d6 | debian               | Exited (143) 3 weeks ago | docker.io/library/debian:latest
```
2. Remove distrobox with root privileges: `distrobox rm -r [container-name]`

**Expected behavior**
Distrobox with root privileges is deleted w/o error.

**Logs**
[distrobox-rm.txt](https://github.com/user-attachments/files/16827939/distrobox-rm.txt)

Attach also the output of `podman logs` or `docker logs`, possibly with `--latest` flag

**Desktop (please complete the following information):**
 - Are you using podman, docker or lilipod? `podman version 5.2.2`
 - Which version of distrobox? `distrobox: 1.7.2.1`
 - Which host distribution? [openSUSE MicroOS](https://microos.opensuse.org/)
 - How did you install distrobox? _Included in OS_

softmoth:
This is a duplicate of #1400. It's already been fixed in master but no release has been made since May 2024.

89luca89:
Duplicate of #1400 fixed by 952ffaa75717dcd8ba05962b945af80057fdba88

PR #1564, opened 2024-09-20
Eval the call to $container_manager rm instead of a plain call

dcermak:
The shell tries to execute `$container_manager` directly, which fails if it is a composite command like `sudo podman` as that is not a valid binary. By instead evaluating the whole command, we circumvent this issue.

This fixes https://github.com/89luca89/distrobox/issues/1542

### DIFF
--- a/distrobox-rm
+++ b/distrobox-rm
@@ -390,7 +390,7 @@ delete_container()
 	# Remove the container
 	printf "Removing container...\n"
 	# shellcheck disable=SC2086,SC2248
-	${container_manager} rm ${force_flag} --volumes "${container_name}"
+	eval "${container_manager} rm ${force_flag} --volumes \"${container_name}\""
 
 	# Remove exported apps and bins
 	cleanup_exports "${container_name}"


softmoth (changes requested):
The source of this problem is that `IFS` is set in `cleanup_exports`, and not restored, so normal word splitting does not happen.

The correct fix is to restore IFS at the bottom of `cleanup_exports`, and no `eval` is required. The other scripts use `unset IFS` to restore it, which should be portable enough.

I think it's worth auditing the rest of the scripts, too, and creating a consistent method for restoring IFS. Some places use `unset IFS` (correct), others use `IFS=" "` (it's incorrect, but kind of works) at the site of needing to split on spaces.

I propose every use of `IFS=X` should be paired with `unset IFS`. And uses of `IFS=" "` should be removed or replaced with `unset IFS` when at the top of a loop body.

dcermak:
> The source of this problem is that `IFS` is set in `cleanup_exports`, and not restored, so normal word splitting does not happen.
> 
> The correct fix is to restore IFS at the bottom of `cleanup_exports`, and no `eval` is required. The other scripts use `unset IFS` to restore it, which should be portable enough.

Isn't it there already? https://github.com/89luca89/distrobox/blob/c05b6a43769bfa56d572a457f1420e0e2589fe3b/distrobox-rm#L329

mirkobrombin:
Thanks for your contribution. Can you clarify which is the case when this happens? Using the `--root` flag (correct way to use podman as root) I don't have the problem. Can you explain your use case so I can evaluate?

89luca89:
This is already fixed in `main` by 952ffaa75717dcd8ba05962b945af80057fdba88

PR #784, opened 2023-06-08
New website

daudix-UFO:
Initial work on improving website experience, now website uses [os-component-website](https://github.com/jimmac/os-component-website) as a template for nicer look

Preview available at [daudix-ufo.github.io/distrobox](https://daudix-ufo.github.io/distrobox)

PR #1565, opened 2024-09-21
Switch to Duckquill for documentation

daudix:
It's been a while since #784, in which I've created my own [Zola](https://www.getzola.org) theme.

*Demo is available on [daudix.github.io/distrobox](https://daudix.github.io/distrobox/).*

Screenshot 2024-09-21 at 20-50-22 Home - Distrobox

The main differences are that all Markdown files now live under the `content` directory, all layouts and styling is now managed by [Duckquill](https://duckquill.daudix.one) submodule, which means easy updates, and of course the looks, copy button for code blocks, cross-site search, and other QoL improvements :D

In terms of syntax everything is pretty similar to Jekyll, the link syntax is different though, which means they don't work without building the site.

Since GitHub Pages automatically build only Jekyll sites, I've added a new workflow that builds the site to the `gh-pages` branch on every new commit in the `release` branch, and tests the build on pull requests. GitHub Pages source will need to be changed to the `gh-pages` branch.

Screenshot From 2024-09-21 21-30-15


89luca89:
Thanks @daudix is there a way to keep the `doc` directory as is?
the idea is that docs here, and website (and man pages too!) are all in sync and easily discoverable

sending the repo docs in `content` seems like "hiding" them from the repo

daudix:
I understand your concerns, the content can't be directly under the `docs` directory, and I didn't know that the man pages are being generated from the docs, so I'll turn this into a draft until the details are worked out.

89luca89:
> I understand your concerns, the content can't be directly under the `docs` directory, and I didn't know that the man pages are being generated from the docs, so I'll turn this into a draft until the details are worked out.

got it thanks

maybe we can have `./docs` in the main repo (like now) and symlink into content? would that work?

89luca89 (changes requested):
@daudix can we do the reverse?

right now ./docs is a link to ./site/content, can we do so that ./site/content is the link to docs?
so that docs are kept where they are

89luca89:
Thanks @daudix 
I really like the website, but I see a problem with the navigation and presentation of the docs here on github (or gitlab for the mirror)

Right now we miss the navigation (toc) due to not using pure markdown but some custom code like:

```
+++
title = "Useful Tips"
[extra]
toc = true
+++
```

this is kind of a deal breaker for me, I really value the docs on gh/gl pages and this makes them hard to follow read and search

daudix:
Yeah that's fair. Current docs work well enough and generate from vanilla markdown, and display properly right on GitHub. Sorry for wasting your time.

89luca89:
Thanks to you, and the big effort you made
Sorry if this didn't work out

PR #1567, opened 2024-09-25
fix: switchup in distrobox-export canon_dirs lookup

TigerGorilla2:
Idk what to put here, the changes should be obvious...
Checking for one variable and then using another isn't ideal :p


89luca89:
This has been solved in ccedbd701de7e711839a5c8fe636583f06fe675d

PR #1577, opened 2024-10-03
Validate if the binary is running in a flatpak container

### LINKED ISSUE #1574: [Suggestion] Validate if the binary is running in a flatpak container
When we export a bin from a distrobox, a script is generated that can be executed on the host. When opening this script, we have something like this:

```bash

#!/bin/sh
# distrobox_binary
# name: boxkit
if [ -z "${CONTAINER_ID}" ]; then
		exec "distrobox-enter"  -n boxkit  --  '/usr/sbin/kubectl'  "$@"	
elif [ -n "${CONTAINER_ID}" ] && [ "${CONTAINER_ID}" != "boxkit" ]; then
	exec distrobox-host-exec '/var/home/lucasribeiro/.local/bin/kubectl' "$@"
else
	exec '/usr/sbin/kubectl' "$@"
fi

```

The thing is, we cannot execute this same binary in an application installed via flatpak because the $PATH of the flatpak container does not identify the distrobox.

It would be nice to add something like this to the logic for creating these scripts to access bins within distroboxes:

```bash

#!/bin/sh
# distrobox_binary
# name: boxkit
if [ -z "${CONTAINER_ID}" ]; then
	if [ -n "${FLATPAK_ID}" ]; then
		exec flatpak-spawn --host distrobox enter -n boxkit  --  '/usr/sbin/kubectl'  "$@"	
	else
		exec "distrobox-enter"  -n boxkit  --  '/usr/sbin/kubectl'  "$@"	
	fi
elif [ -n "${CONTAINER_ID}" ] && [ "${CONTAINER_ID}" != "boxkit" ]; then
	exec distrobox-host-exec '/var/home/lucasribeiro/.local/bin/kubectl' "$@"
else
	exec '/usr/sbin/kubectl' "$@"
fi


```

lucasvsr:
Closes #1574

### DIFF
--- a/distrobox-export
+++ b/distrobox-export
@@ -320,7 +320,11 @@ generate_script()
 # distrobox_binary
 # name: ${container_name}
 if [ -z "\${CONTAINER_ID}" ]; then
-	exec "${DISTROBOX_ENTER_PATH:-"distrobox-enter"}" ${rootful} -n ${container_name} ${enter_flags} -- ${sudo_prefix} ${container_command_suffix}
+	if [ -n "\${FLATPAK_ID}" ]; then
+		exec flatpak-spawn --host "${DISTROBOX_ENTER_PATH:-"distrobox-enter"}" ${rootful} -n ${container_name} ${enter_flags} -- ${sudo_prefix} ${container_command_suffix}
+	else
+		exec "${DISTROBOX_ENTER_PATH:-"distrobox-enter"}" ${rootful} -n ${container_name} ${enter_flags} -- ${sudo_prefix} ${container_command_suffix}
+	fi
 elif [ -n "\${CONTAINER_ID}" ] && [ "\${CONTAINER_ID}" != "${container_name}" ]; then
 	exec distrobox-host-exec '${dest_path}/$(basename "${exported_bin}")' "\$@"
 else


mirkobrombin:
Hi, thanks for your contribution. Can you explain me what's your reasoning? Why is that implemented for the export func only?

lucasvsr:
> Hi, thanks for your contribution. Can you explain me what's your reasoning? Why is that implemented for the export func only?
> 
> PS. I am referring to the PR's title, not the issue. Maybe the title is wrong?

This PR solves the Issue I created: https://github.com/89luca89/distrobox/issues/1574

lucasvsr:
> Can you update the PR's title to reflect the issue?

Ok, it's done.

mirkobrombin:
I discussed this internally with Luca, we think this is a bit out of scope and should be managed by the flatpak itself. Having distrobox handling this case, could open the project to possible bugs. I can think of missing flatpak permissions, unexposed portals and so on.

image

For the records, the above screenshot is a shell inside the Bottles' Flatpak which has permissions to access the home directory.

lucasvsr:
> I discussed this internally with Luca, we think this is a bit out of scope and should be managed by the flatpak itself. Having distrobox handling this case, could open the project to possible bugs. I can think of missing flatpak permissions, unexposed portals and so on.
> 
> image
> 
> For the records, the above screenshot is a shell inside the Bottles' Flatpak which has permissions to access the home directory.

This is another scenario. I understand that the bug shown in the screenshot was caused by a lack of permissions in flatpak and not by any implementation in distrobox. The user who uses distrobox is advanced and has reasonable knowledge of Linux to fix what is happening.

Given my problem, which led me to the solution I am proposing here:

I am a developer, I use IntelliJ as an IDE and I needed kubectl to integrate the app with the cluster I use at work.

IntelliJ in flatpak already has access to the bins I have on the host at the time of installation, they are not available in the sandbox $PATH but it has access in _/run/host/bin_. However, when I tried to use the binary generated by distrobox (available in _$HOME/.local/bin_ and flatpak has full access to my home directory), it simply did not recognize kubectl because distrobox worked fully on the host and not in the sandbox.

When I made this adjustment manually, it started working.

mirkobrombin:
I understand your needs and thanks for the PR, but I can't accept it because it's not Distrobox's job to handle how apps are launched outside of its context, like with flatpak-spawn. Changing the export like this could lead to unexpected behaviors and open the road to 'hacky' solutions in the future, which I want to avoid.

I hope you understand my point. I'm still open to discuss other ideas that solve the problem without denaturing Distrobox.

89luca89:
Thanks @mirkobrombin I agree with you
this has nothing to do with distrobox, if you're running a command that runs on an host, *you* need to use flatpak-spawn, it's not on distrobox to support this

We can continue the discussion on the suggestion issue if needed

PR #1582, opened 2024-10-10
export: Actually fix XDG_DATA_DIRS and XDG_DATA_HOME handling

TigerGorilla2:
This fixes the use of `XDG_DATA_DIRS` and `XDG_DATA_HOME` in accordance to the specification.
As per the [specification](https://specifications.freedesktop.org/basedir-spec/0.8/#basics)
> There is a set of preference ordered base directories relative to which data files should be searched. This set of directories is defined by the environment variable `$XDG_DATA_DIRS`.

and
> There is a single base directory relative to which user-specific data files should be written. This directory is defined by the environment variable `$XDG_DATA_HOME`.

This change now uses the correct fallbacks for each variable (was swapped) and does not try to iterate over `XDG_DATA_HOME` as it should only contain a single directory.

### DIFF
--- a/distrobox-export
+++ b/distrobox-export
@@ -425,19 +425,17 @@ export_application()
 		desktop_files="${exported_app}"
 	else
 		IFS=":"
-		if [ -n "${XDG_DATA_HOME}" ]; then
-			for xdg_data_home in ${XDG_DATA_HOME}; do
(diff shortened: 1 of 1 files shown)


89luca89:
Thanks @TigerGorilla2 LGTM!

PR #1583, opened 2024-10-10
init: separate 'mount --rbind -o rslave ...' into 'mount --rbind ...'…

timwa0669:
… and 'mount --make-rslave ...'

On some legacy distros like Ubuntu 18.04.6, the mount executable does not support '-o rslave'. So we use '--make-rslave' instead.


89luca89:
Thanks @timwa0669 LGTM!

PR #1584, opened 2024-10-10
create: keep $SHELL directory prefix

timwa0669:
When $SHELL variable does not contain directory prefix like 'zsh', the 'script ...' command run via python may failed, causing compatibiltiy dropping.

This bug was found while using Petalinux 2020.1 SDK, running 'petalinux-config -c kernel'.

### DIFF
--- a/distrobox-create
+++ b/distrobox-create
@@ -690,7 +690,7 @@ generate_create_command()
 	result_command="${result_command}
 		--label \"manager=distrobox\"
 		--label \"distrobox.unshare_groups=${unshare_groups}\"
-		--env \"SHELL=$(basename "${SHELL:-"/bin/bash"}")\"
+		--env \"SHELL=${SHELL:-"/bin/bash"}\"
 		--env \"HOME=${container_user_home}\"
 		--env \"container=${container_manager}\"
 		--env \"TERMINFO_DIRS=/usr/share/terminfo:/run/host/usr/share/terminfo\"


timwa0669:
Should this change become optional? Regarding the following situation I list below, maybe this PR will cause the introduction of a new issue.

Assume I use zsh, if the shell executables are in different places between host and containers:
Host: the 'zsh' can be found under /bin
Container: the 'zsh' can be found under /usr/bin, not /bin

89luca89:
> Should this change become optional? Regarding the following situation I list below, maybe this PR will cause the introduction of a new issue.
> 
> Assume I use zsh, if the shell executables are in different places between host and containers: Host: the 'zsh' can be found under /bin Container: the 'zsh' can be found under /usr/bin, not /bin

yes that's the reason this exists

SHELL is then set in the `init` to the regular path afterwards using profiles
https://github.com/89luca89/distrobox/blob/21661400fa65e72b4b160ae905aaa12644538a38/distrobox-init#L2006

timwa0669:
> > Should this change become optional? Regarding the following situation I list below, maybe this PR will cause the introduction of a new issue.
> > Assume I use zsh, if the shell executables are in different places between host and containers: Host: the 'zsh' can be found under /bin Container: the 'zsh' can be found under /usr/bin, not /bin
> 
> yes that's the reason this exists
> 
> SHELL is then set in the `init` to the regular path afterwards using profiles
> 
> https://github.com/89luca89/distrobox/blob/21661400fa65e72b4b160ae905aaa12644538a38/distrobox-init#L2006

Thanks for your comment! I know why the $SHELL isn't updated. I use zsh. Zsh won't source /etc/profile.d/*.sh on Ubuntu 18.04.6 after entering the container. So the $SHELL var was not updated.

Maybe we should change the place of 'export SHELL="\$(getent passwd "\${USER}" | cut -f 7 -d :)" ' to support non-bash shells?

89luca89:
> Thanks for your comment! I know why the $SHELL isn't updated. I use zsh. Zsh won't source /etc/profile.d/*.sh on Ubuntu 18.04.6 after entering the container. So the $SHELL var was not updated.
> 
> Maybe we should change the place of 'export SHELL="$(getent passwd "${USER}" | cut -f 7 -d :)" ' to support non-bash shells? (Or any way to let non-bash shells to source /etc/profile.d/distrobox_*.sh?)

I don't know about zsh, but I'd say we need to find the equivalent file where to put the source

89luca89 (changes requested):
This won't work
let's find a way to make zsh source the profile correctly

PR #1586, opened 2024-10-12
Cut 80% of the source package size by optimizing/cleaning doc images

eugenesan:
While packaging the new version of Distrobox,
I've noticed that 90% of size of the source distributions taken by images.
Some of them were unused, some were not runtime related and stored in non web optimized format.

So I took the initiative and reduced the size by ~80%.
AFAIK,  there are not any visual or functional consequences.


89luca89:
Thanks @eugenesan !

Issue #1425, opened 2024-05-30
[Suggestion] Provide fish shell completions /  Fix commands format in the man page


BuonHobo:
**Is your feature request related to a problem? Please describe.**
Distrobox man pages specify commands using the following layout:
```
distrobox create

              --image/-i:     image to use for the container  default: ${container_image_default}
              --name/-n:      name for the distrobox          default: ${container_name_default}
              --hostname:     hostname for the distrobox      default: <container-name>.$(uname -n)
              --pull/-p:      pull the image even if it exists locally (implies --yes)
              --yes/-Y:       non-interactive, pull images without asking
```

This is fine for a human, but it causes fish shell to have buggy completions:
```
❯ distrobox-create --absolutely-disable-root-password-i-am-really-positively-sure 
--absolutely-disable-root-password-i-am-really-positively-sure  (COMPATIBILITY for a list of compatible images and container managers, please …)
--additional-flags/-a                                                                                                                           
--additional-packages/-ap                                                                                                                       
--clone/-c                                                                  (this will be useful to either rename an existing distrobox or hav…)
--compatibility/-C                                                                                                                              
--dry-run/-d                                                                                                                                    
--help/-h                                                                                                                                       
--home/-H                                                                                                                                       
--hostname                                                                                                                                      
--image/-i                                                                                                                                      
--init/-I                                                                   (this will make host[aq]s processes not visible from within the co…)
--init-hooks                                                                                                                                    
--name/-n                                                                                                                                       
--no-entry                                                                                                                                      
--nvidia                                                                                                                                        
--pre-init-hooks                                                                                                                                
--pull/-p                                                                                                                                       
--root/-r                                                                   (way over [dq]sudo distrobox[dq] (note: if using a program other t…)
```

As you can see, the completion are in the format --longcommand/-shortcommand and if you complete your prompt using TAB then the completion will not work. You have to manually go and delete the last part

**Describe the solution you'd like**
I'd like the distrobox man page to list commands in the man page in the classic format, or at least I'd like for it to provide correct fish shell completions which is all I really care about.

89luca89:
I'm not familiar at all with fish, so I'm going to ask help to anyone interested in this :)

boredsquirrel:
https://github.com/89luca89/distrobox/issues/1671

esteve:
I've started the work to convert the man pages to use a more standard format at https://github.com/89luca89/distrobox/pull/1755, it's still a draft PR

[...]

PR #1588, opened 2024-10-13
provide default prompt for fish shell

Boria138:


scarf005:
doesn't this fix #1425 ?

PR #1591, opened 2024-10-14
Replace tac with awk to be more portable

shtayerc:
Since the latest release distrobox-enter does not work on system without tac command, this should make it more portable.

`tac` is GNU coreutils specific command - it is not present in bsd utils. There is also `tail -r`, but the parameter is not POSIX compliant.


89luca89:
Thanks @shtayerc I assumed tac was mostly there as it's also part of busybox, but also this solution works :) thanks!

PR #1597, opened 2024-10-31
docs: update Podman inside Distrobox

alexandrevicenzi:
Podman requires certain privileges to work properly inside a container, suggest to use a rootful Podman or Docker.

See [this] for some context.

[this]: https://bugzilla.suse.com/show_bug.cgi?id=1231231


89luca89:
Thanks @alexandrevicenzi !

PR #1598, opened 2024-11-02
Added a --dbus option

jmgibson1981:
Took a bit of searching to find the proper way of getting dbus active inside the container. Added a --dbus option to fill it in automatically as called.


89luca89:
Thanks @jmgibson1981 I appreciate the effort
I'm not keen to add tons of flags when this is simply solved by using the already present `additional-flags` option
so I'd like to keep it like now

Thanks for the work

Issue #1475, opened 2024-07-12
[Error] Zsh completions not auto-completing container names


UncleGravity:
**Describe the bug**
Running `distrobox-enter <TAB>` doesn't show the correct completion in zsh.  It works correctly in bash.  I think the zsh completions have a bug so I fixed it in my local machine and now completions work correctly.  I don't know if this is a "shippable" fix but I wanted to share here.

```bash
#compdef distrobox-enter

_distrobox_containers() {
  local -a containers
  containers=(${(f)"$(distrobox list --no-color | tail -n +2 | cut -d'|' -f2 | tr -d ' ')"})
  _describe 'container name' containers
}

_arguments \
  '(--name -n)'{-n,--name}'[name for the distrobox]:container name:_distrobox_containers' \
  '(--root -r)'{-r,--root}'[launch podman/docker/lilipod with root privileges]' \
  '(--help -h)'{-h,--help}'[show this message]' \
  '(--verbose -v)'{-v,--verbose}'[show more verbosity]' \
  '(--version -V)'{-V,--version}'[show version]' \
  '*:container name:_distrobox_containers'
```

same for distrobox-rm
```bash
#compdef distrobox-rm

_distrobox_containers() {
  local -a containers
  containers=(${(f)"$(distrobox list --no-color | tail -n +2 | cut -d'|' -f2 | tr -d ' ')"})
  _describe 'container name' containers
}

_arguments \
  '(--all -a)'{-a,--all}'[delete all distroboxes]' \
  '(--force -f)'{-f,--force}'[force deletion]' \
  '--rm-home[remove the mounted home if it differs from the host users one]' \
  '(--root -r)'{-r,--root}'[launch podman/docker/lilipod with root privileges]' \
  '(--help -h)'{-h,--help}'[show this message]' \
  '(--verbose -v)'{-v,--verbose}'[show more verbosity]' \
  '(--version -V)'{-V,--version}'[show version]' \
  '*:container name:_distrobox_containers'
```

The same issue in distrobox-stop, upgrade, etc. 

**To Reproduce**
1. Create new container
2. In zsh, run `distrobox-enter <TAB>`

**Expected behavior**
Expecting a list of available containers. But it returns a list of files in the local directory.

**Desktop (please complete the following information):**
 - Are you using podman, docker or lilipod?
 Podman v5.1.1

 - Which version of distrobox?
distrobox v1.7.2.1

 - Which host distribution?
NixOS

 - How did you install distrobox?
[...]

PR #1603, opened 2024-11-05
zsh completions reworked. fixed not completing container names.

fredricocalamari:
Fix for issue #1475


89luca89:
Thanks @fredricocalamari :heart:

PR #1605, opened 2024-11-05
Strip single quotes when parsing locale.conf

catgirlconspiracy:


89luca89:
Thanks @catgirlconspiracy

PR #1608, opened 2024-11-07
potential improvement to pull-request #1102

AbduEhab:
I ran into an issue when exporting. I mentioned it here: https://github.com/89luca89/distrobox/issues/1097#issuecomment-2463285814

I have reworked a small part of the code to utilize the `-0` option of `xargs`

I want to know if there are any tests that I need to do to make sure everything is ok since I could not find any.


89luca89:
Thanks @AbduEhab looks great! :)

PR #1614, opened 2024-11-14
fix:repeat trigger update locale when locale exists

heysion:
Description:

The $HOST_LOCALE contains the '-' symbol, but the output of locale -a does not include this symbol, which leads to repeated updates when updating the locale.

Log:

fixed can correctly skip over these duplicates update.
 
```
 + [ -e /etc/locale.gen ]
 + locale -a
 + grep -qi en_us.utf8 
 + locale -a
 + echo zh_CN.UTF-8
 + tr -d - 
 + grep -qi zh_CN.UTF8 
```


89luca89:
Thanks @heysion

PR #1618, opened 2024-11-20
Added clone in distrobox assemble

### LINKED ISSUE #1617: distrobox-assemble clone option
Currently, I am unable to clone an existing Distrobox when using the distrobox-assemble tool. While distrobox-assemble is useful for creating multiple Distroboxes, it does not allow cloning of an existing Distrobox in the same way the create command does.

I would like the ability to clone an existing Distrobox when using the distrobox-assemble command. Specifically, the clone option should be passed to the create command, similar to the behavior of the create command by itself.

At this time, I have not found an alternative method to clone a Distrobox with the distrobox-assemble tool.

**Additional context**
It would be helpful to add a --clone flag to the distrobox-assemble command to support this feature. For reference, here is the [distrobox-assemble image section](https://github.com/89luca89/distrobox/blob/18053c254a83750c49c08e58df2e48a0f04aef48/distrobox-assemble#L342) in the code.

PawsFunctions:
Closes #1617

Added ability to clone an existing Distrobox when using the distrobox-assemble command.

### DIFF
--- a/distrobox-assemble
+++ b/distrobox-assemble
@@ -46,6 +46,7 @@ entry=""
 home=""
 hostname=""
 image=""
+clone=""
 init=""
 init_hooks=""
 nvidia=""
@@ -272,6 +273,7 @@ run_distrobox()
 	home=""
 	hostname=""
 	image=""
+	clone=""
 	init=""
 	init_hooks=""
 	nvidia=""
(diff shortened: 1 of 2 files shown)


89luca89:
Thanks @PawsFunctions :) nice idea!

PR #1636, opened 2024-12-18
Stop importing zsh FPATH into container

jgroman:
FPATH shell variable is used by zsh for looking up shell function definitions. 
Proposed change stops importing host FPATH environment variable into container. FPATH is supposed to be shell variable only and should not even be exported into shell environment in the first place. But in a case where it was exported this will stop propagating host FPATH value into container where it might collide with and overwrite proper guest FPATH contents.


89luca89:
Thanks @jgroman !

PR #1664, opened 2025-01-21
Update install

briandipalma:


89luca89:
Thanks!

briandipalma:
That doesn't seem to be enough, it looks like the distrobox version is specified in multiple places?

PR #1667, opened 2025-01-23
create: Support --userns=keep-id:size=

Jookia:
Currently distrobox will reserve all subuids/subgids available. This has the unfortunate effect of blocking other containers from running that require their own set of subuids/subgids.

podman has added a feature that lets us restrict the amount of subuids/subgids we request. If available, set the amount to 65536 which is the standard amount needed for a typical Linux single container.

This breaks the ability to run nested containers.

---

To use this feature you must compile and run the git version of podman. Without it this happens with regular distrobox use:

```
$ podman run --rm -it --userns=auto archlinux /bin/echo "hello world"
hello world
$ distrobox create --image archlinux test1
...
$ podman run --rm -it --userns=auto archlinux /bin/echo "hello world"
Error: creating container storage: not enough unused IDs in user namespace
$ distrobox rm test1
...
$ podman run --rm -it --userns=auto archlinux /bin/echo "hello world"
hello world
```

An alternative solution to this could be an option flag to allow specifying the userns type and flags.

### DIFF
--- a/distrobox-create
+++ b/distrobox-create
@@ -934,6 +934,11 @@ generate_create_command()
 			result_command="${result_command}
 				--userns keep-id"
 		fi
+
+		# Add :size=65536 if possible
+		if podman run --rm --userns=keep-id:size=65536 ${container_image} /bin/true 2>/dev/null || [ "$?" -eq 127 ] ; then
+			result_command="${result_command}:size=65536"
+		fi
 	fi
 
 	# Add additional flags


89luca89 (approved):
Thanks @Jookia seems like an interesting addition!

Jookia:
I think the right way is to ask for more UIDs for your container if you want to run nested podman. Maybe a way to specify that if you can't already would be good?

89luca89:
> I think the right way is to ask for more UIDs for your container if you want to run nested podman. Maybe a way to specify that if you can't already would be good?

I think we can simply skip this limit if explicitly set (eg via env variable at creation time), nested containers are quite a corner case, so I don't believe we want too much of an ergonomic solution (cli flags and such)
and I think limiting the space is the right default at the moment

I think adding this env variable check, plus updating the docs for nested containers needs, is enough

Jookia:
Yeah I can't figure out how to do this. Sorry.

89luca89 (approved):

Jookia:
Thank you!

PR #1675, opened 2025-02-03
assemble: add unshare-groups support

PoQuatre:
Added support for the `unshare-groups` option in `distrobox assemble`


dottorblaster (approved):
Thanks!

PR #1679, opened 2025-02-05
refactor: replace "Silverblue" with "Atomic Desktops"

boredsquirrel:
Fedora Atomic Desktops is the name for all of the variants, Silverblue is just the GNOME one


dottorblaster (changes requested):
Awesome, thanks! Since Silverblue is pretty popular as a name, we would like to keep a reference to it. Would you like to change this as suggested?

89luca89 (approved):

PR #1694, opened 2025-02-21
docs: replace Arch Linux example in the manifest file

xz-dev:
The existing Arch Linux example in the manifest file was replaced to improve clarity and alignment with the recommended practices.

- Updated additional_packages for better readability
- Refined configuration options to reflect current use cases
- Simplified and optimized pre/post init hooks
- Added comments to prevent potential eval-related issues

This will help users better understand how to write and use manifests for Arch Linux within the system.


dottorblaster:
Thanks! Unfortunately this comes at the expense of simplicity and we try to follow the KISS principle ourselves, even when documenting stuff.

Keep rocking with distrobox!

Issue #1712, opened 2025-03-18
[Error] File globbing triggered during file parsing in distrobox-assemble


jeffmahoney:
**Describe the bug**
If a single-character file exists in the directory where distrobox-assemble is executing *and* it shares a character with one contained in the name of a distrobox to be assembled, it will fail.

```
 - Deleting ...
Error response from daemon: No such container: my-distrobox
Error response from daemon: No such container: my-distrobox
Cannot find container my-distrobox.
 - Creating ...
 already exists
```

`my-distrobox` is the default container name and it's used because `$name` is never set. This is due to file globbing in the matching a single character file in the current directory, so the line becomes:
```w```

instead of
```[tw]```

and the section definition is not interpreted.

This is regression introduced in commit 1bb261a08a2. 

```
        # shellcheck disable=SC2013
        for line in $(cat "${file}"); do
                # Remove comments and trailing spaces
```

The unquoted file contents are being interpreted by the shell rather than being passed as read.

**To Reproduce**
Create a distrobox.ini:

```[tw]
image=registry.opensuse.org/opensuse/distrobox:latest
init=true
nvidia=false
pull=true
root=false
replace=true
start_now=false
additional_packages="git vim ctags go1.24 osc fileb0x obs-service-go_modules"
volume="/src:/src"
volume="/var/scratch:/var/scratch"
```

and create a file in the same directory:
`:> w`

Then `distrobox assemble create`

**Expected behavior**
A distrobox named `foo` will be created

PR #1713, opened 2025-03-18
assemble: fix accidental file globbing during parsing

jeffmahoney:
commit 1bb261a08a2 (assemble: fix regression in file reading) introduced a different regression in file parsing.

When creating a distrobox I observed:
```
     - Deleting ...
    Error response from daemon: No such container: my-distrobox
    Error response from daemon: No such container: my-distrobox
    Cannot find container my-distrobox.
     - Creating ...
     already exists
```

In the loop reading the file, the file contents are being interpreted by the shell rather than being passed as read.  This results in the section header being treated as a single-character glob expression. If a file or directory exists in the working directory and is named with a single character contained in the section header, the line will only contain that character rather than the section header.  This naturally results in the distrobox not being created because the start of the section is never detected, $name, goes unset, and run_distrobox is called without a name.  `distrobox rm` uses a default name of "my-distrobox" which, on my test system doesn't exist.  Then `distrobox create` is called with an empty name, which fails, but that's interpreted as already existing by `run_distrobox`.

It appears a few separate changes went into that commit and restoring the while read loop fixes the issue for me.

Fixes: #1712
Fixes: 1bb261a08a2


ocraviotto:
This is a problem with the latest distrobox
```
$ distrobox --version
distrobox: 1.8.1.2
```

Not sure when it got updated in Fedora, but I started to suffer it recently: :
```
$ distrobox assemble create --replace --file ~/fedora.ini
 - Deleting ...
Please touch the device.
[sudo] password for craviotto:
Error: no such container my-distrobox
Error: no such container my-distrobox
Cannot find container my-distrobox.
 - Creating ...
 already exists
```

The simple change in this PR fixes it for me. Anything missing to merge?

Issue #1554, opened 2024-09-11
[Suggestion] nopull or cached-image flag


lethedata:
**Is your feature request related to a problem? Please describe.**
if `container_always_pull=1` or `DBX_CONTAINER_ALWAYS_PULL=1` there's not a flag to overwrite these and still use local cache. The use case for overwriting these would be that one always wants to pull but there is a locally build image that needs to be tested and without a flag to disable it distrobox always tries to pull.

**Describe the solution you'd like**
Have a `--nopull` flag to disable pulling images all together (new behavior) and only use the local cache OR a `--cached-image` flag to use the cached image if available locally otherwise pull (same as default behavior).

**Describe alternatives you've considered**
Setting `DBX_CONTAINER_ALWAYS_PULL=0` then calling distrobox works but seems like a flag would be useful.

**Additional context**
The `--cached-image` flag is probably easier to implement as it's just another case statement. I have an untested solution on [my fork](https://github.com/lethedata/distrobox/commit/43205473629695c9ed2c415d316e8115eca5a3a4) that might work for this.

PR #1733, opened 2025-04-22
create: add cached-image option

lethedata:
Add --cached-image/-ci flag to set container_always_pull to 0 when container_always_pull or DBX_CONTAINER_ALWAYS_PULL ENV variables are set to 1.

[#1554](https://github.com/89luca89/distrobox/issues/1554)


dottorblaster:
Thanks! Instead of creating another option to set `container_always_pull` to 0, you might just want to set `container_always_pull` to 0.

Keep on rocking with distrobox! :heart:

Issue #1758, opened 2025-05-26
[Discussion] How to best assemble distroboxen that use a mixture of container managers?


klmcwhirter:
Is your feature request related to a problem? Please describe.
How do you recommend we go about using distrobox:
* with both podman and docker on the same host?
* where apps and/or binaries are exported from both?

Describe the solution you'd like
I am not proposing this yet, but I did find a work around. Is this inline with your thinking? 

> Note the patch is just a POC. I did not spend time thinking about edge cases, error handling, etc. I was just curious if I was on the right track or not.

* apply this [patch](https://github.com/klmcwhirter/oci-shared-images/blob/2-support-multiple-dbx_container_managers-on-the-same-system/distrobox-export-1.8.1.2.patch) to `distrobox-export` version 1.8.1.2.
* COPY the result into place in Containerfile (used as spec for image= in distrobox.ini)
* `init_hook="sudo umount /usr/bin/distrobox-export"` in distrobox.ini
* call enter: `DBX_CONTAINER_MANAGER=docker distrobox enter ...`

But is this the best way?

Of course, if the patch was a part of the distrobox code I would not need the first 3 steps.

Please do let me know if the patch is something you would entertain as a PR. (But what am I missing?)

Describe alternatives you've considered

I did consider adding conditional logic - per container - in the distrobox.conf file. But upon inspection, the conf file is read before the command line is parsed. So there is currently no way to determine the container_name in the .conf file. And a change to the code to enable that would be larger and more impactful (read risky).

The 1.8.1.2 version does not provide the container manager to use in the proxy scripts / .desktop files. And fails.

```
$ ./jq --help
Error response from daemon: No such container: debian-bookworm-dx
Create it now, out of image ? [Y/n]:
```

With the patch in place which adds DBX_CONTAINER_MANAGER="podman" (in this example) strategically if CONTAINER_ID is not set ...

```
$ ./jq --help
jq - commandline JSON processor [version 1.6]

Usage:	/usr/bin/jq [options] <jq filter> [file...]
	/usr/bin/jq [options] --args <jq filter> [strings...]
	/usr/bin/jq [options] --jsonargs <jq filter> [JSON_TEXTS...]

...
```

Additional context

I have written a [wrapper](https://github.com/klmcwhirter/oci-shared-images/tree/2-support-multiple-dbx_container_managers-on-the-same-system) around distrobox and devcontainers to ease the assembly of arbitrary development environments.

[...]

PR #1761, opened 2025-05-28
propagate DBX_CONTAINER_MANAGER through proxy scripts

klmcwhirter:
propagate DBX_CONTAINER_MANAGER through proxy scripts and .desktop files; support usage of multiple managers on a single host

This is the patch from #1758 .


dottorblaster:
Thanks! We don't plan to let users leverage multiple container runtimes.

Keep on rocking with distrobox! :heart:

PR #1795, opened 2025-07-15
refactor(distrobox-init): Refactor package manager detection logic an…

SekaiMoe:
…d remove microdnf-related code

1. **Refactor Package Manager Detection Logic**
- Replaced the previous `command -v` based detection of package managers with a more robust approach that reads the `/etc/os-release` file.
- This change ensures accurate identification of the operating system and its associated package manager by leveraging the `ID` or `NAME` fields in `/etc/os-release`.
- For example, distributions like Debian, Ubuntu, Arch, Fedora, CentOS, and others are now detected using `grep` on `/etc/os-release` instead of relying on the presence of specific binaries.

2. **Remove `setup_microdnf` Function**
- The `setup_microdnf` function was removed as it is no longer necessary.
- The logic for handling `microdnf` has been consolidated into the `setup_dnf` function, which now supports both `dnf` and `microdnf` use cases implicitly.
- This reduces redundancy and simplifies the codebase by eliminating a dedicated function for `microdnf`.

- **Improved Reliability**: The previous method of detecting package managers using `command -v` was prone to false positives or negatives if the binary was missing or misconfigured. Reading `/etc/os-release` provides a more reliable and standardized way to identify the operating system and its package manager.
- **Code Simplification**: Removing the `setup_microdnf` function and consolidating its logic into `setup_dnf` reduces code duplication and maintenance overhead.
- **Future-Proofing**: The new detection logic is more adaptable to future changes in package manager implementations or distribution-specific quirks.

- The changes improve the robustness and maintainability of the script while ensuring compatibility with a wide range of Linux distributions.
- Users of `microdnf`-based systems will continue to receive the same functionality through the updated `setup_dnf` implementation.

### DIFF
--- a/distrobox-init
+++ b/distrobox-init
@@ -1043,115 +1043,6 @@ EOF
 	fi
 }
 
-# setup_microdnf will upgrade or setup all packages for microdnf based systems.
-# Arguments:
-#   None
-# Expected global variables:
-#   upgrade: if we need to upgrade or not
-#   container_additional_packages: additional packages to install during this phase
-# Expected env variables:
-#   None
-# Outputs:
-#   None
-setup_microdnf()
-{
-	# If we need to upgrade, do it and exit, no further action required.
-	if [ "${upgrade}" -ne 0 ]; then
-		microdnf upgrade -y
-		exit
-	fi
-	# Check if shell_pkg is available in distro's repo. If not we
-	# fall back to bash, and we set the SHELL variable to bash so
-	# that it is set up correctly for the user.
-	if ! microdnf install -y "${shell_pkg}"; then
-		shell_pkg="bash"
-	fi
-	deps="
-		${shell_pkg}
-		bash-completion
-		bc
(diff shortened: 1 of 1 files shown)


SekaiMoe:
> Thanks! Unfortunately we're pretty satisfied with the current detection logic and with `setup_microdnf` itself.

No, this logic has a major issue.  
If Arch Linux has `apt` installed, it will be mistakenly identified as Debian or its derivatives, leading to incorrect steps being followed.

Additionally, there is [AOSC OS](https://aosc.io), which uses `.deb` packages but is not based on Debian. When installing distrobox, it would also lead to incorrect steps because it has both `apt` and `oma` package managers. However, `apt` is planned to be deprecated at some point in the future.

dottorblaster:
I didn't know about AOSC OS, perhaps in the future we'll reevaluate the approach we use to identify package managers. Thank you for the patch!

PR #1806, opened 2025-08-06
feat: add --unshare-home flag for isolated home directories

stone-w4tch3r:
# Add --unshare-home flag for isolated containers

## What this does

Adds a new `--unshare-home` flag that creates containers without mounting your host home directory. The container gets its own isolated home instead.

## How it works

- New flag in `distrobox create --unshare-home`
- Container gets labeled so `distrobox enter` knows it's unshared
- Home directory mounting is skipped during creation
- Container creates its own home directory on first entry
- `--unshare-all` now includes this flag automatically

## Why you'd want this

I've been working with AI agents and realized I needed better isolation. These tools can be unpredictable and might access files they shouldn't. With `--unshare-home`, I can run potentially dangerous code knowing it can't touch my personal files, configs, or SSH keys.

It's also useful for testing untrusted software or creating clean development environments.

## Testing

I'll attach detailed testing docs in the next message showing how to verify this works correctly across different distros and use cases.

### DIFF
--- a/PR_DESCRIPTION.md
+++ b/PR_DESCRIPTION.md
@@ -1,0 +1,23 @@
+# Add --unshare-home flag for isolated containers
+
+## What this does
+
+Adds a new `--unshare-home` flag that creates containers without mounting your host home directory. The container gets its own isolated home instead.
+
+## How it works
+
+- New flag in `distrobox create --unshare-home`
+- Container gets labeled so `distrobox enter` knows it's unshared
+- Home directory mounting is skipped during creation
+- Container creates its own home directory on first entry
+- `--unshare-all` now includes this flag automatically
+
+## Why you'd want this
+
+I've been working with AI agents and realized I needed better isolation. These tools can be unpredictable and might access files they shouldn't. With `--unshare-home`, I can run potentially dangerous code knowing it can't touch my personal files, configs, or SSH keys.
+
+It's also useful for testing untrusted software or creating clean development environments.
+
+## Testing
+
+I'll attach detailed testing docs in the next message showing how to verify this works correctly across different distros and use cases.
--- a/completions/zsh/_distrobox-create
+++ b/completions/zsh/_distrobox-create
@@ -23,6 +23,7 @@ _distrobox-create() {
         '--nvidia[Try to integrate host nVidia drivers into the guest]'
         '--unshare-devsys[Do not share host devices and sysfs dirs from host]'
         '--unshare-groups[Do not forward users additional groups into the container]'
+        '--unshare-home[Do not share host home directory with container]'
         '--unshare-ipc[Do not share ipc namespace with host]'
         '--unshare-netns[Do not share the net namespace with host]'
         '--unshare-process[Do not share process namespace with host]'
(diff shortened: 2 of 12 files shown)


89luca89:
Hi @stone-w4tch3r thanks for your work
to be honest i'm not really interested in having unshared homes, I'm ok with custom homes (like we have) but the point of distrobox is *integration* not *isolation*, and this is well explicit in the readme here: https://github.com/89luca89/distrobox?tab=readme-ov-file#security-implications

If you need more isolation, you might want to use raw podman or docker, and distrobox is not the right tool for the job

Thanks anyway for your work and effort

Cloudwalk9:
@89luca89

The issue at hand is fundamentally state isolation. This has nothing to do with security. This PR may not be the solution but a recent LWN article indicated you won't accept any patches that try to isolate $HOME and I hope this isn't the case.

The problem is, many tools (pip, pipx in particular, and npm, cargo, rustup, etc) read from and write to $HOME, especially under .config, .cache, .local... even go writes its own state. Distrobox by default allows container and host state in $HOME to comingle and there's lots of software that will transparently read and write to these directories and mutate state of the host and even other Distrobox containers, leading to subtle breakage, non-reproducible environments, and unnecessary headache.

This design has a "smell" to it. When I create a new container, I don't expect perfect isolation, because indeed that's not the purpose of Distrobox. It's meant to integrate with the host. But I expect container state to be fresh.

What we have now looks nice and seems to work until you deviate from the happy path or run into an edge case. The solution could be to isolate the XDG standard directories. Using --home is not only unwieldy but is an all or nothing solution when one perhaps doesn't want it to be. And it doesn't need to be.

There is no slippery slope. The problem is bounded and the solution is not at odds with host integration.

stone-w4tch3r:
For me, the code in this PR actually was too unstable. When using --unshare-home + trying to mount only folders I needed, eg `~/my-git/project`, `~/.claude`, `~/.other-important-config/`, docker container was occasionally exploding during complex manipulations, eg when trying to install system-level dependencies like `flatpak-builder` into it, due to 15+ overlapping mounts in summary

So it's kinda good this code not reached public 😅

For the described home pollution case, I use --home, it's usually enough. But it's sad that this is not default option, because any distrobox creation without --home specified results in complex $HOME pollution and even potentially breakage, in case of eg python

Anyway, I think modern problems like dev AI isolation are not well aligned with current course of distrobox. And sh-based implementaiton is, ugh, bug-prone, to say at least

I think we may need new distrobox implementation 😁, probably python based, to avoid multiple sh related bugs like string escaping + maybe introduce plugins arch. I have some ideas on trying to port distrobox to python (with minimized AI usage to avoid issues), but this is too much work for me alone rn

salim-b:
> I think we may need new distrobox implementation 😁, probably python based, to avoid multiple sh related bugs like string escaping + maybe introduce plugins arch. I have some ideas on trying to port distrobox to python (with minimized AI usage to avoid issues), but this is too much work for me alone rn

You might be interested in [Litterbox](https://litterbox.work/) that was recently [introduced here](https://github.com/89luca89/distrobox/issues/28#issuecomment-3577424406). It's written in Rust and appears promising. I didn't test (or need) it yet, though.

89luca89:
> A compromise would be to set a default "prefix" directory owned by Distrobox like .wine and $WINEPREFIX, so --home could be passed without parameters and Distrobox could make one by itself. That would make it 10x more wieldy.

Like this you mean? https://github.com/89luca89/distrobox/blob/91d2d77106e41bafdebb0b347eb5716822272217/distrobox-create#L28

PR #1807, opened 2025-08-07
docs: Remove --name argument from distrobox-stop usage docs

brennenputh:
Noticed this while doing some completions for Fish.  Looks like the `--name` argument was removed, but this usage example stayed the same.


dottorblaster (approved):
Thanks!

PR #1811, opened 2025-08-14
enter: Allow whitespaces in environment variables

scosu:
Replace whitespaces with escaped whitespaces and switch to using sed to prepend --env to the env variables.


scosu:
Not sure how this PR can cause the markdown linter issue as I am not touching any markdown here. Any hints would be appreciated.

dottorblaster:
@scosu your changes are unrelated. Can you rebase the PR so the markdown linter doesn't complain anymore? Thanks!

PR #1825, opened 2025-08-24
docs: update tips for using custom SHELL in assemble, Fix #1633

### LINKED ISSUE #1633: [Suggestion] Make shell configurable in manifest
Not that this is a problem in itself since it is possible to determine the shell using the `SHELL` environment variable, but it would be interesting if this could be configured in the manifest.

In the manifest you can put `shell=/bin/bash` (for example) and when you run `distrobox assemble` it uses the indicated shell.

The current alternative is to use the environment variable, but that is not comfortable in a command like `assemble`. Also, if with `assemble` you are building two different containers and you want one with one shell and one with a different shell, you can't do it.

yocoldle:


dottorblaster (approved):
LGTM

PR #1826, opened 2025-08-24
init: setup correct chimera base-full-man dep

JamiKettunen:
[`base-core-man`](https://github.com/chimera-linux/cports/blob/8f4bf92/main/base-full/template.py#L120-L121) has been transitional provider of this for a while and will eventually be dropped; I must've still copied this over from a much earlier version of the #1778 patch when I was experimenting long ago

### DIFF
--- a/distrobox-init
+++ b/distrobox-init
@@ -669,7 +669,7 @@ setup_apk()
 		apk add chimera-repo-user
 		apk update
 		deps="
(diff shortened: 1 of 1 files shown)


89luca89:
Thanks, for now I'm going to skip chimera (and doas distros) until we find the culprit for this discrepancy in CI

PR #1833, opened 2025-09-06
Add Hyprland setup instructions to Distrobox guide

AbduEhab:
Added instructions for running the latest Hyprland session in Distrobox, including container creation and environment variable setup. I did not add any images, as I am unsure if they are needed.


AbduEhab:
Can the file name be changed to include Hyprland?

PR #1836, opened 2025-09-16
init: exclude schroot mountpoint from socket search, improve performance.

KarlK90:
This fixes minute long start-up hangs when using distrobox on a host that has active schroot mounts, which I saw on my machine.


dottorblaster (approved):
Thanks! ❤️

PR #1858, opened 2025-10-19
enter: don't filter environment variables with special characters

### LINKED ISSUE #1842: distrobox-enter: envs with `"` are not present in container
**Describe the bug**
If there's a `"` in an environment variable (e.g. a JSON string), `distrobox-enter` won't make it available in the container. Other envs are available.

**To Reproduce**
```bash
$ GOOD='yes' BAD='{"foo":1}' distrobox enter -- sh -c 'echo GOOD=$GOOD BAD=$BAD'
GOOD=yes BAD=
```

**Expected behavior**
All envs are available in the container.

I think `docker exec` supports only set env names and it will read the values: `--env=ENV`, not need to set env value in the command line `--env=ENV=value`. If we don't set the value, this filter can be removed:

https://github.com/89luca89/distrobox/blob/3f2ae51a32540434376ec8c43cb8636eb347e2a1/distrobox-enter#L439

**Logs**
Run the commands with `--verbose` and post the log here as a file upload
Attach also the output of `podman logs` or `docker logs`, possibly with `--latest` flag

- Which version or podman, docker or lilipod?
- Which version of distrobox?
- Which host distribution?

**Additional context**

cshuaimin:
Fix #1842

Docker and Podman support only specifying env names, but seems lilipod doesn't. I'm not sure how to deal with lilipod, should I add an if condition?

### DIFF
--- a/distrobox-enter
+++ b/distrobox-enter
@@ -433,14 +433,13 @@ generate_enter_command()
 	# and export them to the container.
 	set +o xtrace
 	# disable logging for this snippet, or it will be too talkative.
-	# We filter the environment so that we do not have strange variables or
-	# multiline.
 	# We also NEED to ignore the HOME variable, as this is set at create time
 	# and needs to stay that way to use custom home dirs. or it will be too talkative.
+	# We only pass variable names to --env flag to avoid issues with special characters in variable values.
 	result_command="${result_command}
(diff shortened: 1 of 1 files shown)


dottorblaster:
@cshuaimin it looks like the formatter is failing, could you fix it? :heart:

PR #1860, opened 2025-10-21
feat: extend manifest definitions with include

balanza:
This is a proposal to add an `include` option to the `assemble` command. This option allows for definitions to be reused and extended by other definitions **inside the same manifest file**.

The feature works on top of the current `assemble` implementation by creating a temporary copy of the manifest file in which every occurrence of `include` is replaced with the actual definition it points to. It does not touch the current parsing and execution algorithm.

### Example
The following manifest file
```ini
[ubuntu]
image=ubuntu:latest
additional_packages="git vim tmux nodejs"
additional_packages="htop iftop iotop"
additional_packages="zsh fish"

[ubuntu-nvidia]
include=ubuntu
nvidia=true
```
is equivalent to 

```ini
[ubuntu]
image=ubuntu:latest
additional_packages="git vim tmux nodejs"
additional_packages="htop iftop iotop"
additional_packages="zsh fish"

[ubuntu-nvidia]
image=ubuntu:latest
additional_packages="git vim tmux nodejs"
additional_packages="htop iftop iotop"
additional_packages="zsh fish"
nvidia=true
```

and they both produce 
```sh
> distrobox assemble create --dry-run

 - Creating ubuntu...
distrobox create --yes --name ubuntu --image ubuntu:latest --additional-packages " git vim tmux nodejs htop iftop iotop zsh fish"
 - Creating ubuntu-nvidia...
distrobox create --yes --name ubuntu-nvidia --image ubuntu:latest --nvidia --additional-packages " git vim tmux nodejs htop iftop iotop zsh fish"
```

### Implementation
Given the input manifest file (whose detection is left untouched), this feature:
* creates a temporary copy of the file
* scans the copy line-by-line, top-to-bottom
* when `include` attribute is found, the line is replaced with the lines from the given section in the original input manifest
  * if the section is not found, an error is raised
  * if a circular reference is detected, an error is raised
* the copied file is then forwarded to the actual parser for execution

### Caveats
* Duplicated options in the same section are concatenated in lists; this is the current `distrobox` behaviour
```ini
[ubuntu]
image=ubuntu:latest
root=false
root=true
# parsed as: 
#  root=0¤1
# this will break the create command
```
IMHO, this is currently not a big deal for users. However, with the introduction of `include`, the following might be a legit usage:
```ini
[ubuntu]
image=ubuntu:latest
root=false

[ubuntu-root]
root=true
```

To fix this, an intervention on the current `assemble` implementation is needed.

### DIFF
--- a/docs/usage/distrobox-assemble.md
+++ b/docs/usage/distrobox-assemble.md
@@ -221,7 +221,7 @@ each option corresponds to one of the `create` flags.
 	image=ubuntu:latest
 	pre_init_hooks="export SHELL=/bin/bash;"
 
-**Include example**
+**Include example (inherit fields from another distrobox)**
 
 	[ubuntu]
 	image=ubuntu:latest


docs/usage/distrobox-assemble.md
```diff
@@ -214,3 +220,15 @@ each option corresponds to one of the `create` flags.
 	[ubuntu]
 	image=ubuntu:latest
 	pre_init_hooks="export SHELL=/bin/bash;"
+
+**Include example**
```
dottorblaster:
```suggestion
**Include example (inherit fields from another distrobox)**
```

dottorblaster: Yes, I can see where you're coming from and for me it's better than inherit as a keyword. I just wanted a less concise title, sorry for being pedantic!

PR #1873, opened 2025-11-06
docs: update command for GNOME installation

amnexya:
The current command used to install GNOME inside of Distrobox is incorrect, as the dnf command as well as the group name have changed.

Running `dnf group list --hidden` shows GNOME listed under `gnome-desktop`, which I have updated the group name to.

Also `groupinstall` was not a recognised argument in dnf.


dottorblaster (approved):
Thanks!

PR #1881, opened 2025-11-14
docs: note that Podman and Distrobox are now pre-installed on SteamOS 3.5+

kavishgr:
Updated SteamOS compatibility information to reflect that Podman and Distrobox are pre-installed starting with SteamOS 3.5.


dottorblaster (approved):
Thank you!

PR #1883, opened 2025-11-15
use doas and doas-sudo-shim in alpine

adamthiede:
Alpine defaults to 'doas' and the 'doas-sudo-shim' package is more than just a symlink. Since these are distro defaults, and most users won't notice a difference, this should be preferred.


dottorblaster:
Closing this for the moment. Feel free to reopen if you find another viable approach :+1:

Issue #1906, opened 2025-11-18
[Error] distrobox-init: Logic error in rootful check causes "stat: cannot read file system information" on Linux


john-holt4:
**Description**

I am encountering an error message when updating/entering containers: `stat: cannot read file system information for '%u': No such file or directory.`

After investigating `distrobox-init`, it appears to be a logic flaw in the cross-platform check for root permissions. The script attempts to use a lazy `||` (OR) operator to support both Linux (`stat -c`) and BSD/macOS (`stat -f`).

If the container is running as a non-root user (e.g., UID 1000), the first check returns False (because `1000 != 0`). Consequently, the script executes the right side of the OR statement (`stat -f`). On Linux, `stat -f` interprets `%u` as a filename rather than a format string, causing the crash.

**To Reproduce**

1. Run a standard rootless Distrobox container on a Linux host.
2. Enter the container (`distrobox enter ...`).
3. Observe the error output during the initialization phase.

**Logs/Output**

`stat: cannot read file system information for '%u': No such file or directory`

**Technical Analysis**

The issue lies in this logic block within `distrobox-init`:
```
if stat /run/host/etc/shadow > /dev/null &&
    {
        [ "$(stat -c "%u" /run/host/etc/shadow)" = "0" ] || [ "$(stat -f "%u" /run/host/etc/shadow)" = "0" ]
    } &&
```

**The logic flow causing the bug:**

1. `stat -c "%u" ...` runs on Linux. It returns the UID (e.g., `1000`).
2. The test `[ "1000" = "0" ]` evaluates to False.
3. Because of the `||` operator, the shell executes the second command to see if it returns True.
4. `stat -f "%u" ...` runs. On Linux (GNU coreutils), `-f` means "file system status," and it expects a file path. It tries to treat `%u` as the file path.
5. `stat` exits with an error because the file `%u` does not exist.

**Suggested Fix**

The errors from the BSD/macOS check should be suppressed so they don't leak into the console or cause script failures on Linux hosts.

**Diff:**

```
- [ "$(stat -c "%u" /run/host/etc/shadow)" = "0" ] || [ "$(stat -f "%u" /run/host/etc/shadow)" = "0" ]
+ [ "$(stat -c "%u" /run/host/etc/shadow)" = "0" ] || [ "$(stat -f "%u" /run/host/etc/shadow 2>/dev/null)" = "0" ]
```

**Environment**

- **Distrobox Version:** 1.8.2.1
[...]

PR #1907, opened 2025-11-19
init: fix stat syntax error in rootful detection

jokokucing:
## Description
This fixes a logic bug that causes the error `stat: cannot read file system information for '%u'` to appear when initializing or entering a rootless container on Linux.

## The Issue
The rootful detection logic uses a lazy `||` operator to support both Linux (`stat -c`) and BSD/macOS (`stat -f`).

```bash
[ "$(stat -c "%u" ...)" = "0" ] || [ "$(stat -f "%u" ...)" = "0" ]
```

If a user is on Linux but is **not** root (e.g., UID 1000), the first check returns `False` (because 1000 != 0). Consequently, the shell executes the second command (the BSD syntax fallback).

On GNU `stat` (Linux), the `-f` flag stands for "file system status," not format. It interprets `%u` as a filename, fails to find it, and prints an error to stderr.

## The Fix

I added `2>/dev/null` to the BSD fallback command.

- On BSD/macOS: The command succeeds and prints the UID to stdout (logic holds).
- On Linux: The command fails, but the error message is suppressed, preventing the "No such file or directory" noise in the logs.

## Reference
A PR for this [bug](https://github.com/89luca89/distrobox/issues/1906)

### DIFF
--- a/distrobox-init
+++ b/distrobox-init
@@ -235,7 +235,7 @@ fi
 # a good thing, users behold!
 if stat /run/host/etc/shadow > /dev/null &&
 	{
-	        [ "$(stat -c "%u" /run/host/etc/shadow)" = "0" ] || [ "$(stat -f "%u" /run/host/etc/shadow 2>/dev/null)" = "0" ]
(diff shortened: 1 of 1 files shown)


dottorblaster (approved):
LGTM

PR #1916, opened 2025-11-23
feat: set compopt filenames for --file in bash completion

yedayak:
Currently it treats directories as filenames, so it puts a space after them, forcing you to replace the space with a slash to get completions in it.


dottorblaster (approved):
Thanks!

PR #1925, opened 2025-11-27
fix: unintended print to stdout in assemble command

balanza:
The `resolve_includes` function _returns_ the used temporary file path. As the value is not used, it should be redirected to /dev/null to retain it to print on screen.

To reproduce, just `distrobox assemble create` against any manifest file. 

```sh
> distrobox assemble create --file manifest.ini
/tmp/tmp.BmKsHG5Kxy
# ^^^^^ unintended print
```

### DIFF
--- a/distrobox-assemble
+++ b/distrobox-assemble
@@ -789,5 +789,5 @@ replace_line()
 
 # Exec
 expanded_file=$(mktemp -u)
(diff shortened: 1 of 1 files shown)


dottorblaster (approved):
LGTM

Issue #1940, opened 2025-12-19
[Error] `distrobox assemble create` fails to override boolean flags from false->true when "include" key specified


givensuman:
**Describe the bug**
The "include" keyword has incorrect behavior regarding boolean values. It does a good job accumulating keys like "additional_packages", but boolean keys like "root" or "nvidia" are mishandled.

**To Reproduce**
```bash
cat >> bug.ini << EOF
[ubuntu]
image=ubuntu:latest
root=true
nvidia=false
additional_packages="zsh"

[ubuntu-nvidia]
include=ubuntu
root=false
nvidia=true
additional_packages="htop iotop"
EOF
```
then run `distrobox assemble create --dry-run --file bug.ini`. You get an output like this:

```
/usr/bin/distrobox-create --yes --name ubuntu --image ubuntu:latest --root --additional-packages " zsh"
/usr/bin/distrobox-create --yes --name ubuntu-nvidia --image ubuntu:latest --additional-packages " zsh htop iotop"
```

Note the `--root` flag is switched off, but `--nvidia` isn't switch on.

**Expected behavior**

```
/usr/bin/distrobox-create --yes --name ubuntu --image ubuntu:latest --root --additional-packages " zsh"
/usr/bin/distrobox-create --yes --name ubuntu-nvidia --image ubuntu:latest --nvidia --additional-packages " zsh htop iotop"
```

**Logs**
Run the commands with `--verbose` and post the log here as a file upload
Attach also the output of `podman logs` or `docker logs`, possibly with `--latest` flag

**Desktop (please complete the following information):**

- Are you using podman, docker or lilipod?
`podman`
- Which version or podman, docker or lilipod?
`podman version 5.7.1`
- Which version of distrobox?
distrobox: 1.8.2.2
- Which host distribution?
`ublue-main`
- How did you install distrobox?
Included by default

**Additional context**
Also worth noting the leading space in `--additional-packages` is also probably unintentional.

PR #1941, opened 2025-12-19
fix: boolean overrides with assemble's include key

givensuman:
Fixes https://github.com/89luca89/distrobox/issues/1940.

The issue arose from handling of cumulative keys like
`additional_packages` or `additional_flags` being applied to boolean
flags as well.

Also includes a small patch for the leading spaces in `--additional-packages` output's leading space mentioned in #1940.

### Testing

```bash
cat >> bug.ini << EOF
[ubuntu]
image=ubuntu:latest
root=true
nvidia=false
additional_packages="zsh"

[ubuntu-nvidia]
include=ubuntu
root=false
nvidia=true
additional_packages="htop iotop"
EOF
```

then run `./distrobox-assemble create --dry-run --file bug.ini`. This satisfies expected output:

```bash
./distrobox-create --yes --name ubuntu --image ubuntu:latest --root --additional-packages zsh
./distrobox-create --yes --name ubuntu-nvidia --image ubuntu:latest --nvidia --additional-packages "zsh htop iotop"
```


dottorblaster:
I don't know what happened but it looks like you deleted the whole `distrobox-assemble` file. :slightly_smiling_face:

PR #1965, opened 2026-01-16
fix(init): fix support for newer shadow strictier checks

### LINKED ISSUE #1962: [Error] debian:unstable failed on setting up user (assemble)
Please, before opening a bug:

- make sure you've read the documentation.
- Ensure there isn't already an open issue about this.
- Ensure there isn't already a closed/resolved issue about this.

**Describe the bug**
`debian:unstable` image setup is failing on setting up user. Creation using `assemble`.

**To Reproduce**
I've tried to create the container with this `distrobox.ini` (assemble):

```ini
[debian_test]

#image=quay.io/toolbx-images/debian-toolbox:unstable
image=docker.io/debian:unstable

pull=false
init=false
entry=false
start_now=true
replace=false
root=false

unshare_ipc=false
unshare_netns=false
unshare_process=false
unshare_devsys=false

nvidia=false
```

Both `docker.io/debian:unstable` and `quay.io/toolbx-images/debian-toolbox:unstable` have been tried with the same result.

**Expected behavior**
The container creation should work successfully, like `debian:stable` does.

**Logs**
Run the commands with `--verbose` and post the log here as a file upload
Attach also the output of `podman logs` or `docker logs`, possibly with `--latest` flag

[distrobox.log](https://github.com/user-attachments/files/24619983/distrobox.log)

[debian_test.log](https://github.com/user-attachments/files/24619985/debian_test.log)

- Are you using podman, docker or lilipod? _podman_
- Which version or podman, docker or lilipod? _5.7.1_
- Which version of distrobox? _1.8.2.2_
- Which host distribution? _Arch Linux_
- How did you install distrobox? _pacman_

**Additional context**
`debian:stable` image works flawlessly with the same setup.

89luca89:
Newer versions of shadow-utils introduced stricter validation that breaks previous assumptions:

- usermod now errors with "user already exists" or "UID already exists" when setting --login or --uid to the same value the user already has
- chpasswd -e now validates the password hash format and rejects plain strings that aren't valid crypt hashes

Changes:
- Remove --uid and --login from the main usermod call
- Add separate conditional checks to only modify username/UID when they actually differ from current values
- Remove -e flag from chpasswd calls to let it handle hashing internally
- Use passwd -d to remove passwords, with fallback to chpasswd for older systems that don't support it

This maintains backward compatibility with older distributions while fixing breakage on systems with newer shadow-utils.

Fix #1962

### DIFF
--- a/distrobox-init
+++ b/distrobox-init
@@ -2430,9 +2430,7 @@ elif [ ! -e /etc/passwd.done ]; then
 		--home "${container_user_home}" \
 		--shell "${SHELL:-"/bin/bash"}" \
 		--groups "${additional_groups}" \
-		--uid "${container_user_uid}" \
 		--gid "${container_user_gid}" \
-		--login "${container_user_name}" \
 		"${user_to_modify:-"${container_user_name}"}"; then
 
 		printf "Warning: There was a problem setting up the user with usermod, trying manual addition\n"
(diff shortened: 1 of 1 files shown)


dottorblaster (approved):
LGTM

PR #1989, opened 2026-01-29
feat: add ALT Linux compatibility improvements

liannnix:
This PR fixes ALT Linux container compatibility issues, specifically addressing problems with `--init` containers.

### Problem
ALT Linux ships its own `su` implementation that is incompatible with util-linux `su` flags. When `distrobox-enter` runs with `unshare_groups` enabled (i.e., `--init` containers), it passes flags like `-m`, `--pty`, and `-c` that ALT's `su` rejects, causing immediate failure:

```
/bin/su: invalid option -- 'm'
usage: su [-|-l] [-c "command"] [-s "shell"] [username]
```

### Changes Made

1. **ALT Linux-specific `su` wrapper** in `distrobox-init`:
   - Inside the existing ALT Linux detection block (`if command -v control`), creates `/usr/local/bin/su` → `/usr/sbin/runuser` wrapper
   - `runuser` (from `util-linux`, always present in ALT) accepts the same flags as util-linux `su`
   - Wrapper is placed in `/usr/local/bin` to survive `apt-get` upgrades (no package owns it)

2. **Guarded the generic `--pty` wrapper**:
   - Added `[ ! -e /usr/local/bin/su ]` check to prevent overwriting ALT-specific wrapper
   - Maintains compatibility with old util-linux versions and non-util-linux `su` implementations

3. **PATH ordering fix** in `distrobox-enter`:
   - Ensures `/usr/local/bin` precedes `/usr/bin` when host PATH is passed through
   - Follows FHS conventions (`/usr/local` should override `/usr`)
   - Guarantees wrapper is found first regardless of host PATH ordering

4. **Additional ALT Linux fixes**:
   - `control pam_mktemp disabled` – PAM stack compatibility fix
   - `mkdir -p` for TCB directory – prevents errors when directory exists

### Why a Separate Wrapper?
The existing generic wrapper only strips `--pty` flag for old util-linux versions. ALT Linux `su` has broader incompatibilities – it rejects `-m`, `-c` flags entirely. A simple `--pty`-stripping wrapper won't help; we need a complete replacement that understands all util-linux `su` flags.

### Solution Characteristics
- **Transparent**: No new flags, env vars, or user options
- **Upgrade-safe**: Wrapper in `/usr/local/bin` survives package updates
- **Distro-specific**: Only affects ALT Linux containers (detected via `control`)
- **Aligned with precedent**: Follows same pattern as Alpine/Chimera (`su-exec`/`doas` replacements)

The fix ensures ALT Linux containers work seamlessly with `--init`, maintaining distrobox's POSIX compliance while handling distro-specific incompatibilities internally.

### DIFF
--- a/distrobox-enter
+++ b/distrobox-enter
@@ -474,6 +474,42 @@ generate_enter_command()
 		else
 			container_paths="${PATH}"
 		fi
+
+		# Ensure /usr/local/{s,}bin appears before /usr/{s,}bin in PATH.
+		# This follows FHS conventions: /usr/local should override /usr,
+		# and ensures distrobox wrappers in /usr/local/bin are found first.
+		reordered_paths=""
+		IFS_OLD="${IFS}"
+		IFS=":"
+		for p in ${container_paths}; do
+			case "${p}" in
+				/usr/local/bin | /usr/local/sbin)
+					# skip, will be re-inserted before their /usr counterpart
+					;;
+				/usr/bin)
+					# insert /usr/local/bin right before /usr/bin
+					reordered_paths="${reordered_paths:+${reordered_paths}:}/usr/local/bin:${p}"
+					;;
+				/usr/sbin)
+					# insert /usr/local/sbin right before /usr/sbin
+					reordered_paths="${reordered_paths:+${reordered_paths}:}/usr/local/sbin:${p}"
+					;;
+				*)
+					reordered_paths="${reordered_paths:+${reordered_paths}:}${p}"
+					;;
+			esac
+		done
+		IFS="${IFS_OLD}"
+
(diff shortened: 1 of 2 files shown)


distrobox-enter
```diff
@@ -474,6 +474,42 @@ generate_enter_command()
 		else
 			container_paths="${PATH}"
 		fi
+
+		# Ensure /usr/local/{s,}bin appears before /usr/{s,}bin in PATH.
+		# This follows FHS conventions: /usr/local should override /usr,
+		# and ensures distrobox wrappers in /usr/local/bin are found first.
```
dannyhpy: Couldn't this be ensured by modifying the above lines to *prepend* `standard_paths` to `container_paths` instead of *appending*?

PR #2001, opened 2026-02-18
feat: add distrobox-snapshot subcommand

yhekma:
# Summary
- Add a new `distrobox snapshot` subcommand that wraps `podman/docker container commit` to create point-in-time snapshots of distrobox containers
- Support creating, listing, exporting, and removing snapshots for both running and stopped containers
- Include bash/zsh shell completions, a man page with restore instructions, and a guard that rejects lilipod

# Motivation
Saving and restoring container state currently requires users to manually run `podman container commit`, `podman save`, `podman load`, etc. as documented in `docs/useful_tips.md`. This wraps that workflow into a first-class subcommand that follows all existing distrobox conventions.

# Usage
## Create a snapshot (auto-generated timestamp tag)
distrobox snapshot my-container

## Create a snapshot with a custom tag
distrobox snapshot --name my-container --tag before-upgrade

## List snapshots for a container
distrobox snapshot --name my-container --list

## Export a snapshot to a tar archive
distrobox snapshot --name my-container --save --tag before-upgrade --output backup.tar

## Remove a snapshot
distrobox snapshot --name my-container --remove --tag before-upgrade

## Restore from a snapshot (manual, documented in man page)
distrobox rm my-container
distrobox create --image distrobox-snapshot/my-container:before-upgrade --name my-container

# Changes
| File                                | Change                                      |
| ----------------------------------- | ------------------------------------------- |
| distrobox-snapshot                  | New subcommand script (483 lines)           |
| distrobox                           | Added snapshot to dispatcher and help text  |
| completions/bash/distrobox-snapshot | New bash completion                         |
| completions/zsh/_distrobox-snapshot | New zsh completion                          |
| completions/zsh/_distrobox          | Added snapshot to command list and dispatch |
| man/man1/distrobox-snapshot.1       | New man page with restore instructions      |

# Compatibility
- Works with podman, podman-launcher, and docker
- lilipod: prints an error and exits immediately (not supported)
- Containers can be snapshotted while running (uses -p to pause during commit for filesystem consistency) or while stopped

# Testing
## Manually verified with podman 5.7.1 against both running and stopped containers:
- Creating snapshots (auto-tag and custom tag)
- Listing snapshots
- Saving snapshots to tar archive
- Removing snapshots (interactive and non-interactive)
- Positional container name argument
- Error cases: nonexistent container, missing --tag, invalid flags, nonexistent snapshot tag
- lilipod rejection via DBX_CONTAINER_MANAGER=lilipod

# Disclaimer
Note I have used AI to write part of this PR. I have manually gone through the changes and ran tests with docker and podman, not with lilipod since I did not have a setup to test with. To be sure nothing unexpected would be executed I opted to just not add support and error out.

### DIFF
--- a/completions/bash/distrobox-snapshot
+++ b/completions/bash/distrobox-snapshot
@@ -1,0 +1,13 @@
+# shellcheck disable=all
+
+if [ -e /usr/share/bash-completion/completions/distrobox ]; then
+	source /usr/share/bash-completion/completions/distrobox
+fi
+if [ -e /usr/local/share/bash-completion/completions/distrobox ]; then
+	source /usr/local/share/bash-completion/completions/distrobox
+fi
+if [ -e "${HOME}/.local/share/bash-completion/completions/distrobox" ]; then
+	source "${HOME}/.local/share/bash-completion/completions/distrobox"
+fi
+
+complete -F _generate_from_help distrobox-snapshot
--- a/completions/zsh/_distrobox
+++ b/completions/zsh/_distrobox
@@ -12,6 +12,7 @@ _distrobox_commands() {
     'list:List all distrobox containers'
     'ls:Alias for list'
     'rm:Remove a distrobox container'
+    'snapshot:Snapshot a distrobox container'
     'stop:Stop a running distrobox container'
     'upgrade:Upgrade a distrobox container'
     'ephemeral:Create a temporary distrobox container'
@@ -42,6 +43,9 @@ case $state in
         ;;
       rm)
         _distrobox-rm
+        ;;
+      snapshot)
+        _distrobox-snapshot
         ;;
       stop)
         _distrobox-stop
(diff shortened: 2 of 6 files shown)


ScatteredCognition:
Hello, I would like to say that I found this PR as I had the exact same idea in mind.
However, as distrobox is going through a rewrite in go, I guess it will take some time until new functionality like this can be properly implemented.

Although my knowledge of go is limited, once the distrobox go port is done, I would be more than happy to implement such backup/restore/snapshot functionality.

I think the "Export a snapshot to a tar archive" part would be better served via a separate command, something like `distrobox backup/restore``

In my personal opinion, there should be a built-in way for people to easily backup and restore their distrobox containers, as offline backups are also a good way for digital preservation of content, while it will also be tremendously helpful for people who lack good internet connections.

PR #2009, opened 2026-02-27
docs: update README with sandboxing alternatives

Gerharddc:
Added information about alternatives to Distrobox with sandboxing capabilities.


dottorblaster (changes requested):
Just a tiny change and then we can merge 👍

PR #2041, opened 2026-03-31
refactor: load config as value object

balanza:
Refactor the way we handle configuration so that, instead of collapsing the file values into the environment, the entire configuration is calculated and passed as a value object to the commands.

This allows the default container image and name to be configurable (see: 24e9ed8d831a709fc009391f8981e0fa2aa33dfc).

**Expected benefits**
* Defaults, source config files and handled environment variables are now in the same file (`pkg/config/config.go`) 
* The CLI layer is not responsible for reading values from different sources.
* Remove dependency on `godotenv`
* The config values are passed to the command layer by value; callers can decide how to calculate the value set (it can be just the default values, reading from the default files, or a custom datasource)

**What changed**
* package `internal/config` moved to `pkg/config`
* config value trickled down from main to the command layer

Fixes #2040 (the fix: 24e9ed8d831a709fc009391f8981e0fa2aa33dfc)


pkg/config/config.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+func readEnv() map[string]string {
+	envConfig := make(map[string]string)
+
+	if value, exists := os.LookupEnv("DBX_CONTAINER_MANAGER"); exists {
+		envConfig["container_manager"] = value
+	}
+	if value, exists := os.LookupEnv("DBX_SUDO_COMMAND"); exists {
+		envConfig["sudo_command"] = value
```
dottorblaster:
```suggestion
		envConfig["sudo_program"] = value
```

pkg/config/config.go
```diff
@@ -0,0 +1,146 @@
...
+			merged[k] = v
+		}
+	}
+	return merged
+}
+
+func readConfigFile(filePath string) (map[string]string, error) {
+	cfg, err := ini.Load(filePath)
+	if err != nil {
+		if os.IsNotExist(err) {
```
dottorblaster, line 122: I'm not sure `ini` returns `ErrNotExists` in this case

pkg/config/config.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+var defaultsMap = map[string]string{ //nolint:gochecknoglobals // config defaults
+	"container_manager": "podman",
+	"sudo_program":      "sudo",
+	"verbose":           "false",
+	// container_image Fedora toolbox is a sensitive default
+	"container_image": "registry.fedoraproject.org/fedora-toolbox:latest",
+	"container_name":  "my-distrobox",
+}
```
dottorblaster: Making a function out of this would be nice

pkg/config/config.go
```diff
@@ -0,0 +1,146 @@
...
+
+func mergeConfigMaps(maps ...map[string]string) map[string]string {
+	merged := make(map[string]string)
+	for _, m := range maps {
+		for k, v := range m {
+			merged[k] = v
+		}
+	}
+	return merged
+}
```
dottorblaster: "Highest priority first" (user's `~/.distroboxrc` is first in the list), but `mergeConfigMaps` uses last-wins semantics. This means the lowest-priority system file (`selfDir/../share/distrobox/distrobox.conf`) actually overrides the user's `~/.distroboxrc`.

Or at least this is what I understood 😄 can you double check this works?

dottorblaster (changes requested):
Just some feedbacks but great work overall (also thank you for addressing this ❤️ )

pkg/config/config.go
balanza: I think it's covered by this test https://github.com/balanza/distrobox/blob/7ed1ebdc61996877e8c7f4d36c98213b5f1538d3/pkg/config/config_internal_test.go#L97

Anyway, I doublechecked. Here you can see that `ini` just propagates the `ErrNotExists` when `loose==false` (which is the case of `ini.Load`) https://github.com/balanza/distrobox/blob/7ed1ebdc61996877e8c7f4d36c98213b5f1538d3/pkg/config/config_internal_test.go#L97

pkg/config/config.go
balanza: you are right. I reverted the order. Thanks!

dottorblaster (approved):
Let's go!

PR #2047, opened 2026-04-01
fix: various bug fixes and code improvements

mirkobrombin:
Fixes a few issues found during code review:

- fix typo in struct field `ContanerID`
- fix `--rm-home` not working due to `ContainerHome` was never passed to `RmOptions`
- fix potential logic panic if podman retursn a container with no names (might happen?)
- fix `nil` pointer dereference on `prompter` in create cmd when the image is not found locally
- fix `uid` and `gid` trailing newline when reading from exec fallback
- moved shared types and constants that were used by both docker and podman to a common file
- skipped /dev/shm mount if `EvalSymlinks` fails instead of adding an empty path (was this on purpose?)


89luca89 (changes requested):
one nit, lgtm else

balanza (changes requested):
@mirkobrombin thanks a lot for the PR! Some of these changes are good to go, some others require some more attention.  

If you want, you can create another PR including the following 3 commits, so we can merge straight away

* cb89116e1ccff6d740655039655b7b1503d082a4 - good ✔️ 
* 66579683d5f6f3bd12898ebcdcd4d71f318e3a8c - good ✔️ 
* 0dabf6ac658c2ed684626462d872e6542eff8ad0 - good ✔️

mirkobrombin:
> @mirkobrombin thanks a lot for the PR! Some of these changes are good to go, some others require some more attention.
> 
> If you want, you can create another PR including the following 3 commits, so we can merge straight away
> 
> * [cb89116](https://github.com/89luca89/distrobox/commit/cb89116e1ccff6d740655039655b7b1503d082a4) - good ✔️
> * [6657968](https://github.com/89luca89/distrobox/commit/66579683d5f6f3bd12898ebcdcd4d71f318e3a8c) - good ✔️
> * [0dabf6a](https://github.com/89luca89/distrobox/commit/0dabf6ac658c2ed684626462d872e6542eff8ad0) - good ✔️

Happy to help! I am not in a rush, so I can wait for the full review.

pkg/containermanager/providers/podman.go
```diff
@@ -615,10 +615,15 @@ func parsePodmanContainerList(output string) ([]containermanager.Container, erro
 			id = id[:containerIDMaxLength]
 		}
 
+		name := id
```
balanza: **issue:** ‏The fallback to `id` is a bit too arbitrary; I'm afraid it may mislead users in case of errors. In the end, the event of having an empty name is very unlikely, if not impossible, as the field is defined strictly non-empty in the podman codebase https://github.com/containers/podman/blob/e1360b6f594fd66158bd5cd9fa002521559eacfc/libpod/container_config.go#L42.

Bugs can happen anyway, and I like the fact that we are protecting the application. I'd rather use an empty string, though. It seems to me that a user will notice more an empty string than a pseudorandom value. 

What do you think?

pkg/commands/create.go
```diff
@@ -84,11 +84,12 @@ type CreateOptions struct {
 	NonInteractive      bool
 }
 
-func NewCreateCommand(cm containermanager.ContainerManager, progress *ui.Progress) *CreateCommand {
```
balanza, line 87: **praise:** ‏This bug is a great catch! Thank you

pkg/commands/assemble.go
```diff
@@ -42,7 +42,7 @@ func NewAssembleCommand(
 ) *AssembleCommand {
 	return &AssembleCommand{
 		containermanager: cm,
-		createCmd:        NewCreateCommand(cm, ui.NewDevNullProgress()),
+		createCmd:        NewCreateCommand(cm, ui.NewDevNullProgress(), nil),
```
balanza: **issue:** ‏We must pass a proper prompter instance here, too. Thus, we need to build the prompter object in the CLI layer and pass it down,

pkg/commands/ephemeral.go
```diff
@@ -29,7 +29,7 @@ func NewEphemeralCommand(
 ) *EphemeralCommand {
 	return &EphemeralCommand{
 		containerManager: cm,
-		createCmd:        NewCreateCommand(cm, progress),
+		createCmd:        NewCreateCommand(cm, progress, nil),
```
balanza: **issue:** ‏same as above: we need to provide a real prompter instance

internal/userenv/user_environment.go
```diff
@@ -71,14 +71,14 @@ func LoadUserEnvironment(ctx context.Context) *UserEnvironment {
 	if uid := os.Getuid(); uid >= 0 {
 		env.UserID = strconv.Itoa(uid)
 	} else if uid, err := exec.CommandContext(ctx, "id", "-ru").Output(); err == nil {
-		env.UserID = string(uid)
+		env.UserID = strings.TrimSpace(string(uid))
```
balanza, line 74: **praise:** ‏See no evil 🙈

pkg/containermanager/providers/common.go
```diff
@@ -0,0 +1,23 @@
+package providers
```
balanza: **issue:** ‏

short form: I think it's better to drop this commit.

long form: as you can see,`docker.go` and `podman.go` have the same code with a lot of duplication (`podman.go` is literally a copy-paste with punctual editing). Is it good code? Hell no, it's horrible. But it's deliberate.
We couldn't find a suitable abstraction that accommodates both existing providers and those that may be added in the future. We didn't want to rush into an abstraction so we deferred. 

This commit, although highly reasonable, I'm afraid it gives a false sense of _organisation_ that lowers the urge to solve the duplication problem. Furthermore, `commons` modules are magnets for unscoped code, and that would degrade the code quality. 

It would be like hanging a picture in front of a damp stain. I think this codebase is in a phase where it is better to have the mould on sight rather than let it spread hiddenly.

(sorry I got too philosophical 😬)

balanza (changes requested):
You solved A LOT of nasty bugs, thank you! I am requesting some changes to the proposed fixes, mostly to keep them coherent with the rest of the codebase. But I really appreciate this PR!

pkg/containermanager/providers/podman.go
mirkobrombin: I see your point, and it does make sense. Looking at it again, an empty string is indeed more noticeable than a random ID.

I've also noticed that the list of containers is already filtered by the 'manager=distrobox' label abd since distrobox doesn't allow creating a container without a name, the bug would only exist on the podman side. I am fine with restoring the empty string.

pkg/commands/ephemeral.go
mirkobrombin: Done

pkg/commands/assemble.go
mirkobrombin: Done

pkg/containermanager/providers/common.go
mirkobrombin: Fair point, I come from codebases where extracting shared types into a common file is the default move, so it felt natural but thinking about it in this context, it does make sense, deduplication between docker and podman is deliberate and still needs proper abstraction.

Reverted.

balanza: Thanks!

balanza (approved):
Great job! Thanks.

PR #2051, opened 2026-04-05
fix: remove breaking spaces exported desktop files

simon-wg:
I noticed some issues when exporting vscode to my host.

Due to not having any of the rootful/enter_flags/sudo_prefix options set, the Exec command of the exported desktop file had too many spaces, not following standards, and causing bemenu to be unable to launch them.
The same issue was present for the extra_flags.

The solution was to use shell parameter expansion instead of just listing the parameters.

Examples before the fix:
- Exporting an app.
`Exec=/usr/bin/distrobox-enter  -n devel  --   /usr/share/code/code  %F `
- Generating desktop entry
`Exec=/usr/bin/distrobox enter  devel`

After fix:
- Exporting an app.
`Exec=/usr/bin/distrobox-enter -n devel -- /usr/share/code/code %F`
- Generating desktop entry
`Exec=/usr/bin/distrobox enter devel`

### DIFF
--- a/internal/inside-distrobox/assets/distrobox-export
+++ b/internal/inside-distrobox/assets/distrobox-export
@@ -288,7 +288,7 @@ if [ "${is_sudo}" -ne 0 ]; then
 fi
 
 # Prefix to add to an existing command to work through the container
-container_command_prefix="${DISTROBOX_PATH:-"distrobox"} enter ${rootful} -n ${container_name} ${enter_flags} -- ${sudo_prefix} "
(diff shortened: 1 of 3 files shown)


dottorblaster:
Did you try replicating the behavior on the `next` branch? We gladly accept fixes there

simon-wg:
Oh sorry, I didn't read the part about a complete rewrite in go.
I'm able to replicate the behavior on `next` as well, at least for exporting, so I'll rebase my changes to the script there! 

Edit: Was able to replicate for generate-entry as well, so have ported that fix to go templates.

dottorblaster (approved):
Thank you so much!

PR #2055, opened 2026-04-12
fix(enter): su argument order in unshare_groups path (legacy)

Aromatic05:
## Summary

Fix the `su` argument order used by `distrobox-enter` in the `unshare_groups` path.

The current code builds:

```sh
su USER -m --pty -s /bin/sh -c ...
```

This is fragile because `su` documents its syntax as:

```sh
su [options] [-] [user|UID [argument...]]
```

When `user` is specified, additional arguments may be passed to the target shell. In practice, this makes the argument order significant.

This patch changes the generated order to:

```sh
su -m --pty -s /bin/sh USER -c ...
```

without changing the existing argument-wrapping strategy.

## Why this matters

This became user-visible for me on Arch in the `--init` / `unshare_groups` path when the target shell was `zsh`.

Minimal reproducer:

```sh
su USER -m --pty -s /bin/sh -c 'echo OK'
```

Observed on Arch:

* util-linux 2.41.3-2: works
* util-linux 2.42-1: fails with `zsh: no such option: pty`

There is also an external Debian report showing a related util-linux 2.42 regression in `su` shell / argument handling:

* [Debian Bug#1132610](https://www.mail-archive.com/debian-bugs-dist%40lists.debian.org/msg2093584.html): `su <user> -s /bin/sh -c ...` behavior changed in 2.42

I am not claiming distrobox should depend on permissive parsing here. The fix is simply to emit `su` arguments in a robust order.

## Reproducer

The problem is visible in the current ordering:

```sh
su USER -m --pty -s /bin/sh -c 'echo OK'
```

and is fixed by the equivalent robust ordering:

```sh
su -m --pty -s /bin/sh USER -c 'echo OK'
```

## Fix

Move `container_command_user` so that all `su` options are emitted before the username.

## Notes

This is a minimal fix:

* no behavior change outside the `unshare_groups` path
* no rewrite of the existing `set -- ... "$@"` strategy
* only argument ordering is corrected

### DIFF
--- a/distrobox-enter
+++ b/distrobox-enter
@@ -611,12 +611,12 @@ fi
 if [ "${unshare_groups:-0}" -eq 1 ]; then
 	# shellcheck disable=SC2089,SC2016
 	set -- "-c" '"$0" "$@"' -- "$@"
+	set -- "${container_command_user}" "$@"
 	set -- "-s" "/bin/sh" "$@"
 	if [ "${headless}" -eq 0 ]; then
 		set -- "--pty" "$@"
 	fi
 	set -- "-m" "$@"
-	set -- "${container_command_user}" "$@"
 	set -- "su" "$@"
 fi
 


dottorblaster:
Thanks! Did you also have time to check if this occurs on the `next` branch? We also appreciate fixes there :smile:

Aromatic05:
Thanks! I did take a quick look at the `next` branch as well, and I think there may still be a few related issues there.

In `pkg/containermanager/providers/podman.go`, `buildCommandArgs()` currently seems to build the `unshareGroups` path roughly as:

```go
unshareArgs = append(unshareArgs, user, "-s", "/bin/sh", "-c", `"$0" "$@"`, "--")
```

which results in a command shape like:

```sh
su --pty USER -s /bin/sh -c ...
```

So unless I am missing something, the argument ordering issue still appears to be present there as well: `USER` is still placed before `su` options like `-s` / `-c`, which is the same class of problem that caused the regression on the shell implementation side.

While looking at it, I also noticed two additional points that may be worth clarifying:

1. **Missing `-m` / `--preserve-environment`**
   The shell implementation previously added `-m` in the unshared-groups path (`preserve SHELL with unshared groups`). I am not sure whether its absence in `next` is intentional design, or whether it was just omitted during the Go port.

2. **noTTY handling**
```go
if noTTY {
	unshareArgs = append(unshareArgs, "--pty")
}
```
   The `--pty` logic also looks worth double-checking to ensure it matches the intended `--no-tty` semantics across the new implementation.

So from my side, I think there are three things worth addressing in `next`:

* `su` option ordering
* whether `-m` should be preserved
* `pty` / `noTTY` behavior

If that sounds reasonable, I’d be happy to open a follow-up PR for the `next` branch as well.

dottorblaster:
Thanks @Aromatic05, it looks very much reasonable, can you open the PR against `next` as well? Once we have both we can merge both and ensure the behavior is kept consistent across the two versions.

dottorblaster (approved):

PR #2059, opened 2026-04-19
feat: Implement "Remove" action on exported apps to delete them

iTrooz:
This PR implements a "remove" action on exported apps, to be able to delete them.
image

Executing that action will either run `distrobox export --delete ...` inside the container, *or* rm the desktop file directly if the distrobox does not exist anymore (so the user can clean up old exported apps. This is the use-case that made me write this PR)


dottorblaster:
Thanks! Would you be so kind to make this PR against the `next` branch?

iTrooz:
Yeah ! it's done in https://github.com/89luca89/distrobox/pull/2061

PR #2060, opened 2026-04-19
feat: clean up distrobox exports/entries even if container is already deleted

iTrooz:

PR #2062, opened 2026-04-20
feat: clean up distrobox exports/entries even if container is already deleted

iTrooz:
See https://github.com/89luca89/distrobox/pull/2060


iTrooz:
I also added a failsafe on cleanup() because I accidentally triggered it with an empty container name during tests, and ended up deleting a lot of other files on my computer. Tell me if you'd prefer me to remove it :)

pkg/commands/rm.go
```diff
@@ -61,13 +70,17 @@ func (c *RmCommand) Execute(ctx context.Context, options RmOptions) (*RmResult,
 		return nil, fmt.Errorf("failed while listing contaiers: %w", err)
 	}
 
+	explicitelyRequested := options.ContainerNames
 	distroboxesToRemove := getContainersToRemove(listResult.Containers, options.ContainerNames, options.All)
 
 	userEnv := userenv.LoadUserEnvironment(ctx)
 	userHome := userEnv.Home
 
+	// Remove containers
```
balanza: **issue:** ‏Useless comment
```suggestion
```

pkg/commands/rm.go
```diff
@@ -127,6 +146,10 @@ func (c *RmCommand) removeContainer(
 }
 
 func (c *RmCommand) cleanup(ctx context.Context, userHome, containerName string) {
+	if containerName == "" {
+		panic("Refusing to run cleanup for empty container name")
+	}
+
```
balanza: **issue:** ‏Generally speaking, we should prefer explicit errors over `panic`.

More specifically, in this case, the `cleanup` function is good to remain a function that is expected to never fail. Also, I do not understand if this input validation is preventing a real problem.

balanza (changes requested):

pkg/commands/rm.go
iTrooz: > Generally speaking, we should prefer explicit errors over panic.

It's a case of "this should never happen in the first place", that's why I didn't handle it with an error. This code path should be unreachable under normal conditions. But if it *is* reached, it can be destructive.

It's an assert to validate an invariant.

>  I do not understand if this input validation is preventing a real problem.

Only problems that can happen when developing. For example, I accidentally triggered it with an empty container name during my tests, and ended up deleting a lot of other files on my computer.

iTrooz: Just realized that now that user inputs can be passed to cleanup(), it would be safer to do a proper validation of container names first. So I ended up using podman's regex: https://github.com/containers/podman/blob/dbfddb82cb5750b3c22450cb74a3f47e96b5dfe0/vendor/go.podman.io/common/libnetwork/types/define.go#L26

iTrooz:
Hello ! Any news ? Changes I need to make ?

balanza:
Hi, 

Right now, we are focusing on making Distrobox v2 stable, which means having feature parity with Distrobox v1. That implies that, for the moment, we are not considering new features or bug fixes besides those that are trivial to merge and do not require much brain power from us.

On this PR, we are not convinced by the implementation, but at the same time, we cannot give you the level of feedback you deserve.

Let's postpone this contribution until Distrobox v2 is stable. 

Closing for now.

iTrooz:
Would you mind reopening & drafting this PR instead, so it's still on my feed and I remember to come back to it later ? Thanks

iTrooz:
Hi ! Should I re-submit these PRs against `main`, or not yet ?

PR #2063, opened 2026-04-20
fix(enter): unshare_groups su args and tty handling

Aromatic05:
## Summary

Follow-up for #2055, targeting the `next` branch.

While checking `next`, I found that the `unshare_groups` path in `pkg/containermanager/providers/podman.go` still appears to build the `su` invocation with the username placed before some `su` options, which is the same class of ordering problem fixed in #2055 for the shell implementation.

This PR addresses the `next` branch implementation by:

- fixing `su` option ordering in the `unshare_groups` path
- preserving the intended environment behavior where appropriate
- aligning `pty` / `noTTY` handling with the expected semantics

## Why

In #2055, the shell-side implementation was fixed because the generated `su` command relied on fragile argument ordering. The `next` branch appears to still have a related issue in the Go implementation.

Context from the previous discussion:
- https://github.com/89luca89/distrobox/pull/2055#issuecomment-4276427756

## Notes

This is intended as the `next` counterpart to #2055 so the behavior stays consistent across both versions.


pkg/containermanager/providers/docker.go
```diff
@@ -970,10 +970,10 @@ func buildCommandArgs(customCommand string, user string, noTTY bool, unshareGrou
 	// Handle unshare_groups mode - use su to trigger proper login
 	if unshareGroups {
 		unshareArgs := []string{"su"}
-		if noTTY {
+		if !noTTY {
```
dottorblaster, line 973: Image

dottorblaster (approved):
Thanks! ❤️

fabriziosestito (approved):

balanza (approved):

PR #2064, opened 2026-04-22
ci: add commitlint

fabriziosestito:
We decided to adopt the Conventional Commits specification ([https://www.conventionalcommits.org/en/v1.0.0/](https://www.conventionalcommits.org/en/v1.0.0/)).

This PR introduces commitlint to validate the commits in pull requests.

### DIFF
--- a/.github/commitlint.config.js
+++ b/.github/commitlint.config.js
@@ -1,0 +1,7 @@
+module.exports = {
+  extends: ["@commitlint/config-conventional"],
+  rules: {
+    "body-max-line-length": [0, "always", Infinity], // disables the 100-char limit
+    "header-max-length": [0, "always", Infinity], // disables the 100-char limit
+  },
+};
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -184,3 +184,15 @@ jobs:
           for i in ${SHELL_SCRIPTS_DIR}/*; do
           ./shell-funcheck-amd64 check "$i"
           done
+
+  commitlint:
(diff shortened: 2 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -184,3 +184,15 @@ jobs:
...
+
+  commitlint:
+    runs-on: ubuntu-latest
+    if: github.event_name == 'pull_request'
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          fetch-depth: 0
+      - name: Install commitlint
+        run: npm install -D @commitlint/cli @commitlint/config-conventional
```
balanza: **question:** ‏Should we pin the npm package versions?

.github/commitlint.config.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
```
balanza, line 1: **question:** ‏given it's a resource used by the CI, maybe it makes sense to store this config near the workflows (no strong opinion, though)

balanza (approved):
Thanks! A couple of questions, but LGTM

.github/commitlint.config.js
fabriziosestito: Yeah, it’s a bit annoying that it’s not a hidden file, honestly. I don’t think it’ll ever be used in the local setup, so it’s probably best to move it.

balanza (approved):

PR #2068, opened 2026-04-26
fix: passing root correctly to create and generate-entry command

balanza:
The `root` flag is handled in the root cli command. if we re-declare the
flag here we shadow it, resulting in a contaner manager initializated as
non-root.

Before this commit, these two commands produce respectively one root and
one non-root execution
```
distrobox --root create # root
distrobox create --root # non-root

distrobox --root generate-entry # root
distrobox generate-entry --root # non-root
```


dottorblaster (approved):
LGTM

PR #2070, opened 2026-04-27
fix(enter): propagate errors when entering containers

Aromatic05:
## Summary

Fix multiple cases where errors were silently ignored in Docker and Podman providers.

This mainly affects `distrobox enter`, which could report success even when:
- container start failed
- container exec failed
- setup/log monitoring encountered errors

## Changes

- propagate errors from `startContainer` before entering
- propagate errors from the final exec command
- improve error reporting in `run`
- avoid swallowing errors in `waitForSetup`
- improve error context in inspect-related paths
- add tests covering start and exec error propagation

## Rationale

The current implementation ignores several command errors, which can hide real failures and lead to confusing behavior (e.g. `enter` appearing to succeed while the command failed).

This aligns with the contributing guideline to always check the outcome of commands and functions that can fail.

## Testing

- added tests to ensure:
  - start failures are propagated
  - exec failures are propagated
- verified that failing `enter` commands now return non-zero exit status

### DIFF
--- a/pkg/containermanager/providers/docker.go
+++ b/pkg/containermanager/providers/docker.go
@@ -673,7 +673,7 @@ func (d *Docker) InspectContainer(ctx context.Context, containerName string) (*c
 
 	var inspects []inspectOutput
 	if err := json.Unmarshal([]byte(output), &inspects); err != nil {
-		return nil, errors.New("error marshaling json into containerInspect")
+		return nil, fmt.Errorf("error unmarshaling json into containerInspect: %w", err)
 	}
 
 	if len(inspects) == 0 {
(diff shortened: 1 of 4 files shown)


pkg/containermanager/providers/docker.go
```diff
@@ -534,7 +534,9 @@ func (d *Docker) Enter(
 	if err != nil || inspectResult.ContainerStatus != RunningStatus {
 		logTimestamp := timestampNow()
 
-		_ = d.startContainer(ctx, options.ContainerName, progress)
+		if err := d.startContainer(ctx, options.ContainerName, progress); err != nil {
+			return err
+		}
```
dottorblaster, line 540: Whoops, not my proudest moment, I think this is my fault 😬

pkg/containermanager/providers/docker_internal_test.go
```diff
@@ -483,3 +489,98 @@ func hasSubstring(s, substr string) bool {
...
+    if [ -n "$FAKE_EXEC_STDERR" ]; then
+      printf "%s" "$FAKE_EXEC_STDERR" >&2
+    fi
+    exit "${FAKE_EXEC_EXIT:-0}"
+    ;;
+  *)
+    exit 0
+    ;;
+esac
+`
```
dottorblaster, line 573: Love it! ❤️

dottorblaster (approved):
Thank you for both fixing this bug and showcasing the new opportunities we got with the new tooling!

PR #2081, opened 2026-04-29
Fix/command args

89luca89:
This pr does two things:

move common docker/podman functions from docker.go into containermanager.go
fix the custom command to be a []string, so it doesn't mangle and lose eventual shell argument significance
Specifically this fixes cases where we pass a script into enter like:

`distrobox enter foo -- sh -c "tmux a || tmux"`


balanza (approved):
Awesome job!

dottorblaster (approved):
LET'S GO

89luca89:
> LET'S GO

LETS ~SHELL~ GO

PR #2082, opened 2026-04-29
fix: print container name on create when not provided

balanza:
The `create` command, when the container name is not provided, it calculates it based on some defaults. This PR prints the actual container name.

```sh
# before
Distrobox 'ubuntu-noble' successfully created.
To enter, run:

distrobox enter

# after
Distrobox 'ubuntu-noble' successfully created.
To enter, run:

distrobox enter ubuntu-noble
```


89luca89 (approved):

dottorblaster (approved):
❤️

PR #2088, opened 2026-05-04
fix(cli): read correct version

balanza:
Previously, the CLI hardcoded "1.0.0" as the version string in two places (the root command and the generate-entry subcommand).

Introduce pkg/version, a minimal package that exposes a single variable Version, defaulting to "dev". The root command now references such a value; generate-entry does not need it al all

The Makefile computes the version at build time by running:

git describe --tags --always
This produces the exact tag name when HEAD is tagged, or --g when it is not, giving a precise description of the build relative to the nearest ancestor tag. If git is unavailable, the value falls back to "dev".

The resulting string is injected into the binary via:

-ldflags "-X github.com/89luca89/distrobox/pkg/version.Version=<value>" so that distrobox --version and distrobox generate-entry --version both report the exact version of the build.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,11 +27,12 @@ jobs:
 
       - name: Build
         env:
-          CGO_ENABLED: "0"
           GOOS: ${{ matrix.goos }}
           GOARCH: ${{ matrix.goarch }}
         run: |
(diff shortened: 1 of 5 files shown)


dottorblaster (approved):
LGTM

PR #2089, opened 2026-05-04
fix(enter): always attach stdio to final exec

89luca89:
`Interactive: !NoTTY` routed stdio to discarded buffers whenever --no-tty was set, breaking ptyxis. --no-tty only governs whether `--tty` is appended to the exec args, not stdio plumbing.

### DIFF
--- a/pkg/containermanager/providers/docker.go
+++ b/pkg/containermanager/providers/docker.go
@@ -546,7 +546,7 @@ func (d *Docker) Enter(
 		progress.Finalize("Container Setup Complete!")
 	}
 
-	if _, err := d.run(ctx, append(command, commandArgs...), runOptions{Interactive: !options.NoTTY}); err != nil {
+	if _, err := d.run(ctx, append(command, commandArgs...), runOptions{Interactive: true}); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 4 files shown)


dottorblaster (approved):
LGTM

pkg/containermanager/providers/docker.go
```diff
@@ -546,7 +546,7 @@ func (d *Docker) Enter(
 		progress.Finalize("Container Setup Complete!")
 	}
 
-	if _, err := d.run(ctx, append(command, commandArgs...), runOptions{Interactive: !options.NoTTY}); err != nil {
+	if _, err := d.run(ctx, append(command, commandArgs...), runOptions{Interactive: true}); err != nil {
```
balanza, line 549: **issue:** ‏Doing so leave a dangling `NoTTY` options in both cli command and command packages

balanza (changes requested):

balanza (approved):

pkg/containermanager/providers/docker.go
balanza: Wrong assumption

PR #2090, opened 2026-05-05
fix: clean up partial containers when interrupted

89luca89:
Wire signal.NotifyContext in main so SIGINT/SIGTERM cancel the context instead of killing the process.

ephemeral and assemble now defer rm on a detached cleanup ctx, so the in-flight container is removed even when the user hits Ctrl+C.

### DIFF
--- a/cmd/distrobox/main.go
+++ b/cmd/distrobox/main.go
@@ -4,19 +4,32 @@ import (
 	"context"
 	"log"
 	"os"
+	"os/signal"
+	"syscall"
 
 	"github.com/89luca89/distrobox/internal/cli"
 	"github.com/89luca89/distrobox/pkg/config"
 )
 
 func main() {
+	if err := run(); err != nil {
+		log.Fatal(err)
+	}
+}
+
+func run() error {
 	cfg, err := config.LoadValues()
 	if err != nil {
-		log.Fatal(err)
+		//nolint:wrapcheck // main reports errors as-is
+		return err
 	}
 
+	// SIGINT register
+	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
+	defer stop()
+
 	cmd := cli.NewRootCommand(cfg)
-	if err := cmd.Run(context.Background(), os.Args); err != nil {
-		log.Fatal(err)
-	}
+
+	//nolint:wrapcheck // main reports errors as-is
+	return cmd.Run(ctx, os.Args)
 }
--- a/pkg/commands/ephemeral.go
+++ b/pkg/commands/ephemeral.go
@@ -4,11 +4,14 @@ import (
 	"context"
 	"fmt"
 	"math/rand/v2"
+	"time"
 
 	"github.com/89luca89/distrobox/pkg/config"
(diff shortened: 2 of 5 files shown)


cmd/distrobox/main.go
```diff
@@ -15,8 +17,13 @@ func main() {
...
+	// First SIGINT/SIGTERM cancels the context so deferred cleanup can run.
+	// A second one falls through to the default handler and forces exit.
+	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
+	defer stop()
+
 	cmd := cli.NewRootCommand(cfg)
-	if err := cmd.Run(context.Background(), os.Args); err != nil {
-		log.Fatal(err)
+	if err := cmd.Run(ctx, os.Args); err != nil {
+		log.Print(err)
```
fabriziosestito: We could wrap `run` in a separate function so we can keep using `log.Fatal`:

```go
package main

import (
	"context"
	"log"
	"os"
	"os/signal"
	"syscall"

	"github.com/89luca89/distrobox/internal/cli"
	"github.com/89luca89/distrobox/pkg/config"
)

func main() {
	if err := run(); err != nil {
		log.Fatal(err)
	}
}

func run() error {
	cfg, err := config.LoadValues()
	if err != nil {
		return err
	}

	// First SIGINT/SIGTERM cancels the context so deferred cleanup can run.
	// A second one falls through to the default handler and forces exit.
	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
	defer stop()

	cmd := cli.NewRootCommand(cfg)
	return cmd.Run(ctx, os.Args)
}
```

fabriziosestito (changes requested):

cmd/distrobox/main.go
89luca89: Thanks, done :+1:

pkg/commands/ephemeral.go
```diff
@@ -53,29 +58,31 @@ func (c *EphemeralCommand) Execute(ctx context.Context, opts EphemeralOptions) e
...
 	createOpts.DryRun = opts.DryRun
 	createOpts.NonInteractive = true
-	if _, err := c.createCmd.Execute(ctx, createOpts); err != nil {
-		return fmt.Errorf("failed to create ephemeral container: %w", err)
+	if _, createErr := c.createCmd.Execute(ctx, createOpts); createErr != nil {
+		return fmt.Errorf("ephemeral: %w", createErr)
 	}
 
+	defer func() {
+		cleanupCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), ephemeralCleanupTimeout)
```
balanza, line 66: **question:** ‏If I get it right: `enterCmd` cannot take more than 30 seconds, otherwise the container creation is cancelled

89luca89: It's only for the cleanup context, so rm only

fabriziosestito (approved):

fabriziosestito (approved):

dottorblaster (approved):
aw yeah

PR #2092, opened 2026-05-06
fix(assemble): preserve section names from manifest

89luca89:
`makeContainerName` replaced user-provided names with DefaultContainerName whenever `image=` was empty, and `sectionToItem` kept whitespace from headers like `[ generic1]`.
Both made `assemble create` emit `--name my-distrobox` (or a leading-space name) for sections without an explicit image.


dottorblaster (approved):
Love these tests ❤️

PR #2104, opened 2026-05-19
fix(ephemeral): pass custom command to enter

balanza:
## Summary

- Resolves the `// TODO: handle enter command` in `pkg/commands/ephemeral.go` so the trailing positional arguments are actually forwarded to `distrobox enter`.
- Adds a `CustomCommand []string` field to `EphemeralOptions` (alongside the embedded `CreateOptions`), wires it into the synthesized `EnterOptions`, and populates it in the CLI from `cmd.Args().Slice()` (urfave/cli v3 consumes the `--` separator and exposes the remainder as positional args).
- Matches the bash reference (`distrobox-ephemeral` → `distrobox-enter ${extra_flags} "${name}" ${container_command}`).

Without this, `distrobox ephemeral --image alpine -- cat /etc/os-release` silently dropped the trailing command and just dropped the user into the default login shell.

## Test plan

- [x] `go test ./pkg/commands/...` — added two new tests:
  - `TestEphemeralCommand_PassesCustomCommandToEnter` asserts the spy records the exact `CustomCommand` slice on `Enter`.
  - `TestEphemeralCommand_EmptyCustomCommandIsNotForwardedAsArgs` asserts the absence of `CustomCommand` does not synthesize one (regression guard).
- [x] `make` (build)
- [x] `make fmt`
- [x] `make lint` (0 issues)
- [x] `make test` (full suite, including providers)

### DIFF
--- a/internal/cli/ephemeral.go
+++ b/internal/cli/ephemeral.go
@@ -77,7 +77,8 @@ func ephemeralAction(ctx context.Context, cmd *cli.Command, cfg *config.Values) 
 			GenerateEntry:           false,
 			Rootful:                 cmd.Bool("root"),
 		},
-		DryRun: cmd.Bool("dry-run"),
+		CustomCommand: cmd.Args().Slice(),
+		DryRun:        cmd.Bool("dry-run"),
 	}
 
 	progress := ui.NewProgress(os.Stderr)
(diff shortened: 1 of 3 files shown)


dottorblaster (approved):
LGTM

PR #2106, opened 2026-05-19
feat(generate-entry): auto-detect icon from container distro

balanza:
## Summary

Resolves a `TODO` in the Go port of `distrobox-generate-entry`:
`getDesktopIcon` used to ignore `--icon auto` (and the implicit `auto`
applied by `--all`) and always return the generic terminal logo.

- Adds `pkg/commands/distro_icons.go` with the same `DISTRO_ICON_MAP`
  shipped in the Bash `distrobox-generate-entry` (alma, alpine, alt,
  arch, centos, clear, debian, deepin, fedora, gentoo, kali, kdeneon,
  opensuse-leap, opensuse-tumbleweed, rhel, rocky, ubuntu, vanilla,
  void) plus a few obvious aliases (archlinux, alpinelinux, kalilinux,
  tumbleweed, redhat) so common image names match.
- Extends `getDesktopIcon` to take a `distroHint` string. Hints are
  matched case-insensitively against the map; more specific keys
  (e.g. `opensuse-tumbleweed`, `archlinux`) are listed before short
  prefixes (`arch`, `opensuse`) so they win on substring match.
- Plumbs the hint down from `Execute` -> `createEntry` ->
  `composeDesktopEntryData`. The `--all` flow now records each
  container's image alongside its name, so detection works without
  extra calls. When a hint is missing (e.g. single-container call) we
  fall back to the container name, then to the generic terminal icon.
- Extracts the target-resolution `switch` from `Execute` into a
  `resolveTargets` helper to keep cognitive complexity below the
  project's `gocognit` budget.

The Bash original reads `/etc/os-release` from the running container;
matching against the image name covers the common cases (`fedora:39`,
`registry.fedoraproject.org/fedora-toolbox:latest`, `archlinux`,
`docker.io/library/ubuntu:22.04`, ...) without requiring the
container to be running, which is needed for `distrobox generate-entry
--all` to work on stopped boxes too.

## Test plan

- [x] `go test ./pkg/commands/ -run TestGetDesktopIcon` - 8 new
  white-box tests in `generate_entry_internal_test.go` covering:
  fedora detection, ubuntu detection, unknown-distro fallback to the
  default icon, custom-URL pass-through, opensuse-tumbleweed beating
  generic opensuse, archlinux detected from a registry path,
  case-insensitive matching, and the empty-hint fallback.
- [x] `make` - build succeeds.
- [x] `make fmt` - clean.
- [x] `make lint` - 0 issues.
- [x] `make test` - full suite green, no regressions.

### DIFF
--- a/pkg/commands/distro_icons.go
+++ b/pkg/commands/distro_icons.go
@@ -1,0 +1,67 @@
+package commands
+
+import "strings"
+
+// distroIconBaseURL is the upstream raw asset root for the bundled distro
+// logos shipped with the distrobox project.
+const distroIconBaseURL = "https://raw.githubusercontent.com/89luca89/distrobox/main/docs/assets/png/distros/"
+
+// distroIconEntry maps a substring expected to appear in a container image
+// or container name to the matching desktop icon URL.
+//
+// The list is ordered from most specific to most generic so that, when
+// scanning a hint string for the first matching key, more specific variants
+// (e.g. opensuse-tumbleweed, kdeneon) win over short, ambiguous prefixes.
+type distroIconEntry struct {
+	key string
+	url string
+}
+
+// distroIconMap mirrors the DISTRO_ICON_MAP from the upstream
+// distrobox-generate-entry shell script. Keep entries in sync when new
+// distro icons are added under docs/assets/png/distros.
+//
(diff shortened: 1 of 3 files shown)


dottorblaster (approved):
Fair enough, thanks

PR #2108, opened 2026-05-19
fix(ephemeral): avoid name collisions when generating random container name

balanza:
## Summary

Resolves the `FIXME: avoid collisions` in `pkg/commands/ephemeral.go` for the auto-generated ephemeral container name path. When the user does not supply a name, ephemeral generates `distrobox-XXXXXXXXXX`; the previous code never checked whether that name was already used and could blindly attempt to create on top of an existing container.

The fix wraps generation in a small retry loop that consults `ContainerManager.Exists` and returns an explicit error if 10 attempts in a row all collide. Mirrors the bash original's `mktemp -u distrobox-XXXXXXXXXX` intent of avoiding collisions. The existence check is skipped under `--dry-run`, matching the rest of the dry-run pipeline (which never queries the container manager either).

## Changes

- `pkg/commands/ephemeral.go`: new `EphemeralCommand.makeUniqueRandomName(ctx, dryRun)` used by `Execute`; `makeRandomName` is left as a pure name generator; magic `10` for the suffix length lifted into a named constant.
- `pkg/internal/testutil/mock_containermanager.go`: optional `ExistsFn` hook on `MockContainerManager` so tests can simulate collisions. Default behavior unchanged (still returns `false`). `CloneAsRoot()` now propagates `ExistsFn` into the root clone so hooks are consistent across rootless/root variants.
- `pkg/commands/ephemeral_internal_test.go`: new internal tests covering unique-on-first-attempt, retry-after-collisions, all-collide-returns-error, and dry-run-skips-check paths.

## Test plan

- [x] `make` (build)
- [x] `make fmt`
- [x] `make lint` (0 issues)
- [x] `make test` — 4 new tests pass:
  - `TestEphemeralCommand_MakeUniqueRandomName_FirstAttemptUnique`
  - `TestEphemeralCommand_MakeUniqueRandomName_RetriesAfterCollisions`
  - `TestEphemeralCommand_MakeUniqueRandomName_AllCollideReturnsError`
  - `TestEphemeralCommand_MakeUniqueRandomName_DryRunSkipsExistenceCheck`

### DIFF
--- a/pkg/commands/ephemeral.go
+++ b/pkg/commands/ephemeral.go
@@ -11,7 +11,11 @@ import (
 	"github.com/89luca89/distrobox/pkg/ui"
 )
 
-const ephemeralCleanupTimeout = 30 * time.Second
+const (
+	ephemeralCleanupTimeout       = 30 * time.Second
+	ephemeralMaxNameGenAttempts   = 10
+	ephemeralRandomNameSuffixSize = 10
+)
 
 type EphemeralOptions struct {
 	CreateOptions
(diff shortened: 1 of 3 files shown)


dottorblaster (approved):
Thanks!

PR #2109, opened 2026-05-19
fix(rm): propagate verbose flag to generate-entry cleanup

balanza:
## Summary
Resolves the `TODO: handle verbose` in `pkg/commands/rm.go`'s cleanup path. The `--verbose` flag was being ignored at the cleanup-time call to `generate-entry --delete`; it is now propagated from `RmOptions` through `removeContainer` and `cleanup` into the `GenerateEntryOptions` used to delete the desktop entry.

`internal/cli/rm.go` now reads the global `--verbose` flag (matching the existing pattern in the `generate-entry` and `enter` subcommands).

`MockContainerManager` gains optional `ListContainersResult` / `InspectContainerResult` overrides so tests can exercise paths that depend on real container metadata; defaults are unchanged. `CloneAsRoot` propagates these overrides into the root clone so root-aware code paths see consistent behavior.

The ephemeral and assemble cleanup defers keep the zero-value `Verbose` — those are internal teardown paths and intentionally out of scope.

## Test plan
- [x] `make build`
- [x] `make fmt`
- [x] `make lint`
- [x] `make test`
- [x] `TestRmCommand_Execute_CleanupRemovesExportedDesktopApp` covers the per-app exported desktop file branch.
- [x] `TestRmCommand_Execute_CleanupRemovesGenerateEntryDesktop` covers the GenerateEntry delete branch — the call site the verbose plumbing flows through.

### DIFF
--- a/internal/cli/rm.go
+++ b/internal/cli/rm.go
@@ -63,6 +63,7 @@ func rmAction(ctx context.Context, cmd *cli.Command, cfg *config.Values) error {
 		Force:          cmd.Bool("force"),
 		All:            cmd.Bool("all"),
 		RemoveHome:     cmd.Bool("rm-home"),
+		Verbose:        cmd.Bool("verbose"),
 		ContainerNames: names,
 	}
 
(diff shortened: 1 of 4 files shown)


dottorblaster (approved):
Whoopsie, totally forgot about that, thanks (and thanks for the tests)

dottorblaster:
@balanza we have some conflicts here :smile: can you rebase so we can merge right away?

PR #2111, opened 2026-05-19
feat(create): implement --compatibility to fetch supported image list

balanza:
## Summary

- Implements the `distrobox create --compatibility` flag in Go (resolves the `TODO` at `internal/cli/create.go:240`), porting the bash `show_compatibility` helper from the legacy `distrobox-create` script.
- Fetches `docs/compatibility.md` from upstream for the current build version (falling back to `main` for `dev` builds, or to the nearest tag when `git describe --tags` produces a `<tag>-N-g<sha>[-dirty]` suffix), parses the "Containers Distros" table, and caches the deduplicated/sorted image list under `$XDG_CACHE_HOME/distrobox/distrobox-compatibility-<ref>` so subsequent invocations are offline-friendly.
- Cache writes are atomic (temp file + rename) so a crashed or concurrent writer cannot leave a partially populated cache file visible to readers.
- The build-time version string is sanitized before being embedded in the cache filename, preventing any ref containing `/`, `..`, or unusual characters from creating nested paths or escaping the cache directory.
- The action passes its `context.Context` into the HTTP fetch so Ctrl+C / parent cancellation aborts the request.

## Implementation notes

- New file `internal/cli/compatibility.go` houses the fetch / parse / cache logic.
- Parsing mirrors the original bash pipeline (`sed -n '/| Alma/,/| Void/ p' | cut -f4 | sed s/<br>/\n/ | tr -d ' ' | sort -u`) line-by-line in Go.
- 33 unit tests (one function per case, no table-driven subtests) cover the parser, the cache read/write/atomic-replace path, ref normalization (including `git describe` shapes and bare hashes), ref sanitization for filename safety, the HTTP fetch helpers, and the `XDG_CACHE_HOME` / `HOME` fallback chain.

## Test plan

- [x] `make build` succeeds
- [x] `make test` passes (33 new tests in `internal/cli`, full suite green)
- [x] `make lint` passes with zero issues
- [x] Smoke test: `distrobox create --compatibility` returns 122 image entries on first invocation and serves from cache on the second
- [x] Smoke test with `VERSION=dev`: falls back to `main` ref and writes `distrobox-compatibility-main`

### DIFF
--- a/internal/cli/compatibility.go
+++ b/internal/cli/compatibility.go
@@ -1,0 +1,389 @@
+package cli
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"path/filepath"
+	"regexp"
+	"sort"
+	"strings"
+	"time"
+	"unicode"
+
+	"github.com/89luca89/distrobox/pkg/version"
+)
+
+const (
+	// compatibilityURLTemplate is the upstream URL for docs/compatibility.md.
+	// %s is the git ref (tag or branch) to fetch from.
+	compatibilityURLTemplate = "https://raw.githubusercontent.com/89luca89/distrobox/%s/docs/compatibility.md"
+
+	// compatibilityFetchTimeout caps the time we wait for the HTTP request.
+	compatibilityFetchTimeout = 15 * time.Second
+
+	// devFallbackRef is used in place of the "dev" version (set on local
+	// builds) to fetch the latest upstream compatibility list.
+	devFallbackRef = "main"
+
+	// containersDistrosHeader is the markdown heading that precedes the
+	// "Containers Distros" table in docs/compatibility.md. We parse the
+	// first markdown table that follows it; everything else (including
+	// the "Host Distros" table above it) is ignored.
+	containersDistrosHeader = "## Containers Distros"
+)
+
+// showCompatibility prints the list of container images known to work with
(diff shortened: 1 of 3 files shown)


internal/cli/compatibility.go
```diff
@@ -0,0 +1,350 @@
...
+		}
+	}
+
+	images := make([]string, 0, len(seen))
+	for image := range seen {
+		images = append(images, image)
+	}
+	sort.Strings(images)
+	return images
+}
```
dottorblaster: The table boundaries `"| Alma"` and `"| Void"` are extremely fragile structural markers.

Any of the following silently breaks with an error about parsing compatibility list:
- AlmaLinux renamed/removed
- a new distro added alphabetically before "Alma"
- Void renamed
- a new distro added alphabetically after "Void"

Two suggestions:

- Anchor on the `## Containers Distros` heading and the next  `## heading` instead of arbitrary cell content
- When the markers can't be found, return an explicit error like compatibility table layout changed in upstream `docs/compatibility.md`.

I would especially pick the first one 😬

dottorblaster (changes requested):
Just one thing but then we can merge 👍

internal/cli/compatibility.go
balanza: Done it. The previous implementation was mirroring what is done on `main`, but this is an improvement indeed

dottorblaster (approved):
❤️

PR #2114, opened 2026-05-25
fix(`distrobox-init`): Replace remaining non-POSIX `echo -e` with `printf`

fiftydinar:


dottorblaster:
Thanks! Could you also apply the same patch against the `next` branch?

fiftydinar:
> Thanks! Could you also apply the same patch against the `next` branch?

In new PR? or just change from `main` to `next`?

dottorblaster:
@fiftydinar in a new PR if possible :-)

fiftydinar:
> @fiftydinar in a new PR if possible :-)

Done

https://github.com/89luca89/distrobox/pull/2115

dottorblaster (approved):

PR #2131, opened 2026-06-20
chore(experience): align Go rewrite with reference shell distrobox across create/enter/rm/upgrade/assemble

89luca89:
This branch closes a long tail of parity gaps between the Go rewrite and the reference shell distrobox. It is almost entirely behavior fixes — the commands now look at the same env vars, accept the same arguments, prompt the same way, and produce the same side effects as the shell scripts
they replace. No new top-level features; the one feat(cli) commit restores a refusal (sudo/doas guard) that the shell has and the Go port had lost.

What changed, grouped by area

create
- Honor the positional container_name argument, not just --name.
- Default the name to my-distrobox when the default image is used (shell parity).
- Drop the --image default Value so the containerImage == "" branch survives and the name == default fallback stops misfiring.
- Strip trailing slashes from --home and word-split --volume values.
- Show the host's hostname (via os.Hostname()) in --hostname usage text.
- Mount /dev/null:/run/.distrobox.rootless:ro so distrobox-init doesn't fall back to the /etc/shadow heuristic that false-positives on Docker Desktop / macOS.
- Drop rslave bind propagation on macOS; centralize in bindPropagation() / roBindPropagation() (Linux output unchanged).
- Honor DBX_USERNS_NOLIMIT to drop the keep-id size cap.
- Request icon auto-detection so generated entries aren't icon-less.

enter
- Match the shell: prompt to create the container if it doesn't exist.
- Support the noworkdir flag.
- Build the --dry-run command without requiring a live container.
- Set PWD=<workdir> inside the container and stop forwarding the host's PWD.
- Reorder PATH so /usr/local/{s,}bin precede /usr/{s,}bin.

rm
- Confirm deletions once, default prompt to yes (shell behavior).
- Use Container.IsRunning() for the force prompt — strings.Contains(Status, "Up") missed podman's lowercase running.
- Gate the custom-home prompt on --rm-home; without the flag the home is never touched.
- Wire DBX_CONTAINER_RM_CUSTOM_HOME into --rm-home.
- Print Cannot find container for unknown explicit names.

upgrade
- Match shell prompting.
- Print a per-container Upgrading <name>... banner.

stop
- Honor DBX_CONTAINER_NAME when no positional is given.

assemble
- Treat an existing box as a skip, not an error.
- Skip the rm step entirely in --dry-run (previously dryRun was misrouted into RmOptions.NoTTY and containers were actually deleted).
- Accept the singular volume manifest key alongside volumes, and whitespace-split values.

generate-entry
- Refuse a container that doesn't exist.
- Detect the distro from the image, not from the box name.
- Reuse local icons if present before attempting downloads.

CLI-wide
- SUDO/DOAS guard restored to match the shell's refusal of sudo distrobox.
- DBX_* env vars wired per command; container_manager defaults to autodetect; invalid prompt input is rejected; uid 0 is treated as rootful.


dottorblaster (approved):
LGTM

internal/cli/create.go
```diff
@@ -40,6 +40,7 @@ Examples:
 			&cli.StringFlag{
 				Name:    "image",
 				Aliases: []string{"i"},
+				Sources: cli.EnvVars("DBX_CONTAINER_IMAGE"),
```
balanza: **issue:**  Default values are resolved by the `config.LoadValues` method in https://github.com/89luca89/distrobox/blob/next/cmd/distrobox/main.go#L21. The method already considers the precedence of config files and env vars. See this commit for the usage https://github.com/89luca89/distrobox/commit/ec99b5d88432ff3c439e85400725ba5bf0843f70

internal/cli/create.go
```diff
@@ -182,10 +188,18 @@ func createAction(ctx context.Context, cmd *cli.Command, cfg *config.Values) err
 		return errors.New("container manager not found in context")
 	}
 
+	// DBX_CONTAINER_GENERATE_ENTRY=0 disables entry generation (shell parity);
+	// DBX_CONTAINER_HOME_PREFIX seeds a per-box custom home when no --home is set.
+	generateEntry := !cmd.Bool("no-entry")
+	if v := os.Getenv("DBX_CONTAINER_GENERATE_ENTRY"); v == "0" || v == "false" {
+		generateEntry = false
+	}
```
balanza: **issue:** ‏As above, we should use values from the config module

pkg/ui/prompt.go
```diff
@@ -23,24 +23,31 @@ func NewPrompter(reader bufio.Reader, writer io.Writer) *Prompter {
...
-
-	yes := []string{"y", "yes", "Y", "YES"}
-	no := []string{"n", "no", "N", "NO"}
-
-	if slices.Contains(yes, s) {
-		return true
+	yes := []string{"y", "yes"}
+	no := []string{"n", "no"}
+
+	for {
```
balanza, line 29: **question:** ‏Why the infinite loop?

pkg/commands/create.go
```diff
@@ -216,7 +216,9 @@ func (c *CreateCommand) makeContainerImage(opts *CreateOptions) string {
 //	ghcr.io/void-linux/void-linux:latest-full-x86_64 -> void-linux-latest-full-x86_64
 func (c *CreateCommand) makeContainerName(opts *CreateOptions, containerImage string) string {
 	containerName := opts.ContainerName
-	if containerName == "" && opts.ContainerImage == "" {
+	// Mirror the shell: when no name is given and the effective image is the
+	// default image, use the default name (distinguishable from toolbx).
+	if containerName == "" && opts.ContainerClone == "" && containerImage == c.cfg.DefaultContainerImage {
```
balanza, line 224: **question:** ‏What was the original rationale for this?

(ok for now as we are rewriting the same behaviour)

docs/README.md
```diff
@@ -330,6 +330,21 @@ skip_workdir="0"
...
+>
+> - Values are taken **literally**: `container_user_custom_home="$HOME/..."` and
+>   `PATH="$PATH:..."` are *not* expanded — use absolute paths. The `$HOME`/`$PATH`
+>   entries in the example above illustrate the old sourcing behavior and will not
+>   be expanded here.
+> - Arbitrary shell logic in `.distroboxrc` is **not executed**; only recognized
+>   `key=value` settings take effect.
+> - The reference key `distrobox_sudo_program` is accepted (mapped onto
+>   `sudo_program`).
+
```
balanza, line 347: **praise:** ‏Good to note!

pkg/containermanager/providers/podman.go
```diff
@@ -379,6 +379,13 @@ func (p *Podman) makeCreateCommand(
 		options = append(options, "--volume", "/dev/null:/run/.nopasswd:ro")
 	}
 
+	// Signal rootless mode explicitly so distrobox-init does not rely solely on
+	// the /etc/shadow heuristic, which gives false positives on Docker Desktop /
+	// macOS where the container always has root over the VM filesystem.
+	if !p.root {
+		options = append(options, "--volume", "/dev/null:/run/.distrobox.rootless:ro")
```
balanza, line 386: 🐧

internal/cli/enter.go
```diff
@@ -117,11 +118,26 @@ func enterAction(ctx context.Context, cmd *cli.Command, cfg *config.Values) erro
...
+	// dry-run, which builds the command from host state without the container.
+	if !dryRun && !containerManager.Exists(ctx, containerName) {
+		created, err := offerCreateMissing(ctx, cfg, containerManager, containerName, cmd.Bool("yes"))
+		if err != nil {
+			return err
+		}
+		if !created {
+			return nil // user declined; guidance already printed
+		}
+	}
```
balanza: **issue:** ‏I think this behaviour should go in the command layer (`pkg/commands/enter.go`), otherwise it will only work for distrobox binaries and not for apps that import distrobox as a module - unless that is the intention.

internal/cli/enter.go
```diff
@@ -138,3 +154,43 @@ func enterAction(ctx context.Context, cmd *cli.Command, cfg *config.Values) erro
...
+			errPrinter.Println("Ok. For creating it, run this command:")
+			errPrinter.Println("\tdistrobox create <name-of-container> --image <remote>/<docker>:<tag>")
+			return false, nil
+		}
+	}
+
+	errPrinter.Println("Creating the container %s", containerName)
+
+	progress := ui.NewProgress(os.Stderr)
+	prompter := ui.NewPrompter(*bufio.NewReader(os.Stdin), os.Stdout)
```
balanza: **issue:** ‏As per the comment above, we do not create new UI objects on-the-fly; we rather pass them to the command constructor.

internal/cli/enter.go
```diff
@@ -63,6 +63,7 @@ func newEnterCommand(cfg *config.Values) *cli.Command {
 			&cli.BoolFlag{
 				Name:    "no-workdir",
 				Aliases: []string{"nw"},
+				Sources: cli.EnvVars("DBX_SKIP_WORKDIR"),
```
balanza: **issue:** ‏See comment above (TL;DR use the config pacakge)

internal/cli/upgrade.go
```diff
@@ -40,7 +39,7 @@ Examples:
 			&cli.BoolFlag{
 				Name:    "yes",
 				Aliases: []string{"Y"},
-				Usage:   "non-interactive, upgrade without asking",
+				Usage:   "accepted for compatibility; upgrades never prompt (matches the shell)",
```
balanza: **issue:** ‏Actually, we do not need the flag at all. We can remove it.

internal/cli/upgrade.go
```diff
@@ -59,23 +58,16 @@ func upgradeAction(ctx context.Context, cmd *cli.Command, cfg *config.Values) er
...
-		NonInteractive: cmd.Bool("yes"),
 	}
 
 	printer := ui.NewPrinter(os.Stdout, true)
 	errPrinter := ui.NewPrinter(os.Stderr, true)
 	progress := ui.NewProgress(os.Stderr)
-	prompter := ui.NewPrompter(*bufio.NewReader(os.Stdin), os.Stdout)
 
-	upgradeCmd := commands.NewUpgradeCommand(cfg, containerManager, progress, printer, prompter)
+	upgradeCmd := commands.NewUpgradeCommand(cfg, containerManager, progress, printer)
```
balanza, line 67: **praise:** ‏Good point

pkg/commands/generate_entry.go
```diff
@@ -126,7 +126,20 @@ func (c *GenerateEntryCommand) resolveTargets(
...
+		// can't be reached — we fall back to the name in Execute.
+		images := map[string]string{}
+		if listResult, err := c.listCommand.Execute(ctx); err == nil {
+			for _, container := range listResult.Containers {
+				if container.Name == opts.ContainerName {
+					images[opts.ContainerName] = container.Image
+					break
+				}
+			}
+		}
```
balanza, line 160: **thought:** ‏Please note that the original shell implementation uses the contaner's `/etc/os-release` 

```sh
	# Try to detect container's distribution by using /etc/os-release
	${container_manager_cp_command} "${container_name}":/etc/os-release /tmp/"${container_name}".os-release
	container_distro="$(grep "^ID=" /tmp/"${container_name}".os-release |
		cut -d'=' -f2- |
		sed "s|linux||g" |
		tr -d ' ' |
		tr -d '"')"
```

I wonder if this implementation brings performance degradations (list can be expensive when the number of containers is high)

balanza (changes requested):
Couldn't review all the commits, here are the first comments. Mostly about consistency with our internal conventions, plus a couple of questions to double-check the solution is working as intended.

internal/cli/create.go
fabriziosestito: Did we add tests for this? Maybe this could help document the feature.

pkg/ui/prompt.go
89luca89: We re-prompt in case of invalid response, same as shell script

pkg/commands/create.go
89luca89: Compatibility with shell, if all is default, we use my-distrobox

internal/cli/enter.go
89luca89: I think dry-run is only for binaries, other users of imports should handle this by themself

pkg/commands/generate_entry.go
89luca89: Thought the same, but on the other hand, `podman cp` can be expensive too, I'm open to discuss this
alternatively we can leave a TODO here for matching os-release approach in the future

pkg/commands/create.go
balanza: yes I mean, what was the rationale of this when writing the shell implementation at first

internal/cli/create.go
89luca89: Fixed

pkg/commands/create.go
89luca89: well I have fat fingers and when I press y I can press also uth lol so I just re-prompt

internal/cli/create.go
89luca89: Done

internal/cli/enter.go
89luca89: solved

internal/cli/enter.go
89luca89: Done

internal/cli/upgrade.go
89luca89: removed

89luca89:
> Couldn't review all the commits, here are the first comments. Mostly about consistency with our internal conventions, plus a couple of questions to double-check the solution is working as intended.

Addressed the first batch :+1:

internal/cli/util.go
```diff
@@ -1,8 +1,29 @@
...
+	return cfgValue
+}
+
+// boolOrConfig is the bool counterpart of stringOrConfig.
+func boolOrConfig(cmd *cli.Command, name string, cfgValue bool) bool {
+	if cmd.IsSet(name) {
+		return cmd.Bool(name)
+	}
+	return cfgValue
+}
```
balanza: **issue:** ‏We do not need such helpers for handling config defaults in cli arguments. Just assign the `Value` attribute to the config value, and `urfave` will consider it the default value. See https://github.com/89luca89/distrobox/commit/ec99b5d88432ff3c439e85400725ba5bf0843f70 for usage.

pkg/config/config.go
```diff
@@ -58,13 +86,30 @@ func toStruct(configMap map[string]string) *Values {
 		ContainerManagerType:  configMap["container_manager"],
 		SudoProgram:           configMap["sudo_program"],
 		Verbose:               toBool(configMap["verbose"]),
-		DefaultContainerImage: configMap["container_image"],
-		DefaultContainerName:  configMap["container_name"],
+		DefaultContainerImage: configMap["container_image_default"],
+		DefaultContainerName:  configMap["container_name_default"],
+		ContainerImage:        configMap["container_image"],
+		ContainerName:         configMap["container_name"],
```
balanza: **question:** What's the difference between `DefaultContainerImage` and `ContainerImage`? Do we use **both** distinctly in our codebase? 

I suspect that this is something that should be _resolved_ by the config module, not exposed downstream in the codebase. 

(same goes for  `DefaultContainername` and `Containername`)

balanza (changes requested):

PR #2132, opened 2026-06-22
chore(experience): align Go rewrite with reference shell distrobox across create/enter/rm/upgrade/assemble

89luca89:
Reprise of:

This branch closes a long tail of parity gaps between the Go rewrite and the reference shell distrobox. It is almost entirely behavior fixes — the commands now look at the same env vars, accept the same arguments, prompt the same way, and produce the same side effects as the shell scripts
they replace. No new top-level features; the one feat(cli) commit restores a refusal (sudo/doas guard) that the shell has and the Go port had lost.

What changed, grouped by area

create
- Honor the positional container_name argument, not just --name.
- Default the name to my-distrobox when the default image is used (shell parity).
- Drop the --image default Value so the containerImage == "" branch survives and the name == default fallback stops misfiring.
- Strip trailing slashes from --home and word-split --volume values.
- Show the host's hostname (via os.Hostname()) in --hostname usage text.
- Mount /dev/null:/run/.distrobox.rootless:ro so distrobox-init doesn't fall back to the /etc/shadow heuristic that false-positives on Docker Desktop / macOS.
- Drop rslave bind propagation on macOS; centralize in bindPropagation() / roBindPropagation() (Linux output unchanged).
- Honor DBX_USERNS_NOLIMIT to drop the keep-id size cap.
- Request icon auto-detection so generated entries aren't icon-less.

enter
- Match the shell: prompt to create the container if it doesn't exist.
- Support the noworkdir flag.
- Build the --dry-run command without requiring a live container.
- Set PWD=<workdir> inside the container and stop forwarding the host's PWD.
- Reorder PATH so /usr/local/{s,}bin precede /usr/{s,}bin.

rm
- Confirm deletions once, default prompt to yes (shell behavior).
- Use Container.IsRunning() for the force prompt — strings.Contains(Status, "Up") missed podman's lowercase running.
- Gate the custom-home prompt on --rm-home; without the flag the home is never touched.
- Wire DBX_CONTAINER_RM_CUSTOM_HOME into --rm-home.
- Print Cannot find container for unknown explicit names.

upgrade
- Match shell prompting.
- Print a per-container Upgrading <name>... banner.

stop
- Honor DBX_CONTAINER_NAME when no positional is given.

assemble
- Treat an existing box as a skip, not an error.
- Skip the rm step entirely in --dry-run (previously dryRun was misrouted into RmOptions.NoTTY and containers were actually deleted).
- Accept the singular volume manifest key alongside volumes, and whitespace-split values.

generate-entry
- Refuse a container that doesn't exist.
- Detect the distro from the image, not from the box name.
- Reuse local icons if present before attempting downloads.

CLI-wide
- SUDO/DOAS guard restored to match the shell's refusal of sudo distrobox.
- DBX_* env vars wired per command; container_manager defaults to autodetect; invalid prompt input is rejected; uid 0 is treated as rootful.


pkg/containermanager/providers/podman.go
```diff
@@ -680,6 +682,14 @@ func commandExists(cmd string) bool {
 	return err == nil
 }
 
+// usernsNoLimitEnabled reports whether DBX_USERNS_NOLIMIT requests dropping the
+// keep-id:size cap (any non-empty, non-"0" value), mirroring the shell's
+// userns_nolimit (distrobox-create:159,962-977).
+func usernsNoLimitEnabled() bool {
+	v := os.Getenv("DBX_USERNS_NOLIMIT")
+	return v != "" && v != "0" && v != "false"
+}
```
balanza: **issue:** ‏According to the current architecture, we should not reference env vars from the container manager layer. Env vars should be resolved by the `pkg/config` package and passed to the container manager object at construction time (hence, this flag must be in the container manager struct)

internal/cli/enter.go
```diff
@@ -137,7 +137,7 @@ func enterAction(ctx context.Context, cmd *cli.Command, cfg *config.Values) erro
 	options := commands.EnterOptions{
 		ContainerName:   containerName,
 		AdditionalFlags: cmd.String("additional-flags"),
-		CustomCommand:   args,
+		CustomCommand:   customCommand,
```
balanza, line 140: **issue:** ‏This is reverting a modification made in a previous commit of the same PR (https://github.com/89luca89/distrobox/pull/2132/changes/f40cdcc3f9913ca4504a02a235ae8625041ddb2b). This commit should be a fixup and squashed

internal/cli/create.go
```diff
@@ -59,7 +59,7 @@ Examples:
 			&cli.StringFlag{
 				Name:    "hostname",
 				Sources: cli.EnvVars("DBX_CONTAINER_HOSTNAME"),
-				Usage:   "hostname for the distrobox",
+				Usage:   fmt.Sprintf("hostname for the distrobox (default: %s)", defaultHostname()),
```
balanza: **issue:** ‏With the current setup, the default hostname is not the one determined by `defaultHostname()` (despite the usage message's claim). The value of `DBX_CONTAINER_HOSTNAME` env var is the actual default. 

(It should be resolved by the `config` module, though)

internal/cli/stop.go
```diff
@@ -55,6 +55,14 @@ func stopAction(ctx context.Context, cmd *cli.Command, cfg *config.Values) error
...
 	containerNames := cmd.Args().Slice()
 
+	// Shell distrobox-stop seeds container_name from DBX_CONTAINER_NAME
+	// (distrobox-stop:90, 200-202) when no positional and not --all.
+	if !all && len(containerNames) == 0 {
+		if envName := os.Getenv("DBX_CONTAINER_NAME"); envName != "" {
+			containerNames = []string{envName}
+		}
+	}
+
```
balanza, line 66: **issue:** ‏To be handled in the config pacakge

pkg/commands/rm.go
```diff
@@ -68,6 +68,22 @@ func (c *RmCommand) Execute(ctx context.Context, options RmOptions) (*RmResult,
...
+	// typos surface instead of silently no-op'ing.
+	if !options.All && len(options.ContainerNames) > 0 {
+		found := make(map[string]bool, len(distroboxesToRemove))
+		for _, d := range distroboxesToRemove {
+			found[d.Name] = true
+		}
+		for _, name := range options.ContainerNames {
+			if !found[name] {
+				//nolint:forbidigo // waiting for the logger implementation
+				fmt.Fprintf(os.Stderr, "Cannot find container %s.\n", name)
```
balanza: **issue:**  The problem with `fmt.Fprintf` in the command layer is that it is not controllable from third-party applications importing Distrobox as a library. Prefer a `printer` object instead

internal/cli/rm.go
```diff
@@ -37,8 +37,9 @@ func newRmCommand(cfg *config.Values) *cli.Command {
 				Usage:   "non-interactive mode",
 			},
 			&cli.BoolFlag{
-				Name:  "rm-home",
-				Usage: "Remove container's home directory",
+				Name:    "rm-home",
+				Sources: cli.EnvVars("DBX_CONTAINER_RM_CUSTOM_HOME"),
```
balanza: **issue:** ‏As above, to be handled in the config package

pkg/commands/rm.go
```diff
@@ -115,6 +105,21 @@ func (c *RmCommand) Execute(ctx context.Context, options RmOptions) (*RmResult,
...
+// requested name that didn't resolve to a distrobox, mirroring the shell
+// (distrobox-rm:353-356) so typos surface instead of silently no-op'ing.
+func warnUnknownContainers(requested []string, resolved []containermanager.Container) {
+	found := make(map[string]bool, len(resolved))
+	for _, d := range resolved {
+		found[d.Name] = true
+	}
+	for _, name := range requested {
+		if !found[name] {
+			fmt.Fprintf(os.Stderr, "Cannot find container %s.\n", name)
```
balanza: **issue:** ‏As above, we should avoid using fmt prints from inside the command layer (should we add a lint rule?)

balanza (changes requested):
Some changes are still needed, mostly regarding the topic of respecting the architectural boundaries.

pkg/commands/rm.go
89luca89: Solved and yes we should add a lint in the future :+1:

pkg/commands/rm.go
89luca89: Done

internal/cli/enter.go
89luca89: Done

internal/cli/create.go
89luca89: Refactored config as requested

internal/cli/create.go
```diff
@@ -40,29 +53,30 @@ Examples:
...
 				Name:    "name",
 				Aliases: []string{"n"},
-				Usage:   fmt.Sprintf("name for the distrobox (default: %s)", cfg.DefaultContainerName),
+				Value:   cfg.ContainerName,
+				Usage:   fmt.Sprintf("name for the distrobox (default: %s)", nameDefault),
 			},
 			&cli.StringFlag{
 				Name:  "hostname",
-				Usage: "hostname for the distrobox",
+				Value: cfg.ContainerHostname,
```
balanza:
```suggestion
				Value: hostnameDefault,
```

internal/cli/create.go
```diff
@@ -40,29 +53,30 @@ Examples:
...
-				),
-				Value: cfg.DefaultContainerImage,
+				Value:   cfg.ContainerImage,
+				Usage:   fmt.Sprintf("image to use for the container (default: %s)", imageDefault),
 			},
 			&cli.StringFlag{
 				Name:    "name",
 				Aliases: []string{"n"},
-				Usage:   fmt.Sprintf("name for the distrobox (default: %s)", cfg.DefaultContainerName),
+				Value:   cfg.ContainerName,
```
balanza:
```suggestion
				Value:   nameDefault,
```

internal/cli/create.go
```diff
@@ -40,29 +53,30 @@ Examples:
 			&cli.StringFlag{
 				Name:    "image",
 				Aliases: []string{"i"},
-				Usage: fmt.Sprintf(
-					"image to use for the container (default: %s)",
-					cfg.DefaultContainerImage,
-				),
-				Value: cfg.DefaultContainerImage,
+				Value:   cfg.ContainerImage,
```
balanza:
```suggestion
				Value:   imageDefault,
```

balanza (changes requested):

internal/cli/create.go
89luca89: Done

internal/cli/create.go
89luca89: Done

internal/cli/create.go
89luca89: Done

PR #2133, opened 2026-06-25
Fix/next cleanup

89luca89:
Finish v1 cleanup and finalize v2 release shape

The Go binary is the only `distrobox` entry point. v1 shell scripts are
gone, docs/man/completions describe the `distrobox <cmd>` interface, and
install/uninstall cover the full v2 artefact set with v1 cleanup on upgrade.


pkg/commands/assemble.go
```diff
@@ -215,24 +215,22 @@ func (ac *AssembleCommand) createItem(ctx context.Context, item manifest.Item, d
...
-			andAndRegex := regexp.MustCompile(`&&[[:space:]]{0,1}$`)
-
-			separator := "  " // two spaces just because v1 does that, so it's comparable in regression tests
-			if !semicolonRegex.MatchString(hook) && !andAndRegex.MatchString(hook) {
-				separator = " && "
-			}
-
-			sb.WriteString(separator)
+		if i == len(hooks)-1 {
+			continue
```
dottorblaster, line 226: I'm not sure I like this trend of putting `continue`s into for loops that started lately

dottorblaster (approved):
LGTM

PR #2139, opened 2026-06-25
feat(migrate): add `distrobox migrate` subcommand for v1→v2 containers

dottorblaster:
This PR adds a new `distrobox migrate` subcommand that lets users upgrade their existing v1 distrobox containers to work with v2. The problem is that v1 containers have bind-mounts pointing to the old script locations (e.g. `/usr/lib/distrobox/distrobox-init`), while v2 provisions those scripts to a fixed directory in the home folder (`~/.local/share/distrobox/v2/`). Since podman and docker don't allow modifying bind-mounts on an existing container, the migrate command recreates the container: it inspects the old one to recover the original creation options, stops it, commits its filesystem to a temporary image, removes the old container, and recreates it from the committed image with the v2 mountpoints. The distrobox-init entrypoint already checks for a `/.containersetupdone` marker, so it skips the full package installation phase on the recreated container and the migration is fast. It supports `--all` to migrate every distrobox, `--force` to recreate even if already migrated, `--dry-run` to preview, and `--yes` for non-interactive use.


pkg/containermanager/providers/docker.go
```diff
@@ -671,6 +685,30 @@ func (d *Docker) InspectContainer(ctx context.Context, containerName string) (*c
...
+	config.Env = inspect.Config.Env
+
+	// Docker exposes the entrypoint command (distrobox-init args) as Config.Cmd,
+	// not top-level Args (which is the parsed entrypoint path array). Prefer
+	// top-level Args if present (podman compatibility), else use Config.Cmd.
+	if len(inspect.Args) > 0 {
+		config.Cmd = inspect.Args
+	} else {
+		config.Cmd = inspect.Config.Cmd
+	}
```
balanza: **question:** ‏We do not need podman compatibility here. Does the condition make sense for docker, too?

pkg/commands/migrate.go
```diff
@@ -0,0 +1,389 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to inspect container %s: %w", name, err)
+	}
+
+	// Check if migration is needed
+	if !opts.Force {
+		if c.isAlreadyMigrated(inspectResult) {
+			return ErrMigrateAlreadyMigrated
+		}
+	}
```
balanza, line 164:
```suggestion
	if !opts.Force && c.isAlreadyMigrated(inspectResult) {
		return ErrMigrateAlreadyMigrated
	}
```

pkg/commands/migrate.go
```diff
@@ -0,0 +1,389 @@
...
+	c.printer.Println("Removing old container '%s'...", name)
+	if err := c.containerManager.Remove(ctx, name, containermanager.RmOptions{Force: true}); err != nil {
+		return fmt.Errorf("failed to remove old container %s: %w", name, err)
+	}
+
+	// Step 4: Provision v2 scripts (ensure they exist before creating)
+	if _, err := insidedistrobox.ProvisionScripts(); err != nil {
+		return fmt.Errorf("failed to provision v2 scripts: %w", err)
+	}
+
```
balanza: **suggestion:** ‏We can actually run this check before anything else, once for every container to migrate

pkg/commands/migrate.go
```diff
@@ -0,0 +1,389 @@
...
+			if mount.Source == expected {
+				return true
+			}
+			// Also accept if the source is within the v2 directory (e.g., symlink resolution)
+			if strings.HasPrefix(mount.Source, v2Dir) {
+				return true
+			}
+			return false
+		}
+	}
```
balanza: **suggestion:** ‏This introspection works right, indeed. However, it might not be resilient to future development. I propose a different approach:
- On container creation, we add a label indicating the current distrobox version, let's say `dbx-$version`
- We maintain a constant in the code with the minimum version for which a container is compatible to be executed with the current distrobox version
- a container has to be migrated if its `dbx-$version` label is inferior to the minimum required
- A container with no such a label, is always considered to be migrated 

Thoughts?

balanza: **thought:** ‏Also, I believe this function belongs to the container manager layer, as every platfom may have its own requirements and methods to determine an outdated container

pkg/commands/migrate.go
```diff
@@ -0,0 +1,389 @@
...
+
+	c.printer.Println("Migrating '%s'...", name)
+
+	// Recover the original creation options from the inspect data
+	createOpts := c.recoverCreateOptions(ctx, name, inspectResult)
+
+	if opts.DryRun {
+		c.printer.Println("[dry-run] Would stop, commit, remove and recreate container '%s' with image '%s'", name, createOpts.ContainerImage)
+		return nil
+	}
```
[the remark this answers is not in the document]
balanza: @dottorblaster I don't actually have a clue for that.

pkg/commands/migrate.go
```diff
@@ -0,0 +1,389 @@
...
+		if err := c.migrateContainer(ctx, name, opts); err != nil {
+			if errors.Is(err, ErrMigrateAlreadyMigrated) {
+				c.printer.Println("Container '%s' is already migrated to v2, skipping.", name)
+				continue
+			}
+			c.printer.PrintErrorln("error migrating %s: %s", name, err)
+			lastErr = err
+			continue
+		}
+	}
```
balanza: **suggestion:** ‏We can remove one indentation level if `migrateContainer` return (done, error)
```suggestion
		done, err := c.migrateContainer(ctx, name, opts);
		if err != nil {
			c.printer.PrintErrorln("error migrating %s: %s", name, err)
			lastErr = err
			continue
		}
	    if  !done {
			c.printer.Println("Container '%s' is already migrated to v2, skipping.", name)
			continue
		}
	}
```

pkg/containermanager/providers/docker.go
dottorblaster: You're right, I addressed it by just reducing it to one line, my bad

pkg/commands/migrate.go
dottorblaster: I moved it up, I don't think it's necessary to do it in a for loop fashion tho, we can just run this once I guess :+1:

pkg/commands/migrate.go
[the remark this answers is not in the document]
dottorblaster: I really don't like the proposed change, I'd push back on this honestly

pkg/commands/migrate.go
dottorblaster: `(done, error)` is a little bit lossy as "already migrated" and "user declined" for example are squashed into a wrong error display. But I can relate about the nesting, I tried to address this in `d5e4bd8`, let me know what you think!

pkg/commands/migrate.go
```diff
@@ -0,0 +1,376 @@
...
+		containerNames = opts.ContainerNames
+	default:
+		return ErrMigrateNoContainerSpecified
+	}
+
+	// Provision v2 scripts once, up front: they don't depend on which
+	// container is being migrated and the same files are reused for all.
+	// Skipped in dry-run to avoid host-side file writes.
+	if !opts.DryRun {
+		if _, err := insidedistrobox.ProvisionScripts(); err != nil {
```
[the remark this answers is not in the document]
dottorblaster: This feels too defensive

pkg/containermanager/providers/docker.go
```diff
@@ -671,12 +702,34 @@ func (d *Docker) InspectContainer(ctx context.Context, containerName string) (*c
 	inspect := inspects[0]
 	config.ContainerID = inspect.ID
 	config.ContainerStatus = inspect.State.Status
+	config.ContainerImage = inspect.ImageName
+	config.NetworkMode = inspect.HostConfig.NetworkMode
```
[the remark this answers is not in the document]
dottorblaster: Too defensive for now, deferring

pkg/commands/migrate.go
```diff
@@ -0,0 +1,376 @@
...
+		if standardSources[mount.Source] {
+			continue
+		}
+
+		// Skip user home mount (source == destination == $HOME)
+		if mount.Source == mount.Destination {
+			// This is likely the home directory mount or XDG_RUNTIME_DIR mount
+			// These are handled by create from the user environment
+			continue
+		}
```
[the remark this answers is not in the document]
dottorblaster: Done in [1832fb6](https://github.com/89luca89/distrobox/pull/2139/commits/1832fb65e1404972418b88cd23f2e3de7e2a6c15)

internal/inside-distrobox/scripts.go
```diff
@@ -80,3 +80,22 @@ func exists(name string) bool {
 
 	return false
 }
+
+// ScriptsDir returns the directory path where the distrobox scripts should be
+// stored on the host. This is the fixed v2 location that containers are
+// expected to bind-mount their entrypoint/export/host-exec scripts from.
+func ScriptsDir() string {
```
89luca89: how is it different from using `config.resolveScriptsDir()`?

internal/cli/migrate_internal_test.go
```diff
@@ -0,0 +1,287 @@
+// SPDX-License-Identifier: GPL-3.0-only
```
89luca89, line 1: it would be nice here to have an unit tests that actually runs the migration on mock data:

v1 json -> migrate -> v2 json -> compare to known good

89luca89 (changes requested):
Added a couple of comments, other than those, rest lgtm

internal/inside-distrobox/scripts.go
dottorblaster: Yeah this effort started before the merge, done in 8abacf7, thanks :+1:

internal/cli/migrate_internal_test.go
dottorblaster: Makes sense, done in 8abacf7

89luca89 (approved):

PR #2140, opened 2026-06-28
chore(build): create shell command wrapper for V1 compatibility.

balanza:
In the artefact directory (`./bin`), alongside the `distrobox` executable, other shell executable files are added to mimic the content of the distrobox v1 bundle. Shell executables are of two kinds:

* `assets`: those files that are executed inside the container. In v2, those files are budled into the binary and provisioned just-in-time. However, containers created with distrobox v1 expect those files to be reachable and thus the need for copying them and make them available. 
* `shims`: shell executables in the form `distrobox-<cmd>` that are alias for the relative commands and are not present anymore in v2. For example: `distrobox-enteri` file is a shim for the command `distrobox enter`.

Usage:
Run `make` and the files are provisioned only if not already.
```
› make
CGO_ENABLED=0 GOOS=linux go build -ldflags "-X github.com/89luca89/distrobox/pkg/version.Version=1.8.2.5-203-g2fa1339" -o bin/distrobox ./cmd/distrobox
  COPY    bin/distrobox-export
  COPY    bin/distrobox-host-exec
  COPY    bin/distrobox-init
  SHIM    bin/distrobox-assemble
  SHIM    bin/distrobox-create
  SHIM    bin/distrobox-enter
  SHIM    bin/distrobox-ephemeral
  SHIM    bin/distrobox-generate-entry
  SHIM    bin/distrobox-list
  SHIM    bin/distrobox-rm
  SHIM    bin/distrobox-stop
  SHIM    bin/distrobox-upgrade
```

PR #2141, opened 2026-06-28
chore(ci): bundle release

balanza:
The release now contains a bundled directory for every arch, so that in the bundled directory, all the executables for distrobox are included

This is especially required after #2140


dottorblaster (approved):
LGTM

PR #2142, opened 2026-06-29
feat(compat): preserve v1 entry points after v2 single-binary split

89luca89:
v1 containers bind-mount distrobox-{init,export,host-exec} from host $bindir, and v1 desktop entries launch via `distrobox-enter`. Both would silently break under v2's single binary. argv[0] dispatch routes distrobox-* symlinks to the right subcommand, and the helpers ship at the v1 paths so v1 containers keep working.

BREAKING CHANGE: v2 rc1 and rc2 prerelease containers used ~/.local/share/distrobox/v2/ for their helpers and must be recreated.

### DIFF
--- a/internal/inside-distrobox/scripts_test.go
+++ b/internal/inside-distrobox/scripts_test.go
@@ -11,47 +11,108 @@ import (
 	insidedistrobox "github.com/89luca89/distrobox/internal/inside-distrobox"
 )
 
-func TestProvisionScripts_CustomDir(t *testing.T) {
-	tmpDir := t.TempDir()
-	t.Setenv("DBX_SCRIPTS_DIR", tmpDir)
+// expectedScripts is the canonical triad of in-container helper scripts.
+// ProvisionScripts must always end with all three present in the returned
+// directory, regardless of which resolution branch produced it.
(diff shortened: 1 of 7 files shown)


internal/inside-distrobox/scripts_test.go
```diff
@@ -11,47 +11,109 @@ import (
...
+// test as a defensive skip when the binary's directory isn't writable
+// (e.g. read-only test harness, distros with hardened tmpfs).
+func isDirWritable(dir string) bool {
+	probe, err := os.CreateTemp(dir, ".dbx-write-probe-")
+	if err != nil {
+		return false
 	}
+	name := probe.Name()
+	_ = probe.Close()
+	_ = os.Remove(name)
```
dottorblaster: Can't we `defer` these?

dottorblaster (commented):
Just one nit, then looks good :)

internal/inside-distrobox/scripts_test.go
89luca89: Fixed

dottorblaster (approved):
LGTM

PR #2149, opened 2026-07-03
fix: use overlay mount for fonts to prevent cache poisoning

### LINKED ISSUE #1945: [Error] Fonts overriding font cache of main system
**Describe the bug**
Running `fc-cache -fv` after installing new fonts modifies the host system's files of fontconfig. This is undesirable.

podman
- Which version of distrobox?
1.8.2.2
- Which host distribution?
Void Linux
install script

bhartigautam156:
Problem : The previous approach of bind mounting host fonts at /usr/local/share/fonts inside the container caused font cache poisoning. When any app inside the container updated the shared fontconfig cache (~/.cache/fontconfig), it
recorded /usr/local/share/fonts/... paths which do not exist on the host.This left the host desktop environment unable to find its own fonts.

Fix this by using an overlay mount that merges host fonts directly into the container's /usr/share/fonts. This ensures that both host and container see fonts at the same path, so any cache update is valid for both sides.

The overlay mount uses:
- lowerdir: host fonts at /run/host/usr/share/fonts (read-only)
- upperdir: container's own fonts (read-write)
- merged result at /usr/share/fonts inside the container

Kernel OverlayFS is tried first, with fuse-overlayfs as fallback for rootless containers. fuse-overlayfs is added to the deps list of all supported package managers so it is available when needed.

Fixes: https://github.com/89luca89/distrobox/issues/1945

### DIFF
--- a/internal/inside-distrobox/assets/distrobox-init
+++ b/internal/inside-distrobox/assets/distrobox-init
@@ -727,6 +727,7 @@ setup_apk()
 		util-linux-login
 		util-linux-misc
 		vulkan-loader
+		fuse-overlayfs
 		wget
 		xauth
 		xz
@@ -851,6 +852,7 @@ setup_apt()
 		libglx-mesa0
 		libvulkan1
 		mesa-vulkan-drivers
+		fuse-overlayfs
 	"
 	# shellcheck disable=SC2086,2046
 	apt-get install -y $(apt-cache show ${deps} 2> /dev/null | grep "Package:" | sort -u | cut -d' ' -f2-)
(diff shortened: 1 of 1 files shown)


bhartigautam156:
Hey @89luca89 , can you please approve the workflows and review this PR ?

bhartigautam156:
Hey @dottorblaster @89luca89 , Can I get review on this PR?

bhartigautam156:
Hey @dottorblaster  , can you please approve the workflows and also review this PR ?

89luca89:
Hi @bhartigautam156 i've started the CI on it, i'm usually a bit cautious on adding dependencies in the init, due to the large compatibility portfolio we need to support. Let's see how CI goes, if the package is available on all the testing suite, then we should not have problems. If it's not we either solve this at container manager level, or we will have to find a way to gracefully handle this.

PR #2151, opened 2026-07-05
[wontfix] remove v1 hyphenated command compatibility layer

kairosci:
This change removes the ResolveArgs function that dispatched hyphenated commands by argv[0] (distrobox-enter, distrobox-create, etc.) and the symlink installation in both the install script and the Makefile, keeping only the native v2 subcommand dispatch.


balanza:
This PR reverts the job done in #2142 only a few days ago, without providing any explanation of the rationale behind the choice. 

Closing now to preserve the maintainer's time.

@kairosci thanks for the contribution. Consider opening this PR again once you can make clear your intentions and the benefits of the changes.

kairosci:
@balanza Hi, you're absolutely right, and I apologize for the lack of context. It was my mistake not to explain the reasoning.

My goal with this PR is to simplify the code by switching exclusively to native v2 subcommand dispatch. By completely removing the ResolveArgs logic and symbolic link handling, we can permanently eliminate a source of technical debt and potential bugs in the installation scripts. Additionally, this resolves the UX issue caused by duplicate commands (such as distrobox-create and distrobox create) cluttering the terminal’s autocomplete, while also aligning Distrobox with modern CLI standards used by toolbx and podman.

I know backward compatibility is important, so if immediately removing the hyphenated commands proves too disruptive, we could instead introduce a deprecation warning or postpone this cleanup to a future major release. 

If you think this is feasible, I'd appreciate it if you could reopen it so we can discuss it directly here.

Thanks for your time!

balanza:
> I know backward compatibility is important

I think you said it all here. But even if I may share your concerns, when you remove some code, you MUST provide an alternative solution to the problem that code was addressing.

Furthermore, you are brutally reverting a specific contribution from the project owner and maintainer for more than 4 years. Have you considered the idea that such code has a valid rationale behind it, and maintainers had gone through a series of trade-offs? And better, have you considered raising your doubts before?

> you're absolutely right,

Last but not least, this PR appears to me as slop. _Your_ phrasing is very indicative in this regard.

kairosci:
Is saying that I agree with what you said considered a "slop"? I didn't know that. In any case, I'll open an issue to discuss it.

PR #2155, opened 2026-07-08
fix: prevent host-spawn from consuming profile stdin

hypengw:
Fix `distrobox_profile.sh` and fish profile generation so `host-spawn` does not consume the profile shell's stdin.

When a login shell is started with script input from stdin, for example:

```sh
echo 'echo xxx;' | bash -l
```

The generated profile can run host-spawn while inheriting that same stdin. This may consume the remaining script input before the login shell executes it.

This redirects host-spawn calls to use `/dev/null` .

### DIFF
--- a/internal/inside-distrobox/assets/distrobox-init
+++ b/internal/inside-distrobox/assets/distrobox-init
@@ -2231,20 +2231,20 @@ test -z "\${DBUS_SESSION_BUS_ADDRESS:-}" && export DBUS_SESSION_BUS_ADDRESS="uni
 # Ensure we have these two variables from the host, so that graphical apps
 # also work in case we use a login session
 if [ -z "\$XAUTHORITY" ]; then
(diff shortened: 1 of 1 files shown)


balanza:
Hi @hypengw, thanks for the contribution. I am not sure I understand correctly what is the scenario that you are addressing. Can you provide more instructions to replicate your setup? 

Thanks

MythreyaK:
Facing the same issue, and cooked up the same solution. 


> Can you provide more instructions to replicate your setup?

I agree this usecase is perhaps a little niche: 
* Install vscodium
* Install the remote ssh extension [`jeanp413.open-remote-ssh`](https://open-vsx.org/extension/jeanp413/open-remote-ssh)
* Attempt to ssh to the distrobox from the host
* codium server install script (from the extension) fails because stdin is lost in `/etc/profile.d/distrobox_profile.sh` 

But is incredibly handy, amongst other things, to test how extensions behave when working over an ssh remote. Would love to see this merged!

PR #2158, opened 2026-07-08
fix: ensure rootful containers add user to sudo/wheel group

### LINKED ISSUE #2017: [Error] Distrobox doesn't add user to `wheel` group
Hi!

I'm creating a test container like this:
```bash
distrobox create \
    --name RockyLatestSystemd \
    --hostname RockyLatestSystemd \
    --init \
    --image rockylinux:9.3 \
    --additional-packages "systemd" \
    --root
```

The problem is that, once inside the container, I can't use `sudo` (and my user is not in the `wheel` group).

For example:
```bash
# Outside the container, sudo works just fine!
$ sudo echo hello, world
hello, world

# I enter the container
$ distrobox enter --root RockyLatestSystemd
# I try to use `sudo` ...
$ sudo echo hello, world

We trust you have received the usual lecture from the local System
Administrator. It usually boils down to these three things:

    #1) Respect the privacy of others.
    #2) Think before you type.
    #3) With great power comes great responsibility.

[sudo] password for vvz3n: 
vvz3n is not in the sudoers file.  This incident will be reported.

$ groups
vvz3n

# Outside the container, I'm in the `wheel` group!
$ exit
$ logout
$ groups
vvz3n wheel
```

This is happening on
```
$ distrobox version
distrobox: 1.8.2.4
```

kairosci:
Remove the rootful host-group-check that cleared additional_groups when /run/host/etc/group was inaccessible or the host had no wheel/sudo membership. Also fix the useradd failure fallback to properly add the user to supplementary groups.

Closes https://github.com/89luca89/distrobox/issues/2017

### DIFF
--- a/internal/inside-distrobox/assets/distrobox-init
+++ b/internal/inside-distrobox/assets/distrobox-init
@@ -2423,16 +2423,6 @@ elif grep -q "^root" /etc/group; then
 	additional_groups="root"
 fi
 
-# If we're rootful, search for host's groups, if we're not in anyone, let's not
-# add the current user to any sudoers group, so that host's sudo settings are
-# respected
-if [ "${rootful}" -eq 1 ] &&
-	! grep -q "^wheel.*${container_user_name}" /run/host/etc/group &&
-	! grep -q "^wheel.*${container_user_name}" /run/host/etc/group &&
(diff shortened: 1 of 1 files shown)


dottorblaster (approved):
LGTM

PR #2161, opened 2026-07-11
fix(enter): parse additional-flags when placed after the distrobox name

### LINKED ISSUE #2160: [Error] Inconsistent argument parsing for `distrobox enter` in Distrobox 2.0.0-rc.3
**Describe the bug**

In Distrobox 2.0.0-rc.3, invoking `distrobox enter` with `--additional-flags` placed after the container name results in an error. This regression breaks Distrobox integration in Ptyxis, where the flag is used in [`agent/ptyxis-distrobox-container.c`](https://gitlab.gnome.org/chergert/ptyxis/-/blob/main/agent/ptyxis-distrobox-container.c) for TTY redirection.

**To Reproduce**
```console
$ distrobox enter --no-tty test --additional-flags "--tty"
Error: crun: executable file `--additional-flags` not found in $PATH: No such file or directory: OCI runtime attempted to invoke a command that was not found
```

**Expected behavior**

The container should enter successfully, as it does in Distrobox v1.x.

**Desktop**

+ podman 5.8.2
+ Distrobox 2.0.0-rc.3
+ Vanilla OS 3.0

dottorblaster:
Fixes #2160.

I took a look, and it looks like this is the only missing case that we need to parse. All of the other dash-dash args are nonsensical when placed after the box' name.

### DIFF
--- a/internal/cli/enter.go
+++ b/internal/cli/enter.go
@@ -93,6 +93,17 @@ func enterAction(ctx context.Context, cmd *cli.Command, cfg *config.Values) erro
 
 	args := cmd.Args().Slice()
 
+	// When --additional-flags (or -a) is placed after the first positional
+	// argument it lands in the positional tail instead of being consumed as
+	// a flag, because urfave/cli's StopOnNthArg: 1 stops flag parsing after
+	// the first positional arg. We recover it here.
+	args, tailAdditionalFlags := extractAdditionalFlagsFromPositionals(args)
+
+	additionalFlags := cmd.String("additional-flags")
+	if tailAdditionalFlags != "" {
+		additionalFlags = tailAdditionalFlags
+	}
+
 	// If the user placed -e/--exec AFTER the container name, it lands in
 	// the positional tail. In that case the first positional arg is still
 	// the container name and the custom command starts right after the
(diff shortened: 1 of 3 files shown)


NN708:
@dottorblaster Thank you for the quick fix! I've tested this branch with Ptyxis, and while the `--additional-flags` issue is resolved, Ptyxis still fails to enter Distrobox containers. The new error is:

```
Error: crun: executable file `--` not found in $PATH: No such file or directory: OCI runtime attempted to invoke a command that was not found
```

It appears that `--` also needs proper parsing to maintain consistent behavior with Distrobox v1.x.

dottorblaster:
Ouch! Thanks for letting me know in advance @NN708, maybe I can come up with something even tho it looks like the easiest solution is to go to the Ptyxis source code and fix it :grimacing:

I'm a little worried about the approach we have to use to also parse `--` but still... let's see if I can make up something.

NN708:
> Ouch! Thanks for letting me know in advance @NN708, maybe I can come up with something even tho it looks like the easiest solution is to go to the Ptyxis source code and fix it 😬

@dottorblaster I've considered that too, but fixing it in Ptyxis could potentially break compatibility with Distrobox v1, which people are still using. It could also take a long time for the fix to reach users through packaging on different distros. So I think it's better to handle it in a way that keeps Distrobox v2 100% feature-parity with v1.

89luca89:
Opened another PR here https://github.com/89luca89/distrobox/pull/2165
it should do the original behaviour

89luca89:
closing this as now is superseded by #2165

PR #2162, opened 2026-07-11
fix(init): check if group exists before appending user

kairosci:
This fixes a long-standing bug in the user creation fallback where `sed` crashes if an additional group (like `wheel` or `sudo`) does not exist in `/etc/group`. We now explicitly verify that the group exists before attempting any substitution.


dottorblaster (approved):
It doesn't change much from a CI standpoint because the red checks we see are mainly due to distros' failures, but I don't mind a more defensive approach

Issue #2152, opened 2026-07-06
[Error] `distrobox --version` no longer available in 2.0.0-rc.3


NN708:
**Describe the bug**

In Distrobox 2.0.0-rc.3, running `distrobox --version` no longer outputs version information.

**To Reproduce**
```console
$ distrobox --version
flag provided but not defined: -version
```

**Expected behavior**

v1.x:
```console
$ distrobox --version
distrobox: 1.8.2.5
```

v2:
```console
$ distrobox --version
distrobox version dev
```

**Desktop**

+ podman 5.8.2
+ Distrobox 2.0.0-rc.3
+ Vanilla OS 3.0

PR #2165, opened 2026-07-17
fix(enter): parse distrobox flags after the container name

### LINKED ISSUE #2160: [Error] Inconsistent argument parsing for `distrobox enter` in Distrobox 2.0.0-rc.3

89luca89:
`distrobox enter my-box --help` and any distrobox flag placed after the container name, was passed to the container manager as the command instead of being parsed, so it failed with a crun error. This diverged from the original bash distrobox-enter, which keeps parsing flags after the name and only starts the command at -e/--exec/--.

The port stopped flag parsing at the first positional (StopOnNthArg), which cannot tell the name from the command: when --name or DBX_CONTAINER_NAME already supplies the name, the first positional *is* the command. So the command boundary is resolved before parsing and everything past it is handed to urfave verbatim, letting flags be recognized wherever they appear while the custom command is still passed through untouched. ephemeral gets the same treatment.

Unknown flags after the name are rejected, as in bash, so a mistyped flag surfaces instead of being silently executed inside the container.

Fixes https://github.com/89luca89/distrobox/issues/2160

### DIFF
--- a/internal/cli/parse.go
+++ b/internal/cli/parse.go
@@ -1,17 +1,184 @@
 package cli
 
-// findExecMarkerIndex returns the index of the first -e or --exec in
-// args, or -1 if neither is present. It is used by the enter and
-// ephemeral commands to recover the marker when the user placed it
-// after the container name (or, for ephemeral, after the first
-// positional arg) — in those cases urfave/cli leaves the marker in
-// the positional tail instead of consuming it as a flag, and we need
-// to know where the custom command starts.
-func findExecMarkerIndex(args []string) int {
-	for i, arg := range args {
-		if arg == "-e" || arg == "--exec" {
-			return i
+import (
+	"strings"
+
+	"github.com/urfave/cli/v3"
+)
+
+// PrepareArgs lets `enter` and `ephemeral` accept distrobox flags after the
+// container name, like the bash distrobox-enter (which only starts the command
+// at -e/--exec/--). urfave's StopOnNthArg can't express this: it stops at the
+// first positional and can't tell the name from the command. So we find the
+// command ourselves and splice a bare "--" in front of it, then let urfave
+// parse the rest. It runs at the entrypoint because urfave has no pre-parse hook.
+func PrepareArgs(root *cli.Command, args []string) []string {
+	if len(args) < 2 {
+		return args
+	}
+
+	// Completion appends this flag last; splicing a "--" ahead of it would
+	// hide it from urfave.
+	if args[len(args)-1] == "--generate-shell-completion" {
+		return args
+	}
+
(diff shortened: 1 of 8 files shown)


89luca89:
@NN708 This should behave like intended, can you test with ptyxis if this matches the V1 behaviour?

EDIT: i just did a basic test (select a dbox from the drop down menu) and it worked, but not sure if this is enough I'll remand to you to reproduce what you did on the issue

NN708:
@89luca89 Thank you! I can confirm that Ptyxis is now working as expected.

~However, I noticed that `distrobox --version` (#2152) appears to be broken again.~

89luca89:
> @89luca89 Thank you! I can confirm that Ptyxis is now working as expected.
> 
> However, I noticed that `distrobox --version` (#2152) appears to be broken again.

:thinking: locally `./bin/distrobox --version` worked, do you mean `--version` in subcommands?

NN708:
Oh, it works. I didn't realize the binary had been overwritten during the tests. Sorry for the confusion!

NN708 (approved):

89luca89:
Right, thanks for reporting :+1: @dottorblaster i'll wait for a review of yours

internal/cli/parse.go
```diff
@@ -1,17 +1,184 @@
...
+		return name, true
+	}
+	return base, false
+}
+
+func isExecMarker(tok string) bool { return tok == "-e" || tok == "--exec" }
+
+func isHelpToken(tok string) bool {
+	base, _ := flagBaseName(tok)
+	return isFlagToken(tok) && (base == "help" || base == "h")
```
dottorblaster, line 173: I don't want to have `cli` in here but instead of hardcoding `help` we could also use `cli.HelpFlag.Names()`

dottorblaster (approved):
Let's consider also adding a huge suite of end to end tests to lock this ❤️

PR #2171, opened 2026-07-21
fix(gpu): isolate NVIDIA libraries and fix Docker device groups for ROCm

89luca89:
Some improvements on both NVIDIA and AMD gpu integration

NVIDIA: the `--nvidia` path mapped each driver library to the guest by rewriting its host
path (x86_64-linux-gnu / lib64 / lib32 via sed), which is fragile across distros and can
shadow the guest's own libGL/glvnd/Mesa. Libraries now land in an isolated
/usr/lib/distrobox-nvidia/{lib64,lib32}, bucketed by ELF class and exposed via
one ld.so.conf entry + ldconfig; Xorg/GBM/VDPAU/other files plugins keep their real path. 

Docker: ROCm reaches the GPU purely through device nodes (/dev/kfd, /dev/dri/renderD*),
which are group-owned (render, 0660). podman+crun keeps the user's supplementary groups via
`run.oci.keep_original_groups`, but the docker path forwarded only the primary gid, so the
user hit EACCES. 
This puts docker in feature parity with podman+crun.

**This needs some testing on nvidia and rocm hardware**. I don't have one so This is also a call for testing!

### DIFF
--- a/.github/workflows/nvidia-integration.yml
+++ b/.github/workflows/nvidia-integration.yml
@@ -1,0 +1,92 @@
+---
+name: NVIDIA integration
+
+# distrobox-init holds the --nvidia mirroring logic, so only run this expensive
+# VM-based test when that script (or the test harness itself) changes.
+on:
+  pull_request:
+    branches: [main]
+    paths:
+      - internal/inside-distrobox/assets/distrobox-init
+      - hack/test/**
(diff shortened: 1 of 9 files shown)


dottorblaster (approved):
LGTM

Issue #2180, opened 2026-07-30
[Error]failed to commit container 'xxxxxxx' : command execution failed: Error : name or ID cannot be empty


Barlafuss:
using rc4 version to clone a container it fails with above error msg.

host = Kalpa (openSUSE)
podman = 5.8.3-1.3
distrobox = 2.0.0-rc.4
command = distrobox-create --clone 'Source' --name 'Copy'

PR #2190, opened 2026-08-07
ci(e2e): add end-to-end behavioral tests across distros and engines

89luca89:
Builds on the existing compatibility CI. The per-distro e2e suite now also runs
as a fast image subset that gates every PR, alongside a new distro-agnostic
CLI/flag suite so the v2 binary is proven on each supported distro and on both
podman and docker before merge, not only in the post-merge full matrix. The gate
and the full matrix now share one reusable workflow, and image lists live only in
the Makefile, so `make e2e-one IMAGE=… CM=…` reproduces any CI cell locally.

Contextually added some fixtures tests for the behavior of slug generation, env-var filtering, and the
ps parser so refactors can't silently change them

### DIFF
--- a/.github/workflows/compatibility.yml
+++ b/.github/workflows/compatibility.yml
@@ -1,25 +1,23 @@
 ---
-# This is a basic workflow to help you get started with Actions
-
 name: CI
 
-# Controls when the workflow will run
+# Full compatibility suite: the e2e matrix over every supported image
+# (docs/compatibility.md -> Makefile E2E_FULL_IMAGES), both backends. Runs on main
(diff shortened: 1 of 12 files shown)


89luca89:
just an heads up, failing e2e tests now are a real finding, they found and confirmed bug #2180

dottorblaster (approved):
This is perfect to me, the only question I have is, do we want to merge this and solve the bugs in another branch or do we want to solve the bugs and then rebase this branch against `main` when the fixes land?

89luca89:
We fix it in another PR, they should have a matching issue somewhere already when ch is very good!

PR #2193, opened 2026-08-09
fix(enter): use a real isatty check for --tty allocation

dottorblaster:
`containermanager.IsTTY()` detected a "terminal" by checking for a character device (`os.ModeCharDevice`). `/dev/null` is a character device, so headless invocations (`distrobox enter ... < /dev/null > /dev/null 2>&1`) were treated as interactive, allocating `docker exec --tty`.

Docker rejects that with "cannot attach stdin to a TTY-enabled container because stdin is not a terminal", so the exec failed before running the command, e.g. `sudo -n touch` in the e2e clone test never ran and the clone lost its rootfs marker. Podman tolerated the flag, hiding the bug.

### DIFF
--- a/pkg/containermanager/containermanager_test.go
+++ b/pkg/containermanager/containermanager_test.go
@@ -20,6 +20,7 @@
 package containermanager_test
 
 import (
+	"os"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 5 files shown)


pkg/containermanager/containermanager_test.go
```diff
@@ -108,3 +109,30 @@ func TestContainer_IsDistrobox_DistroboxKeyPrefix(t *testing.T) {
...
+// /dev/null is a character device but NOT a terminal, and treating it as one
+// made headless invocations (`enter ... > /dev/null 2>&1`) allocate a --tty
+// that docker exec rejects with "cannot attach stdin to a TTY-enabled
+// container because stdin is not a terminal".
+func TestIsTTY_DevNullNotATerminal(t *testing.T) {
+	// Swap both stdio fds to /dev/null for the duration of the test.
+	devNull, err := os.OpenFile(os.DevNull, os.O_RDWR, 0)
+	if err != nil {
+		t.Fatalf("open %s: %v", os.DevNull, err)
+	}
```
dottorblaster: We could just use `assert.NoError()` here I think

89luca89 (approved):

PR #2194, opened 2026-08-09
fix(e2e): apply system upgrade in Arch Linux setup during nvidia tests

dottorblaster:
Since a stale image was causing a quirky behavior with shared libraries, just applying a full system upgrade during the setup of `vulkan-tools` for Arch Linux tests.


89luca89:
I'll work on this

PR #2195, opened 2026-08-09
fix(e2e): update host distros on nvidia tests, avoid problems with stale dependencies

89luca89:
It happens that cloud images are old and installing newer packages would lead to stale dependencies (eg archlinux old libc but new podman package). This PR will introduce a precautionary update on the VM before proceeding with the test.


dottorblaster (approved):
Aw yissss

89luca89:
> Aw yissss

sssssso guuuuuuuuuuuuuud

PR #2202, opened 2026-08-22
Fix duplicate systemd user managers

enihcam:
In initful containers, Distrobox starts `user@<name>.service` and then enables linger for the same user. systemd-logind starts `user@<uid>.service`, so both managers can run for one UID.

Use the numeric UID for the user manager, runtime integration, and linger setup.

This avoids duplicate user managers and conflicting user generator warnings.

Tests:

- `make test`
- ShellCheck on `distrobox-init`


dottorblaster:
I'm not understanding if we should review this or the previous one. Feel free to reopen the one with the implementation you think it's the most correct one.

enihcam:
this one please. I cannot reopen it.

PR #2214, opened 2026-09-01
fix(tests): run tests rootless, fix archlinux, fix internal script extraction path

89luca89:
- scripts.go: ProvisionScripts reuses the helpers if already present, else extracts into the install dir when writable, otherwise falls back to $XDG_DATA_HOME/distrobox. Fixes rootless create/enter for a system-wide install (/usr/local/bin), which died with "permission denied"
- hack/test/nvidia-test.sh: run box ops rootless as the cloud user
- hack/test/vm-run.sh: bump the ubuntu image to resolute.

### DIFF
--- a/internal/inside-distrobox/scripts.go
+++ b/internal/inside-distrobox/scripts.go
@@ -35,15 +35,10 @@ var initScript string
 //go:embed assets/distrobox-export
 var exportScripts string
 
-// ProvisionScripts ensures that all necessary scripts are created in the given directory.
-// It returns scriptsDir unchanged on success.
+// ProvisionScripts ensures the helper scripts exist on disk and returns their
+// directory: scriptsDir when they are already there or it is writable, else a
+// per-user data dir (so a system-wide install still works when run rootless).
 func ProvisionScripts(scriptsDir string) (string, error) {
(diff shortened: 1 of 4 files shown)


internal/inside-distrobox/scripts.go
```diff
@@ -68,15 +85,26 @@ func ProvisionScripts(scriptsDir string) (string, error) {
...
-			return true
-		}
+// dirWritable reports whether a file can be created in dir (probes with a temp file).
+func dirWritable(dir string) bool {
+	probe, err := os.CreateTemp(dir, ".dbx-write-probe-")
+	if err != nil {
+		return false
+	}
+	_ = probe.Close()
+	_ = os.Remove(probe.Name())
```
dottorblaster: Wouldn't it be more idiomatic to just `defer` these calls?

dottorblaster (dismissed):
Just one nit but LGTM

internal/inside-distrobox/scripts.go
89luca89: done fr :atm:

dottorblaster (approved):
LGTM

PR #2223, opened 2026-09-17
fix(init): resync watched mounts after host file replacement

89luca89:
A bind mount tracks the inode, not the path. When the host replaces one of the HOST_WATCH files, the container mount keeps pointing at the unlinked inode and findmnt reports it with a "//deleted" source. The keepalive loop only compared file contents, so a byte-identical recreate left the mount disconnected, and when /run/host exposed the same deleted peer the content check could never trigger at all.


dottorblaster (approved):
LGTM

